PR #65, opened 2017-04-24
[ASCollectionView] Prevent prefetching from being enabled to eliminate overhead.

appleguy:
Some apps explicitly enable this, which is counterproductive with ASDK. This can also worsen data inconsistency issues as it creates additional traffic from UIKit.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -251,12 +251,12 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   if (!(self = [super initWithFrame:frame collectionViewLayout:layout]))
     return nil;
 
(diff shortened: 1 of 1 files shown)


maicki:
@Adlai-Holler @nguyenhuy Did we not recently have a conversation about this, as we discovered some problem around it? What was the conclusion again? Otherwise looks good to me!

nguyenhuy:
@maicki The conclusion is that we want to disable prefetching universally and rely on our built-in ranges. 

Prior to https://github.com/facebook/AsyncDisplayKit/pull/3263, `ASCollectionNode` and `ASTableNode` don't prefetch when they enter preload state.  This problem should be fixed now.

appleguy:
Awesome, thanks everyone!

PR #74, opened 2017-04-27
Make Cell Node Properties Atomic

Adlai-Holler:
This way if you do logging or other tasks on background threads, you can read `indexPath` and `supplementaryElementKind` safely.

It kind of spiraled out of control, so:
- Add `ASRangeManagedNode` protocol to cover `ASCollectionNode` and `ASTableNode`.
- Remove `ASDataControllerEnvironmentDelegate` – instead pass in `ASRangeManagedNode` to init.
- Make the storage for `ASDataController.visibleMap/pendingMap` `atomic`.
- Make the storage for `ASCellNode.owningNode` `atomic`.
- (For good measure) Modify places in ASDataController where we repeatedly access a weak variable to use a strong variable instead.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRangeManagedNode.h:</p>
<pre><code>//
//  ASRangeManagedNode.h
//  Texture
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTableViewTests.mm:</p>
<pre><code>//
//  ASTableViewTests.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASCellNode.h
```diff
@@ -126,10 +125,9 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 
 
 /**
- * The owning node (ASCollectionNode/ASTableNode) of this cell node, or @c nil if this node is
- * not a valid item inside a table node or collection node or if those nodes are nil.
+ * The table- or collection-node that this cell is a member of, if any.
  */
-@property (weak, nonatomic, readonly, nullable) ASDisplayNode *owningNode;
+@property (weak, readonly, nullable) id<ASRangeManagedNode> owningNode;
```
Adlai-Holler: This s technically a breaking API change but it's so minor that we shouldn't worry about it. Virtually nobody uses this property – it was recently added and probably shouldn't be public anyway – and the migration pathway is trivial.

Adlai-Holler:
Hmm I wonder if `ASRangeManagedNode` should be a base class rather than a protocol. It may well evolve into one over time. I remember @appleguy you proposed having a shared base class for these kinds of things. Let's leave it as a protocol for now, then see what the world looks like after we make naked-ASTableView-and-ASCollectionView usage unavailable.

Source/ASCellNode.h
```diff
@@ -87,7 +88,7 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  * @return The supplementary element kind, or @c nil if this node does not represent a supplementary element.
  */
 //TODO change this to be a generic "kind" or "elementKind" that exposes `nil` for row kind
-@property (nonatomic, copy, readonly, nullable) NSString *supplementaryElementKind;
+@property (copy, readonly, nullable) NSString *supplementaryElementKind;
```
appleguy: I know it is implied, but stylistically, I think it would be a good idea to require atomic to be stated explicitly for properties that intend it. Although it might become the more common thing, currently it's uncommon enough that most cases where it is omitted could be confused for mistakes. Any thoughts?

appleguy:
@Adlai-Holler Yes, I still think having a true bass class would be a very good thing. It would be a substantial project to build - I attempted it over Christmas holiday the year before this most recent, and it got complicated. I think it would be an architectural gift that keeps on giving, though :). In particular, it could even provide some APIs that are appropriate to offer publicly, and allow advanced apps to create custom range-managed scrollers.

Source/ASCellNode.h
Adlai-Holler: I'm down with it. The explicitness is nice.

Source/ASRangeManagedNode.h
```diff
@@ -0,0 +1,35 @@
...
+#import <AsyncDisplayKit/ASTraitCollection.h>
+
+@class ASCellNode;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Basically ASTableNode or ASCollectionNode.
+ */
+@protocol ASRangeManagedNode <NSObject, ASTraitEnvironment>
```
garrettmoon: Maybe I'm being overly semantic, but the methods these adopt don't seem to match the name. I can imagine nodes that range manage other nodes that don't have index paths. Anyway, wondering if there's a better name but I don't have suggestions.

Also, should it be `ASRangeManagingNode` instead? That would better align with `ASHierarchyStateRangeManaged` state naming.

Source/ASTableNode.mm
```diff
@@ -80,7 +80,7 @@ - (instancetype)initWithStyle:(UITableViewStyle)style
     [self setViewBlock:^{
       // Variable will be unused if event logging is off.
       __unused __typeof__(self) strongSelf = weakSelf;
```
garrettmoon: Is the strongSelf necessary? If the viewBlock is running, self shouldn't be nil, right?

Source/Details/ASDataController.h
```diff
@@ -160,17 +155,30 @@ extern NSString * const ASCollectionInvalidUpdateException;
...
+ */
+@property (nonatomic, nullable, weak, readonly) id<ASRangeManagedNode> node;
 
 /**
  * The map that is currently displayed. The "UIKit index space."
+ *
+ * This property will only be changed on the main thread.
  */
-@property (nonatomic, strong, readonly) ASElementMap *visibleMap;
+@property (atomic, copy, readonly) ASElementMap *visibleMap;
```
garrettmoon: Are there asserts that ensure this?

Source/ASTableNode.mm
Adlai-Holler: Yep but we should strongify once, rather than implicitly strongifying each time we use `weakSelf`.

Source/Details/ASDataController.h
Adlai-Holler: We don't need asserts to check this. We only change it internally. And we can't override atomic accessors so it's impossible to assert.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/31971e91f938ff345dc936e81fea1970/raw/c1cebc72826d3fd1fda58ad9495e509a7f66e581/log.txt)

PR #80, opened 2017-04-27
Fix inserting a sublayer in an index of NSNotFound

maicki:
It can happen that the view get's loaded while `_setSupernode` is called. In this it's still assumed that the `sublayerIndex` is `NSNotFound` although the view is loaded now and we should have a valid sublayer index. To prevent this let's grab the sublayer index again if we discover this situation.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #82, opened 2017-04-28
Tighten Rasterization API, Undeprecate It

Adlai-Holler:
After playing with Pinterest for a little bit, there are some views that would majorly benefit from rasterization if we could get it going. Not that I'm addicted to it! But here's what this diff does:

- Replaces the `@property (readwrite) shouldRasterizeDescendants` with a one-way method `-[node setShouldRasterizeDescendants]`. So once it's enabled on a node, there's no disabling it. This is breaking deprecated API but nobody will care.
- Adds assertions against triggering rasterization on nodes that are already loaded.
- Adds assertions against triggering rasterization on nodes whose subnodes are already loaded.
- Adds assertions against adding a loaded subnode into a rasterized tree.

So the rasterization feature remains in the Beta header, and we're not going out of our way to support it, but if we can get rid of some old code, and tamp down the API, let's keep it around. Thoughts @maicki @nguyenhuy @appleguy ?


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASDisplayNodeTests.mm:</p>
<pre><code>//
//  ASDisplayNodeTests.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/89a8d6904b4f969b1abdb8e52ac82810/raw/4ef5a32cdb90e2af4b1b8a6882fa522e13bcb1db/log.txt)

appleguy:
@Adlai-Holler :-D I never thought I'd see the day, but what an awesome surprise! Definitely planning on making use of this. Let me know if I can help, will review this weekend.

Source/ASDisplayNode+Beta.h
```diff
@@ -154,8 +154,11 @@ typedef struct {
  *
  * Note: this has nothing to do with -[CALayer shouldRasterize], which doesn't work with ASDisplayNode's asynchronous
  * rendering model.
+ *
+ * Note: You cannot add subnodes whose layers/views are already loaded to a rasterized node.
+ * Note: You cannot call this method after the receiver's layer/view is loaded.
  */
-@property (nonatomic, assign) BOOL shouldRasterizeDescendants ASDISPLAYNODE_DEPRECATED_MSG("Deprecated in version 2.2");
+- (void)setShouldRasterizeDescendants;
```
appleguy: @Adlai-Holler I think it's a good idea to make this an enable-only method. However, this name seems kinda sucky because it sounds exactly like a property that expects an argument.

How about something like -enableSubtreeRasterization, -enableRasterizingDescendants, etc? I think coming up with the right term for a subtree of subnodes could help here. .automaticallyManagesSubnodes is one reference point, but it does not apply recursively to all levels below, so Subtree might be better.

Source/ASDisplayNode.mm
```diff
@@ -1972,60 +1951,41 @@ - (void)setDisplaysAsynchronously:(BOOL)displaysAsynchronously
...
+  }
+
+  // Ensure not loaded.
+  if ([self _locked_isNodeLoaded]) {
+    ASDisplayNodeFailAssert(@"Cannot call %@ on loaded node: %@", NSStringFromSelector(_cmd), self);
+    return;
+  }
+
+  // Ensure no loaded subnodes
+  for (ASDisplayNode *subnode in _subnodes) {
```
appleguy: Is this loop enough to check all subnodes' subnodes? Should we use ASPerformBlockOnEverySubnode, or otherwise recursively call _enableRasterizingToParent or something on each?

Source/ASDisplayNode.mm
```diff
@@ -2773,15 +2739,10 @@ - (void)_insertSubnode:(ASDisplayNode *)subnode atSubnodeIndex:(NSInteger)subnod
...
-      [node enterHierarchyState:ASHierarchyStateRasterized];
-      if (node.isNodeLoaded) {
-        [node __unloadNode];
-      }
-    });
-    if (self.isInHierarchy) {
+  // If this subnode will be rasterized, enter hierarchy if needed
+  // TODO: Move this into _setSupernode: ?
+  if (isRasterized) {
+    if (self.inHierarchy) {
```
appleguy: I think it may have been important for the original code to enter the Rasterized hierarchy state before __enterHierarchy is called, because it may have invoked delegate methods that check this. I haven't looked closely though, this might not be true - just something to take a close look at.

Source/ASDisplayNode+Beta.h
Adlai-Holler: Hmmm OK! I'll change it to `enableSubtreeRasterization`.

Issue #64, opened 2017-04-23
[ASDisplayNode] Locking on isSynchronous is avoidable and sometimes deadlocks.


appleguy:
I attempted converting it to a const flag, but there are a couple places this doesn't work - mainly the setViewBlock: method (can we get rid of those / why do they exist?)

It would work for all use cases to convert it to a std::atomic.

Unless we make a change, there does exist a scenario where we can deadlock due to hitting isSynchronous.

```
(lldb) bt all
* thread #1: tid = 0x31f6, 0x0000000180b03f6c libsystem_kernel.dylib`__psynch_mutexwait + 8, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #0: 0x0000000180b03f6c libsystem_kernel.dylib`__psynch_mutexwait + 8
    frame #1: 0x0000000180bd239c libsystem_pthread.dylib`_pthread_mutex_lock_wait + 96
    frame #2: 0x0000000180bd257c libsystem_pthread.dylib`_pthread_mutex_lock_slow + 296
  * frame #3: 0x00000001015d5d48 App`ASDN::Mutex::lock(this=0x000000015925c458) + 60 at ASThread.h:198
    frame #4: 0x00000001015d5ce0 App`ASDN::Locker<ASDN::Mutex>::Locker(this=0x000000016fddf4b0, l=0x000000015925c458) + 68 at ASThread.h:84
    frame #5: 0x00000001015d2d1c App`ASDN::Locker<ASDN::Mutex>::Locker(this=0x000000016fddf4b0, l=0x000000015925c458) + 64 at ASThread.h:83
    frame #6: 0x0000000101611f48 App`::-[ASDisplayNode isSynchronous](self=0x000000015925c450, _cmd="isSynchronous") + 80 at ASDisplayNode.mm:753
    frame #7: 0x0000000101627428 App`shouldDisableNotificationsForMovingBetweenParents(from=0x000000015925c450, to=0x0000000157d2df10) + 148 at ASDisplayNode.mm:2453
[...]

PR #88, opened 2017-04-30
[ASDisplayNode] Convert isSynchronous to an Objective-C atomic BOOL.

appleguy:
This reduces lock contention, and should also fix a very rarely seen deadlock.

See https://github.com/TextureGroup/Texture/issues/64 for details.

@Adlai-Holler -- this is a case where I think a std::atomic would probably be better if we had a convenient access pattern. If we carried this pattern to other flags, the message send overhead would definitely become significant.

Although it's in the margin (most apps spend between 8-13% of all CPU time in objc_msgSend), for a core library it's great to keep that overhead as low as practical; BOOL flags seem like a possible opportunity to do something really efficient.

I'm also wondering if we could use a bitfield or something else more packed. It's not an issue for one BOOL, but I bet we'd get significant gains out of moving all the _flags to std::atomic bitfields if they were kept within 32 bits per field. We'd be able to avoid locking the mutex in many situations we do today, which would help not just CPU time but also contention / scheduler waiting.

### DIFF
--- a/Source/ASDisplayNode.h
+++ b/Source/ASDisplayNode.h
@@ -217,8 +217,7 @@ extern NSInteger const ASDefaultDrawingPriority;
  *
  * @return NO if the node wraps a _ASDisplayView, YES otherwise.
  */
-@property (nonatomic, readonly, assign, getter=isSynchronous) BOOL synchronous;
-
+@property (atomic, readonly, assign, getter=isSynchronous) BOOL synchronous;
 
 /** @name Getting view and layer */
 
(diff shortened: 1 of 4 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

appleguy:
Here's a very rough cut at what a better _atomicFlags setup might look like:

https://gist.github.com/appleguy/9b7c325b1678d27312f33abaea21d2ce

appleguy:
Closing in favor of #89

Issue #94, opened 2017-05-01
ASTextKitFontSizeAdjuster longest word


davevdveen:
The ASTextKitFontSizeAdjuster class tries to find te longest word to prevent character wrapping. This is done by finding the longest word in terms of character count. For a monospaced font this is correct but not for proportional font. For instance the word '_Thanksgiving_' counts 12 characters but is in terms of size longer than '_Tellerfleisch_' which counts 13 characters.

Also a NSAttributedString can contain multiple styles. The current implementation assumes that the style of first occurrence of the longest word is the style of the complete text. So the current implementation isn't accurate. 

I implemented a version which checks the actual width of the different words. This comes with a performance penalty, but is accurate.

PR #95, opened 2017-05-01
Fix for finding the longest word in terms of size instead of characte…

davevdveen:
…r count

Issue #94

### DIFF
--- a/Source/TextKit/ASTextKitFontSizeAdjuster.mm
+++ b/Source/TextKit/ASTextKitFontSizeAdjuster.mm
@@ -153,25 +153,29 @@
     // 2. The entire text fits in the given constrained size.
     
     NSString *str = textStorage.string;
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Adlai-Holler:
Hey thanks for the diff. We've seen thread-safety issues in the past with `boundingRectWithSize:` despite the claims of the documentation, and the performance penalty here is likely to be tremendous. 

One improvement we can make off the bat is to avoid using rangeOfString and instead use one of the enumerateSubstrings/enumerateMatches methods that will give us the range directly. Beyond that, I'll try to find time to profile this change in the near future, but of course you're right that this algorithm is correct.

nguyenhuy:
@Adlai-Holler @davevdveen  Please follow up this PR.

PR #223, opened 2017-05-02
Improve Our Handling of Subnodes

Adlai-Holler:
- Remove ASDisplayNode's conformance to NSFastEnumeration
  - We were creating a new copy of subnodes for each chunk of enumeration.
  - The enumerated collection wasn't stable – technically nodes could be deallocated during enumeration under the right circumstances.
- Instead of making a new copy of subnodes each time, just store a cached copy.
- Remove silly `dealloc` line that explicitly nils the ivar.

### DIFF
--- a/Source/Private/ASDisplayNodeInternal.h
+++ b/Source/Private/ASDisplayNodeInternal.h
@@ -127,6 +127,9 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
   ASDisplayNode * __weak _supernode;
   NSMutableArray<ASDisplayNode *> *_subnodes;
 
+  // Set this to nil whenever you modify _subnodes
+  NSArray<ASDisplayNode *> *_cachedSubnodes;
+
   ASLayoutElementStyle *_style;
   ASPrimitiveTraitCollection _primitiveTraitCollection;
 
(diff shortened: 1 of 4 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Private/ASDisplayNodeInternal.h
```diff
@@ -127,6 +127,9 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
   ASDisplayNode * __weak _supernode;
   NSMutableArray<ASDisplayNode *> *_subnodes;
 
+  // Set this to nil whenever you modify _subnodes
+  NSArray<ASDisplayNode *> *_cachedSubnodes;
```
garrettmoon: This seems like a dangerous optimization. Could we at least put in an assert that _subnodes and _cachedSubnodes have the same content in the subsides getter?

Adlai-Holler:
This is a breaking API change (removing the conformance) but I think the impact will be small because most people probably just iterate through `node.subnodes`.

PR #225, opened 2017-05-02
Move Last Few Properties from ASTableView,ASCollectionView to Node

Adlai-Holler:
Ugh this is awful, I know.

But we should do this, and then immediately land another diff that removes a ton of this junk and makes those properties and the initializers for ASTableView/ASCollectionView unavailable. This is just an interim step.


Source/ASCollectionNode.mm
```diff
@@ -275,6 +277,44 @@ - (BOOL)inverted
...
+
+- (CGFloat)leadingScreensForBatching
+{
+  if ([self pendingState]) {
+    return _pendingState.leadingScreensForBatching;
+  } else {
+    return self.view.leadingScreensForBatching;
+  }
+}
+
```
maicki: I guess the pending state would be another improvement we could build some kind of broader infrastructure around it.

PR #232, opened 2017-05-03
[ASTextNode] Move to class method of drawRect:withParameters:isCancelled:isRasterizing: for drawing

maicki:
In our effort going forward to make the framework safer to work with and to enable future optimizations like lifting up caching into a generalized way out of `ASImageNode` and `ASTextNode` specific ways this PR will remove the instance version of `drawRect:withParameters:isCancelled:isRasterizing:` in favor of using the class method.

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -52,11 +52,6 @@ static const CGFloat ASTextNodeHighlightLightOpacity = 0.11;
 static const CGFloat ASTextNodeHighlightDarkOpacity = 0.22;
 static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncationAttribute";
 
-struct ASTextNodeDrawParameter {
-  CGRect bounds;
-  UIColor *backgroundColor;
-};
-
 #pragma mark - ASTextKitRenderer
 
 @interface ASTextNodeRendererKey : NSObject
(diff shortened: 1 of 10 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f263f26018688ea38094549036ee2271/raw/2723f28ad881d7fc4633d9419b32baa1dbd9bff7/log.txt)

Source/ASTextNode.mm
```diff
@@ -471,34 +458,40 @@ - (NSArray *)exclusionPaths
...
-  return nil;
+ 
+  NSMutableDictionary *drawParameters = [[NSMutableDictionary alloc] init];
+  if (self.backgroundColor != nil) {
+    drawParameters[ASTextNodeBackgroundColorDrawParameterKey] = self.backgroundColor;
+  }
+  if (!UIEdgeInsetsEqualToEdgeInsets(_textContainerInset, UIEdgeInsetsZero)) {
+    drawParameters[ASTextNodeContainerInsetsDrawParameterKey] = [NSValue valueWithUIEdgeInsets:_textContainerInset];
+  }
+  drawParameters[ASTextNodeRendererDrawParameterKey] = [self _rendererWithBoundsSlow:self.bounds];
```
Adlai-Holler: Getting this renderer is indeed slow so we shouldn't run it on the main thread – can we do one of these two things:

- Make a bigger change where `_rendererWithBoundsSlow:` becomes a class method
- Or make this a smaller change where we return `self` from this method and use it in `+drawRect:withParameters:(self)` and update this implementation later.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/330c08f96baefa1a6a3a4e2f05ac641e/raw/9e792a9aa1e4f64ffab60ac3feb39d2211789d69/log.txt)

PR #239, opened 2017-05-05
Don't Use Associated Objects for Drawing Priority #trivial

Adlai-Holler:
No idea why this was implemented this way in the first place, but let's get rid of it.


garrettmoon:
Super weird!

PR #242, opened 2017-05-05
[Examples] Tweak Kittens app with demo of an interesting layout case.

appleguy:
(Ignore the change to ASDisplayNode - it's included here because it addresses an assertion triggered when running this code).

This layout behaves correctly / works as desired in production. However, during specifically the unconstrained / unpositioned initial pass of the stack, it encounters an infinite "parentSize" and triggers a warning.

I've thought about this quite a bit and tried a bunch of things, but haven't found a correct way to provide more specification on the layout to avoid this warning. Looking for advice on how to easily implement this. I think we might need to adjust the warning to not occur in the case of an item with alignSelf = stretch inside a stack.

Here's what the UI looks like, displaying correctly below. But the warning message here is always shown. Normally this warning is shown only when things will truly end up being zero sized, and makes sense to show for those cases. However in cases where the layout ends up being correct, it is misleading.

`2017-05-04 22:36:56.335 Sample[90753:3568319] Cannot calculate size of node: constrainedSize is infinite and node does not override -calculateSizeThatFits: or specify a preferredSize. Try setting style.preferredSize. Node: <ASDisplayNode: 0x7ff879c51620; alpha = 1; frame = (0 0; 0 0); interfaceState = { No state }; viewClass = _ASDisplayView>
`

kittennode alignselfstretch separator line

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -3934,11 +3934,13 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
     [result addObject:@{ @"alpha" : @(_pendingViewState.alpha) }];
     [result addObject:@{ @"frame" : [NSValue valueWithCGRect:_pendingViewState.frame] }];
   }
-  
-  // Check supernode so that if we are cell node we don't find self.
-  ASCellNode *cellNode = ASDisplayNodeFindFirstSupernodeOfClass(self.supernode, [ASCellNode class]);
-  if (cellNode != nil) {
-    [result addObject:@{ @"cellNode" : ASObjectDescriptionMakeTiny(cellNode) }];
+
+  if (ASDisplayNodeThreadIsMain()) {
+    // Check supernode so that if we are cell node we don't find self.
+    ASCellNode *cellNode = ASDisplayNodeFindFirstSupernodeOfClass(self.supernode, [ASCellNode class]);
+    if (cellNode != nil) {
+      [result addObject:@{ @"cellNode" : ASObjectDescriptionMakeTiny(cellNode) }];
+    }
   }
   
   [result addObject:@{ @"interfaceState" : NSStringFromASInterfaceState(self.interfaceState)} ];
--- a/examples/Kittens/Sample/KittenNode.mm
+++ b/examples/Kittens/Sample/KittenNode.mm
@@ -35,6 +35,7 @@ static const CGFloat kInnerPadding = 10.0f;
   
   ASNetworkImageNode *_imageNode;
   ASTextNode *_textNode;
+  ASDisplayNode *_separator;
   ASDisplayNode *_divider;
   BOOL _isImageEnlarged;
   BOOL _swappedTextAndImage;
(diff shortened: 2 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for KittenNode.mm:</p>
<pre><code>//
//  KittenNode.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

nguyenhuy:
Hm, if we want to do special handling for `alignSelf = .Stretch`, we also have to handle `stack.alignItems = .Stretch` which isn't possible at the display node level.

I've re-read the CSS Flexbox spec. From my understanding, measuring stretched items using an unconstrained max cross size to determine their hypothesis/intrinsic cross sizes is expected behavior. One use case in which this behavior is absolutely needed is when all items are stretched. The biggest intrinsic cross size must be computed and against which all other items are stretched.

I also checked Yoga. If I didn't miss anything, it does the same thing ([here](https://github.com/facebook/yoga/blob/master/yoga/Yoga.c#L1493)).

I sadly also can't come up with a better solution that doesn't produce the warning. All being said, if the warning isn't always correct, maybe we should consider removing it? We're correctly returning a zero size anyway. Thoughts, @maicki?

appleguy:
@nguyenhuy thanks for the detailed look at this, especially referencing the spec & other projects!

I agree - I believe this is the first case (although an important one) where this log is basically not correct. It seems reasonable for something which is of type Stretch, but has no intrinsic size, to default to 0 when given the max bounds of Inf -- but then to become the concrete size when a concrete boundary is given.

So basically, everything about the current behavior appears correct in this case, except for the error logs. It's good that we de-escalated it from an assertion, as otherwise it would be blocking this type of layout.

Should we change this to an eventLog style message, so someone looking at debug information (possibly through the future layout debugger UI) could see that a node had returned a zero-sized measurement for an Inf constraint? Open to other ideas for how to preserve any utility for developers of this log.

Probably the best action is to remove it for now, but file a task for followup considerations of how to integrate this into developer workflow (layout debugger, eventLog, logging verbosity mode, etc).

nguyenhuy:
Agreed.  Over the past few days, I've toyed with the idea of adding more event logs to each and every step of the stack algorithm, as well as to other layout specs.  That would help developer understanding the layout logics without the need to dive into the actual code which can be challenging for new eyes.  We should probably automatically print out the event logs whenever a user selects an element in the (coming) layout debugger.

In general, we should consider moving toward event logs rather than asserts or NSLog warnings. There have been complaints about them being too annoying and verbose, even blocking development in case of easy-to-hit assertions. Of course migrations should be considered on a case-by-case basic.

@appleguy Do you have time to make the change that removes this warning?

maicki:
@appleguy @nguyenhuy I agree with both of you. I can put up a PR that will remove the warning and looking into into adding logging via the event log.

nguyenhuy:
That would be great, @maicki!

maicki:
@appleguy I just merged #299 you think we can close this PR now?

appleguy:
Awesome, thank you both @maicki and @nguyenhuy ! #299 resolves this issue and therefore I'll close this out without merging.

PR #244, opened 2017-05-05
[ASImageNode] Move to class method of displayWithParameters:isCancelled: for drawing

maicki:
In our effort going forward to make the framework safer to work with and to enable future optimizations like lifting up caching into a generalized way out of ASImageNode and ASTextNode specific ways this PR will remove the instance version of displayWithParameters:isCancelled: in favor of using the class method.


Source/ASDisplayNode.mm
```diff
@@ -833,6 +831,11 @@ - (CGRect)threadSafeBounds
   return _threadSafeBounds;
```
garrettmoon: This should probably call `_locked_threadSafeBounds` instead?

Source/ASImageNode.mm
```diff
@@ -137,7 +148,7 @@ @implementation ASImageNode
   void (^_displayCompletionBlock)(BOOL canceled);
   
   // Drawing
-  ASImageNodeDrawParameters _drawParameter;
+  //ASImageNodeDrawParameters _drawParameter;
```
garrettmoon: delete?

PR #241, opened 2017-05-05
[ASDisplayNode] Adjust displayNodeRecursiveDescription to avoid touching .layer off main.

appleguy:
See this issue for a backtrace: https://github.com/TextureGroup/Texture/issues/240

Issue #240, opened 2017-05-05
Some debug methods / asserts cause issues when triggered off the main thread.


appleguy:
For example, the layout warning in calculateSizeThatFits can trigger a crash for an unrelated reason to what it is trying to inform the user about:

```
2017-05-04 22:17:03.174 Sample[88326:3536944] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'This method must be called on the main thread'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010afc8b0b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000010a6bd141 objc_exception_throw + 48
	2   CoreFoundation                      0x000000010afcccf2 +[NSException raise:format:arguments:] + 98
	3   Foundation                          0x0000000107d5d3b6 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 193
	4   Sample                              0x000000010779cf88 -[ASDisplayNode view] + 920
	5   Sample                              0x000000010779d256 -[ASDisplayNode layer] + 198
	6   Sample                              0x00000001077cdc14 ASDisplayNodeFindFirstSupernode + 84
	7   Sample                              0x00000001077cde07 ASDisplayNodeFindFirstSupernodeOfClass + 119
	8   Sample                              0x00000001077c1d6a -[ASDisplayNode propertiesForDebugDescription] + 5594
	9   Sample                              0x00000001077c2c46 -[ASDisplayNode debugDescription] + 54
	10  Sample                              0x00000001077c5079 -[ASDisplayNode(Debugging) _recursiveDescriptionHelperWithIndent:] + 105
	11  Sample                              0x00000001077c5005 -[ASDisplayNode(Debugging) displayNodeRecursiveDescription] + 53
	12  Sample                              0x00000001077a1ac3 -[ASDisplayNode calculateSizeThatFits:] + 835
	13  Sample                              0x00000001077a072e -[ASDisplayNode calculateLayoutThatFits:] + 958
	14  Sample                              0x00000001077a01b1 -[ASDisplayNode calculateLayoutThatFits:restrictedToSize:relativeToParentSize:] + 577
	15  Sample                              0x00000001077c413f -[ASDisplayNode(ASLayoutElement) layoutThatFits:parentSize:] + 2511
[...]

PR #246, opened 2017-05-05
Improve Ancestry Handling, Avoid Assertion Failure

Adlai-Holler:
This is a more complete fix for #240 and supersedes #241 .

- The assertion failure is only due to _loading_ the layer during upward traversal, which we shouldn't do. This diff prevents that from happening.
- Add public category methods `supernodes` `supernodesIncludingSelf` and `supernodeOfClass:includingSelf:`.
  - I thought about these names a lot. The word `ancestry` is more precise but I think these names are more user-friendly.
- Deprecate `ASDisplayNodeFindFirstSupernode()` and `ASDisplayNodeFindFirstSupernodeOfClass()` in favor of the new category methods.
- Remove a redundant/silly code path in `ASDisplayNode.layer` that checks `self.view.layer`. No! We either have `_layer` or we need to load it, end of story.
- Renames `-_locked_loadViewOrLayerIsLayerBacked:` to simply `-_locked_loadViewOrLayer`. We can just check the flag ourselves inside the method.

### DIFF
--- a/Source/ASDisplayNodeExtras.h
+++ b/Source/ASDisplayNodeExtras.h
@@ -125,12 +125,12 @@ extern void ASDisplayNodePerformBlockOnEverySubnode(ASDisplayNode *node, BOOL tr
 /**
  Given a display node, traverses up the layer tree hierarchy, returning the first display node that passes block.
  */
-extern ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernode(ASDisplayNode * _Nullable node, BOOL (^block)(ASDisplayNode *node)) AS_WARN_UNUSED_RESULT;
+extern ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernode(ASDisplayNode * _Nullable node, BOOL (^block)(ASDisplayNode *node)) AS_WARN_UNUSED_RESULT ASDISPLAYNODE_DEPRECATED_MSG("Use the `supernodes` property instead.");
 
 /**
  Given a display node, traverses up the layer tree hierarchy, returning the first display node of kind class.
  */
-extern __kindof ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernodeOfClass(ASDisplayNode *start, Class c) AS_WARN_UNUSED_RESULT;
+extern __kindof ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernodeOfClass(ASDisplayNode *start, Class c) AS_WARN_UNUSED_RESULT  ASDISPLAYNODE_DEPRECATED_MSG("Use the `supernodeOfClass:includingSelf:` method instead.");
 
 /**
  * Given a layer, find the window it lives in, if any.
--- a/Source/ASDisplayNodeExtras.mm
+++ b/Source/ASDisplayNodeExtras.mm
@@ -18,6 +18,7 @@
 #import <AsyncDisplayKit/ASDisplayNodeExtras.h>
 #import <AsyncDisplayKit/ASDisplayNodeInternal.h>
 #import <AsyncDisplayKit/ASDisplayNode+FrameworkPrivate.h>
+#import <AsyncDisplayKit/ASDisplayNode+Ancestry.h>
 
 #import <queue>
 #import <AsyncDisplayKit/ASRunLoopQueue.h>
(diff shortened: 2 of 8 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1698455137da4c21e6c9608d883300a5/raw/88a5ac58bac9e258f44fa7a221097a38be80e40e/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/32f7cbf0c200d4eca316f2647809db01/raw/29f66e7a3a69a6e62475fc41dc89039f19f48aa4/log.txt)

Source/ASDisplayNodeExtras.mm
```diff
@@ -138,24 +139,21 @@ extern void ASDisplayNodePerformBlockOnEverySubnode(ASDisplayNode *node, BOOL tr
...
 ASDisplayNode *ASDisplayNodeFindFirstSupernode(ASDisplayNode *node, BOOL (^block)(ASDisplayNode *node))
 {
-  CALayer *layer = node.layer;
-
-  while (layer) {
-    node = ASLayerToDisplayNode(layer);
-    if (block(node)) {
-      return node;
+  // This function has historically started with `self` but the name suggests
+  // that it wouldn't. Perhaps we should change the behavior.
```
garrettmoon: If we're deprecating it, shouldn't we leave it as is?

garrettmoon: Ah, I think you're keeping the behavior but the comment is incorrect? Same with below?

Source/ASDisplayNodeExtras.h
```diff
@@ -125,12 +125,12 @@ extern void ASDisplayNodePerformBlockOnEverySubnode(ASDisplayNode *node, BOOL tr
 /**
  Given a display node, traverses up the layer tree hierarchy, returning the first display node that passes block.
  */
-extern ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernode(ASDisplayNode * _Nullable node, BOOL (^block)(ASDisplayNode *node)) AS_WARN_UNUSED_RESULT;
+extern ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernode(ASDisplayNode * _Nullable node, BOOL (^block)(ASDisplayNode *node)) AS_WARN_UNUSED_RESULT ASDISPLAYNODE_DEPRECATED_MSG("Use the `supernodes` property instead.");
```
garrettmoon: should the msg also mention `supernodesIncludingSelf`?

Source/Base/ASDisplayNode+Ancestry.h
```diff
@@ -22,7 +22,36 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * Returns an object to enumerate the supernode ancestry of this node, starting with its supernode.
+ *
+ * For instance, you could write:
+ *   for (ASDisplayNode *node in self.supernodes) {
+ *     if ([node.backgroundColor isEqual:[UIColor blueColor]]) {
+ *       node.hidden = YES;
+ *     }
+ *   }
+ *
+ * Note: If this property is read on the main thread, the enumeration will attempt to go up
```
garrettmoon: Isn't the above example code dangerous then? It's casting to an `ASDisplayNode` when it could be either a node or a layer? I don't think the above would crash but it's a little misleading.

garrettmoon: At the very least the note should be an `@warning`

garrettmoon: Frankly this behavior actually seems pretty crazy, are we sure we want to expose this?

Adlai-Holler: No these are always going to be nodes. We are just traversing the layer hierarchy to find more nodes when there's a break. Same behavior that we use when traversing subnodes in ASPerformBlockOnEveryNode.

Source/ASDisplayNodeExtras.mm
Adlai-Holler: Yep we're keeping it the same.

Source/ASDisplayNodeExtras.h
Adlai-Holler: Eh it's a matter of taste but I don't think so – they'll see it when they search around, and these messages are better when they're shorter.

Adlai-Holler:
Hey great a CHANGELOG conflict I can't wait to not be able to merge this until the CI runs again.

PR #247, opened 2017-05-05
Assert only once we know URL has changed

garrettmoon:
We need this in case someone is setting the URL to nil before setting
the image directly. I.e. they have a node which has been used as a network
node or an image node but now they want it to be an image node.

They should then set the URL to nil and then set the image.
*if* it was used as a network node previously, this would work.
If it was previously used as an image node it would historically assert.
This addresses that.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #253, opened 2017-05-09
Simplify Override Checking, Only Do It When Assertions Are Enabled #trivial

Adlai-Holler:
A quickie.

- Remove call to `[super initialize]`. The runtime will traverse the class hierarchy for us.
- Only run the override checking code when assertions are enabled.
- Check `Method != Method` rather than `IMP != IMP` because it's faster and simpler.

PR #271, opened 2017-05-14
Fix Release Builds #trivial

Adlai-Holler:
There was a regression in #253. Most of the code in `+initialize` is only needed if assertions are enabled, but not all of it.

This diff makes a clear dilineation between code that is required to be run in all configs, and code that is only used to make assertions.


ay8s:
Thanks for this one @Adlai-Holler. Thought I was going crazy over the weekend with a TestFlight build showing no nodes. Bumping buffer to p6.27.

PR #285, opened 2017-05-17
Only call -layout and -layoutDidFinish if the node is already loaded

nguyenhuy:
After https://github.com/facebookarchive/AsyncDisplayKit/pull/3263, a layout pass can be triggered even before a node is loaded. This is to ensure the node has valid bounds to fetch its content as soon as it enters preload state, regardless of its backing store.

However, the API contracts of `-layout` and `-layoutDidFinish` state that those methods are triggered by the view's `layoutSubviews`. That means the node must be loaded by the time they are called.

This PR adds a node loaded check to `__layout` that ensures those 2 methods are called only if the condition meets.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

nguyenhuy:
@maicki Agreed...

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d80938495e722bada54781cdd768b992/raw/a46d031f5674174894ce1391e180986378f3a6d9/log.txt)

PR #287, opened 2017-05-18
Fix Alignment of Hashed Structs

Adlai-Holler:
- We initialize our structs using `{ 7, 5, 3 }`. When you do this, if the struct has padding, the memory contained in the padding bytes contains garbage.
- Our ASImageNodeContentsKey hashes a struct that has padding bytes.
- So that hash wasn't reliable, and our cache hit rate plummeted.

The solution is:
- Enable the `-Wpadded` warning around structs that we use `{}` and subsequently hash.
- Update the ASImageNodeContentsKey struct to be aligned (change BOOL to NSInteger).
- Update the documentation on ASHashBytes to point out this warning.

We could enable `-Wpadded` globally but usually we don't care too much about padding so let's not do it for now.

This patch is extremely important and should be released in a hotfix. The image cache hit rate before this patch is extremely low, at least on my mac.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/709196425f44b7e1826a9af916b46f2e/raw/602a813e0cb5a47824a5b20ec5a7253b3e0cc524/log.txt)

PR #288, opened 2017-05-18
Simplify and Optimize ASMainSerialQueue #trivial

Adlai-Holler:
I should be doing something else this evening, but nope I'm doing this.

- Only schedule onto the main queue _once per run loop_ instead of every time they add a block.
- When running blocks, grab the whole batch instead of `lock, get first, remove first, unlock, run, repeat`. Don't be silly.

### DIFF
--- a/Source/Details/ASMainSerialQueue.mm
+++ b/Source/Details/ASMainSerialQueue.mm
@@ -24,6 +24,7 @@
 {
   ASDN::Mutex _serialQueueLock;
   NSMutableArray *_blocks;
+  std::atomic<BOOL> _scheduledOnMainQueue;
 }
 
 @end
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/613f7ecc1c0ec9d0badad832e93207ed/raw/f6dac456dfae82cc8aefd812d34bd8923c31e4f5/log.txt)

Adlai-Holler:
Well I guess I'm dumb.

maicki:
@Adlai-Holler Yeah you are, why do you not know that a separator height of 0 is not equals to 42.5 😆 

`testConstrainedSizeForRowAtIndexPath, ((rect.size.height) equal to (42 + separatorHeight)) failed: ("0") is not equal to ("42.5")`

Joke beside - I think this PR is pretty cool, let me know when you fixed the CI and I will review!

PR #294, opened 2017-05-18
ASBatchFetching to not round scroll velocity #trivial

nguyenhuy:
- Doing so causes remaining time to be slightly inaccurate.


nguyenhuy:
Thanks @maicki!

PR #299, opened 2017-05-21
[ASDisplayNode] Remove assertion in calculateSizeThatFits: and log an event #trivial

### LINKED ISSUE #296: [Layout] Remove assertion in `calculateSizeThatFits:` if constrainedSize is invalid
Based on investigations within PR #242 we will remove the assertion in `calculateSizeThatFits:`.

There are certain cases were indeed it could be passed in what it first look like an invalid size, especially if flexing needs to be calculated, but what in the end is the correct behavior.

maicki:
Based on discussion in #242 we remove the assertion in `-[ASDisplayNode calculateSizeThatFits:]` and log an event

Resolves #296

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -1049,11 +1049,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 {
   __ASDisplayNodeCheckForLayoutMethodOverrides;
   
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/3187e4b7de5a62a3493bb334af1ad057/raw/6d4d6f6e39c59ba864a19c4ca01e719733e67eed/log.txt)

PR #305, opened 2017-05-24
[ASCollectionNode] Fix missing properties and layoutInspector #trivial

flovouin:
Hi guys,

Some `ASCollectionView` properties have recently been deprecated in favour of their `ASCollectionNode` counterparts, but it turns out that they're not fully functional yet:

- The `layoutInspector` from the `_pendingState` is not forwarded to the view in `didLoad`, which means that setting it only works *after* the view has been loaded, which kinda defeats the purpose of having the property at the node level.
- The `scrollDirection` and `scrollableDirections` properties are not implemented, which means that we never get the actual value from the underlying view.

These two problems are demonstrated in this sample project (along with a waterfall layout that could be useful to some Texture users :-)): https://github.com/flovouin/Texture-CollectionNodeProperties

The `layoutInspector` fix is rather straightforward. However concerning the property forwarding, I'm not entirely sure what the guidelines are. I just made sure not to load the view needlessly, but you might have a better way to do that. Please tell me and I'll update the PR ASAP if needed.

Cheers,
Flo

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -428,6 +428,16 @@
   }
 }
 
+- (ASScrollDirection)scrollDirection
+{
+  return [self isNodeLoaded] ? self.view.scrollDirection : ASScrollDirectionNone;
+}
+
+- (ASScrollDirection)scrollableDirections
+{
+  return [self isNodeLoaded] ? self.view.scrollableDirections : ASScrollDirectionNone;
+}
+
 - (ASElementMap *)visibleElements
 {
   ASDisplayNodeAssertMainThread();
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

flovouin:
Thanks! As this is not a new feature but a straightforward bug fix, I'm marking it as trivial.
Cheers,

Issue #75, opened 2017-04-27
Sometimes calling setNeedsLayout does not re-trigger layoutSpecThatFits


christianselig:
In my project, each cell represents a specific URL. In this particular case, I make an API call to figure out what the URL represents, and re-layout the cell based on the new information I receive. It's very similar in concept to `ASDKgram` and how it fetches the comments specific to a post, and re-layouts the cell upon receiving. 

It works great 90% of the time, but every so often, if I put a breakpoint at my `setNeedsLayout` call (once the cell has found out what it represents), and another at my overridden `layoutSpecThatFits` method, I see that the latter is never called, and I'm not sure why given that most of the time it acts as I would expect.

In the cell's `init`, I perform the network call, and in its completion handler I create the new node (important to note perhaps that this is the first time this node is created; it's `nil` prior to this as the node instantiated is dependent on what the API returns). I then call `setNeedsLayout` immediately, and then `layoutSpecThatFits` should pick up on the fact it's not nil and then include it in the node hierarchy.

If I log these calls, it looks like this when it works as expected:

```
layoutSpecThatFits called.
layoutSpecThatFits called.
setNeedsLayout called.
layoutSpecThatFits called.
layoutSpecThatFits called.
```

and when not:

```
layoutSpecThatFits called.
layoutSpecThatFits called.
setNeedsLayout called.
```

In the example project below, you can see by restarting the app a few times, that *normally*, the network call is received, the cell performs another layout pass, and the cell shows the new information. However, if you restart it enough (sometimes takes 5+, and is much easier to replicate on a device rather than the simulator), you'll see it won't show the image after the API call returns:

<img width="703" alt="screen shot 2017-04-26 at 7 29 51 pm" src="https://cloud.githubusercontent.com/assets/4311814/25465669/0b7ad39a-2ada-11e7-8d88-c14b4ade6a90.png">

I'm really at wit's end trying to figure out what may be causing it. It should already be in the view hierarchy at the time it's called so it's not as if `setNeedsLayout` is being called on a non-visible node. It's totally possible (even likely) that it's something I'm doing accidentally to cause this, but any light anyone could shine onto the issue would be greatly appreciated.

**Sample project:** 
[TextureTest.zip](https://github.com/TextureGroup/Texture/files/960314/TextureTest.zip)
[...]

PR #309, opened 2017-05-26
Fix non layout

garrettmoon:
Lock released between add to pend controller and modifying pend state
    
The existing design is pretty fraught with error. We should probably
rethink this but in the meantime, this fixes a bug where calling

setNeedsLayout can start failing for nodes.
    
Essentially the method ASDisplayNodeShouldApplyBridgedWriteToView has
a side effect of registering a node to apply it's pending state *if*
it doesn't currently need the pending state applied. My guess is this
was to avoid continually registering the node and this behavior actually
helped expose this bug.
    
The bug: after the node is registered for flushing it's state, several
code paths released the lock before applying that state to the pending
state object. Before it could re-obtain the lock to apply it to the pending
state, the pending state controller flushed it on the main thread.
    
On subsequent calls to setNeedsLayout, the pending state had pending state
already (from previous calls which missed the flush) and thus wasn't
registered for future flushing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Add your own contributions to the next release on the line below this with your name.
+- Fixed an issue where calls to setNeedsDisplay and setNeedsLayout would stop working on loaded nodes. [Garrett Moon](https://github.com/garrettmoon)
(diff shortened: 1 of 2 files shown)


timonus:
Huzzah!

christianselig:
YES! This fixes #75 for me. :) I'll close that one.

garrettmoon:
@christianselig sweet!

PR #315, opened 2017-05-29
Clean Up ASDisplayLayer #trivial

Adlai-Holler:
- Make `asyncDelegate` an atomic property, rather than using a mutex.
  - The weird `NS_VALID_UNTIL_END_OF_SCOPE` code has been there since the original version of this file in 2014. Originally, the async delegate was accessed using a block, so I guess this was extra safety since they didn't know what you would do in the block. Anyway, ARC will retain the delegate throughout the method call so there's no behavior change here.
- Make `displaySuspended` main-thread-only. This isn't actually a behavior/API change, since changing the value off-main in the past would trigger an assertion failure downstream (in `-setNeedsDisplay` or `-cancelAsyncDisplay`) anyway. Better to just bite the bullet.
- Other miscellaneous cleanup.

### DIFF
--- a/Source/Details/_ASDisplayLayer.mm
+++ b/Source/Details/_ASDisplayLayer.mm
@@ -28,41 +28,16 @@
 
 @implementation _ASDisplayLayer
 {
-  ASDN::Mutex _asyncDelegateLock;
-  // We can take this lock when we're setting displaySuspended and in setNeedsDisplay, so to not deadlock, this is recursive
-  ASDN::RecursiveMutex _displaySuspendedLock;
-  BOOL _displaySuspended;
   BOOL _attemptedDisplayWhileZeroSized;
 
   struct {
     BOOL delegateDidChangeBounds:1;
   } _delegateFlags;
-
-  id<_ASDisplayLayerDelegate> __weak _asyncDelegate;
 }
 
 @dynamic displaysAsynchronously;
 
-#pragma mark -
-#pragma mark Lifecycle
-
-- (instancetype)init
-{
-  if ((self = [super init])) {
-
-    self.opaque = YES;
-  }
-  return self;
-}
-
-#pragma mark -
-#pragma mark Properties
-
-- (id<_ASDisplayLayerDelegate>)asyncDelegate
-{
-  ASDN::MutexLocker l(_asyncDelegateLock);
-  return _asyncDelegate;
-}
+#pragma mark - Properties
 
 - (void)setDelegate:(id)delegate
 {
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Details/_ASDisplayLayer.mm
```diff
@@ -179,13 +138,14 @@ + (id)defaultValueForKey:(NSString *)key
 {
   if ([key isEqualToString:@"displaysAsynchronously"]) {
     return @YES;
+  } else if ([key isEqualToString:@"opaque"]) {
```
appleguy: I'm not very familiar with the semantics of defaultValueForKey when used with standard layer properties like this. That said:

- I think having the code in init is more visible for a user exploring where a potentially-unexpected default opacity is coming from.
- This implementation will perform isEqualToString: on potentially thousands of accesses to unrelated properties. Although unlikely to be a focused performance hotspot, it probably would create distributed cost.
- What do you think about removing this method entirely and setting both self.opaque = YES and self.displaysAsynchronously = YES in the init method?

Source/Details/_ASDisplayLayer.mm
```diff
@@ -209,7 +169,7 @@ - (void)display
   ASDisplayNodeAssertMainThread();
   [self _hackResetNeedsDisplay];
```
appleguy: It's time to rename this method. YYAsyncLayer does the same thing :). Maybe just _clearNeedsDisplayFlag

Source/Details/_ASDisplayLayer.mm
Adlai-Holler: @appleguy In my testing, CALayer only calls this method once per class and caches the result, and it also correctly handles overriding standard layer properties as far as I can tell. So the overall performance is probably better with this implementation, which for me overrides the point about a user understanding where an unexpected opacity came from. Thoughts?

Source/Details/_ASDisplayLayer.mm
Adlai-Holler: Agreed but too lazy to do it now 😴

Adlai-Holler:
Since CA caches the default values, the current implementation should be faster than overriding init so I'll land this in its current state. We can always revise later.

PR #316, opened 2017-05-29
Clean Up ASAsyncTransaction #trivial

Adlai-Holler:
- Make operation values simply `id` rather than `id<NSObject>`. `id<NSObject>` can be a pain to deal with since it can't be implicitly cast, plus it's ugly.
- Make transactions always call back on the main thread, rather than parameterized `callbackQueue`.
  - Previously, you couldn't pass any queue for callback other than the main queue anyway.
  - Code was inconsistent. Some places directly used main thread, other places used parameterized `callbackQueue`.
- Make operation value `atomic` since it's set from background thread, but read from main thread.
- Other misc cleanup.

### DIFF
--- a/Source/Details/Transactions/_ASAsyncTransaction.mm
+++ b/Source/Details/Transactions/_ASAsyncTransaction.mm
@@ -35,7 +35,7 @@ NSInteger const ASDefaultTransactionPriority = 0;
 @interface ASAsyncTransactionOperation : NSObject
 - (instancetype)initWithOperationCompletionBlock:(asyncdisplaykit_async_transaction_operation_completion_block_t)operationCompletionBlock;
 @property (nonatomic, copy) asyncdisplaykit_async_transaction_operation_completion_block_t operationCompletionBlock;
-@property (nonatomic, strong) id<NSObject> value; // set on bg queue by the operation block
+@property (atomic, strong) id value; // set on bg queue by the operation block
 @end
 
 @implementation ASAsyncTransactionOperation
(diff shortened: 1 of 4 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -445,7 +432,8 @@ - (void)addOperationWithBlock:(asyncdisplaykit_async_transaction_operation_block
 - (void)addCompletionBlock:(asyncdisplaykit_async_transaction_completion_block_t)completion
 {
   __weak __typeof__(self) weakSelf = self;
-  [self addOperationWithBlock:^(){return (id<NSObject>)nil;} queue:_callbackQueue completion:^(id<NSObject> value, BOOL canceled) {
+  dispatch_queue_t bsQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
+  [self addOperationWithBlock:^(){return (id)nil;} queue:bsQueue completion:^(id value, BOOL canceled) {
```
appleguy: I bet the (id)nil cast isn't needed anymore.

Source/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -445,7 +432,8 @@ - (void)addOperationWithBlock:(asyncdisplaykit_async_transaction_operation_block
 - (void)addCompletionBlock:(asyncdisplaykit_async_transaction_completion_block_t)completion
 {
   __weak __typeof__(self) weakSelf = self;
-  [self addOperationWithBlock:^(){return (id<NSObject>)nil;} queue:_callbackQueue completion:^(id<NSObject> value, BOOL canceled) {
+  dispatch_queue_t bsQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
```
appleguy: wtf does bsQueue mean? lol

Either way, shouldn't this operation be called on the main queue, if _callbackQueue used to be main? This seems like it might be a behavior change.

Source/Details/Transactions/_ASAsyncTransaction.mm
Adlai-Holler: Surprisingly it is! The compiler inferred `nullptr` instead of `(id)nil` when I just put nil. Bizarre I know =/

Source/Details/Transactions/_ASAsyncTransaction.mm
Adlai-Holler: `bsQueue` means "bullshit queue." The completion will be called on main still, but the bs "work" that we submitted should be done on another queue. If we try to do the work on the main queue, the transaction won't be safe to wait on (the group will get stuck.)

This method isn't currently used and I'd be glad to remove it. Thoughts?

PR #326, opened 2017-06-02
Fixes an issue with GIFs that would always be covered by their placeh…

garrettmoon:
…older

Also adds a subclass override for when the animated image is set.

### DIFF
--- a/Source/ASImageNode.h
+++ b/Source/ASImageNode.h
@@ -148,6 +148,9 @@ typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
  *
  * @discussion Set this to an object which conforms to ASAnimatedImageProtocol
  * to have the ASImageNode playback an animated image.
+ * @warning this method should not be overridden, it may not always be called as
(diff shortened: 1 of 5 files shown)


Source/ASImageNode.h
```diff
@@ -168,6 +171,15 @@ typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
...
 @property (nonatomic, strong) NSString *animatedImageRunLoopMode;
 
+/**
+ * @abstract Method called when animated image has been set
+ *
+ * @discussion This method is for subclasses to override so they can know if an animated image
+ * has been set on the node.
+ * @warning this method is called with the node's lock held.
+ */
+- (void)animatedImageSet:(id <ASAnimatedImageProtocol>)newAnimatedImage previousAnimatedImage:(id <ASAnimatedImageProtocol>)previousAnimatedImage;
```
Adlai-Holler: Nit: Maybe `didSetAnimatedImage:oldValue:` is a little clearer and more terse?

PR #331, opened 2017-06-06
[Performance] Convert ASLayoutElementSize to atomic #trivial

### LINKED ISSUE #318: [Performance] Convert ASLayoutElementSize to atomic. #trivial
ASLayoutElementStyle is one of the most frequently accessed objects in the framework. Sometimes it's also used on the main thread, further increasing the value of performance optimizations.

We saw very significant wins once other properties on this object were converted to atomics. A profile I just took suggests that there's meaningful opportunity in converting ASLayoutElementSize as well, of about 2% total CPU.

hannahmbanana:
Resolves #318

### DIFF
--- a/Source/Layout/ASLayoutElement.mm
+++ b/Source/Layout/ASLayoutElement.mm
@@ -124,9 +124,9 @@ do {\
 
 @implementation ASLayoutElementStyle {
   ASDN::RecursiveMutex __instanceLock__;
-  ASLayoutElementSize _size;
   ASLayoutElementStyleExtensions _extensions;
   
+  std::atomic<ASLayoutElementSize> _size;
   std::atomic<CGFloat> _spacingBefore;
   std::atomic<CGFloat> _spacingAfter;
   std::atomic<CGFloat> _flexGrow;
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Layout/ASLayoutElement.mm
```diff
@@ -180,93 +180,91 @@ - (instancetype)init
...
+  return _size.load().height;
 }
 
 - (void)setHeight:(ASDimension)height
 {
-  ASDN::MutexLocker l(__instanceLock__);
-  _size.height = height;
+  ASLayoutElementSize newSize = _size.load();
+  newSize.height = height;
+  _size.store(newSize);
```
Adlai-Holler: Unfortunately this breaks atomicity =/ Imagine:

Thread 1: read size A
Thread 2: read size A
Thread 1: write size { A, height = x }
Thread 2: write size { A, width = y }

So we end up with `size { A, width = y }`, thread 2 overwrote thread 1's change. We should have ended up with `size { A, height = x, width = y }`.

We have to do the read-update-write all in one operation.

appleguy:
@Adlai-Holler @nguyenhuy good catch. I hadn't considered the size writing problem with this.

Is it correct that the issue only occurs in the writing methods? I think this would still be worth the win if we take the lock for the writing cases and let reading use the atomic directly. Reads are drastically more common than writes, especially because of relayouts that often get performed on the main thread.

nguyenhuy:
@appleguy Sadly I don't think acquiring the lock only in writing methods will work. @Adlai-Holler Please correct me if I'm wrong.

hannahmbanana:
@nguyenhuy @Adlai-Holler If the atomic were kept and then also the lock used in the writing methods, why would this not work? It would definitely capture most of the performance win.

Thanks for explaining if you can think of an access pattern that would break this!

appleguy:
@Adlai-Holler @nguyenhuy I've taken a look, and could be missing something, but this seems safe to me...very curious to know what I might have failed to consider.

hannahmbanana:
@appleguy - I don't have merge permissions, could you land this?

nguyenhuy:
Don't worry, I got it :)

PR #334, opened 2017-06-06
[ASDataController] Add event logging for transaction queue flush duration #trivial

### LINKED ISSUE #333: [ASDataController / Developement Ease] Add event logging for transaction queue flush duration
One of the most common performance problems in a Texture app occurs when a developer makes multiple edit transactions to a collection / table right after each other, before the processing of the previous operation is finished. In this case, Texture blocks on the completion of the current edit operation, even though it would have been asynchronous if the second / subsequent operation had not been started. 

The `ASEventLog` provides the ability to capture the information on this subtle perf issue. In the `-ASDataController updateWithChangeSet:` method, we can time the duration of the dispatch group wait of the editing transaction queue and log both the time as well as the stack trace so the developer can see which edit operations are triggering performance issues.

hannahmbanana:
One of the most common performance problems in a Texture app occurs when a developer makes multiple edit transactions to a collection / table right after each other, before the processing of the previous operation is finished. In this case, Texture blocks on the completion of the current edit operation, even though it would have been asynchronous if the second / subsequent operation had not been started.

The ASEventLog provides the ability to capture the information on this subtle perf issue. In the -ASDataController updateWithChangeSet: method, we can time the duration of the dispatch group wait of the editing transaction queue and log both the time as well as the stack trace so the developer can see which edit operations are triggering performance issues.

### DIFF
--- a/Source/Details/ASDataController.mm
+++ b/Source/Details/ASDataController.mm
@@ -510,7 +510,8 @@ typedef void (^ASDataControllerCompletionBlock)(NSArray<ASCollectionElement *> *
   NSString *changeSetDescription = ASObjectDescriptionMakeTiny(changeSet);
   [changeSet addCompletionHandler:^(BOOL finished) {
     ASDataControllerLogEvent(self, @"finishedUpdate in %fms: %@",
-                             (CACurrentMediaTime() - changeSetStartTime) * 1000.0f, changeSetDescription);  }];
+                             (CACurrentMediaTime() - changeSetStartTime) * 1000.0f, changeSetDescription);
+  }];
 #endif
   
   // Attempt to mark the update completed. This is when update validation will occur inside the changeset.
--- a/Source/Details/ASDataController.mm
+++ b/Source/Details/ASDataController.mm
@@ -486,7 +486,7 @@ typedef void (^ASDataControllerCompletionBlock)(NSArray<ASCollectionElement *> *
     _initialReloadDataHasBeenCalled = YES;
   }
   
-  NSTimeInterval transactionQueueFlushDuration;
+  NSTimeInterval transactionQueueFlushDuration = 0.0f;
   {
     ASDN::ScopeTimer t(transactionQueueFlushDuration);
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Details/ASDataController.mm
```diff
@@ -485,7 +486,11 @@ - (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet
     _initialReloadDataHasBeenCalled = YES;
   }
   
-  dispatch_group_wait(_editingTransactionGroup, DISPATCH_TIME_FOREVER);
+  NSTimeInterval transactionQueueFlushDuration;
```
appleguy: Add an initialization, = 0.0f;  This would be initialized to 0 automatically if an instance variable, but in this case it's technically needed.

Source/Details/ASDataController.mm
```diff
@@ -498,7 +503,7 @@ - (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet
   [self invalidateDataSourceItemCounts];
   
   // Log events
-  ASDataControllerLogEvent(self, @"triggeredUpdate: %@", changeSet);
+  ASDataControllerLogEvent(self, @"triggeredUpdate (waited on editing queue for %@ms): %@", transactionQueueFlushDuration * 1000, changeSet);
```
appleguy: Change %@ms to %fms

hannahmbanana:
@appleguy: The event log shows fractions of a ms, while the console log timestamp shows ~100+ms. Any ideas?

hannahmbanana:
@nguyenhuy - FYI, this logging may be incorrect - I suspect it's understating the number of ms. @appleguy - did you have a chance to take a look at some test values?

PR #337, opened 2017-06-06
Small changes required by the coming layout debugger

nguyenhuy:
- `ASDisplayNode` can be told to not flatten its layout immediately but later on. The unflattened layout is also stored in a separate property. It's needed for inspecting not only display nodes but also layout specs used to compute a layout tree.
- `ASLayout` can be told to always retain its sublayout elements. This is needed especially for layout specs since they are usually not retained after an `ASLayout` is computed.

### DIFF
--- a/Source/ASDisplayNode.h
+++ b/Source/ASDisplayNode.h
@@ -554,6 +554,19 @@ extern NSInteger const ASDefaultDrawingPriority;
 @interface ASDisplayNode (Debugging) <ASDebugNameProvider>
 
 /**
+ * Whether ASDisplayNode should store their unflattened layouts. The layout can be accessed via `-unflattenedCalculatedLayout`.
+ * Flattened layouts use less memory and are faster to lookup, while unflattened ones are useful for debugging
+ * because they reserve original information.
+ *
+ * Defaults to NO.
+ */
++ (void)setShouldStoreUnflattenedLayouts:(BOOL)shouldStore;
+
++ (BOOL)shouldStoreUnflattenedLayouts;
+
+@property (nonatomic, strong, readonly, nullable) ASLayout *unflattenedCalculatedLayout;
+
+/**
  * @abstract Return a description of the node hierarchy.
  *
  * @discussion For debugging: (lldb) po [node displayNodeRecursiveDescription]
--- a/Source/ASDisplayNode.h
+++ b/Source/ASDisplayNode.h
@@ -554,14 +554,25 @@ extern NSInteger const ASDefaultDrawingPriority;
 @interface ASDisplayNode (Debugging) <ASDebugNameProvider>
 
 /**
- * Whether ASDisplayNode should store their unflattened layouts. The layout can be accessed via `-unflattenedCalculatedLayout`.
- * Flattened layouts use less memory and are faster to lookup, while unflattened ones are useful for debugging
- * because they reserve original information.
+ * Set to YES to tell all ASDisplayNode instances to store their unflattened layouts.
  *
- * Defaults to NO.
+ * The layout can be accessed via `-unflattenedCalculatedLayout`.
+ *
+ * Flattened layouts use less memory and are faster to lookup. On the other hand, unflattened layouts are useful for debugging
+ * because they preserve original information.
  */
 + (void)setShouldStoreUnflattenedLayouts:(BOOL)shouldStore;
 
+/**
+ * Whether or not ASDisplayNode instances should store their unflattened layouts. 
+ *
+ * The layout can be accessed via `-unflattenedCalculatedLayout`.
+ * 
+ * Flattened layouts use less memory and are faster to lookup. On the other hand, unflattened layouts are useful for debugging
+ * because they preserve original information.
+ *
+ * Defaults to NO.
+ */
 + (BOOL)shouldStoreUnflattenedLayouts;
 
 @property (nonatomic, strong, readonly, nullable) ASLayout *unflattenedCalculatedLayout;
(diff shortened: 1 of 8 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/c78bfa4c8547850eb6d750aede8bc00b/raw/fdc82d4c21cc8d1ed55a0a223b13639612d733c3/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/6932e56081bdcb0086667c078ec2c6ed/raw/f7e79592567c3aede1b1332dde68ebc143f04927/log.txt)

Source/ASDisplayNode.h
```diff
@@ -554,6 +554,19 @@ extern NSInteger const ASDefaultDrawingPriority;
 @interface ASDisplayNode (Debugging) <ASDebugNameProvider>
 
 /**
+ * Whether ASDisplayNode should store their unflattened layouts. The layout can be accessed via `-unflattenedCalculatedLayout`.
+ * Flattened layouts use less memory and are faster to lookup, while unflattened ones are useful for debugging
+ * because they reserve original information.
```
garrettmoon: nit: typo reserve should be preserve

Source/ASDisplayNode.h
```diff
@@ -554,6 +554,19 @@ extern NSInteger const ASDefaultDrawingPriority;
...
+ *
+ * Defaults to NO.
+ */
++ (void)setShouldStoreUnflattenedLayouts:(BOOL)shouldStore;
```
garrettmoon: Any  reason this isn't also a property?

garrettmoon: Nm, I see now that this is a class method :)

Source/ASDisplayNode.mm
```diff
@@ -3286,6 +3294,21 @@ - (UIView *)preferredFocusedView
 
 @implementation ASDisplayNode (Debugging)
 
++ (void)setShouldStoreUnflattenedLayouts:(BOOL)shouldStore
+{
+  storesUnflattenedLayouts = shouldStore;
```
garrettmoon: These don't appear to be thread safe. Should we use OSAtomic at least here? I wouldn't be surprised if setting BOOLs *is* actually atomic on most architectures but we probably shouldn't at least have a comment about that or fix it?

Source/ASDisplayNode.mm
```diff
@@ -3286,6 +3294,21 @@ - (UIView *)preferredFocusedView
...
+}
+
++ (BOOL)shouldStoreUnflattenedLayouts
+{
+  return storesUnflattenedLayouts;
+}
+
+- (ASLayout *)unflattenedCalculatedLayout
+{
+  return _unflattenedLayout;
```
garrettmoon: This needs a lock?

Source/Layout/ASLayout.mm
```diff
@@ -76,6 +76,13 @@ @implementation ASLayout
 
 @dynamic frame, type;
 
+static BOOL static_retainsSublayoutLayoutElements = NO;
```
garrettmoon: Any reason to have the prepended static here? Seems like a newish convention? Is there an existing one we can use?

Source/Layout/ASLayout.mm
```diff
@@ -76,6 +76,13 @@ @implementation ASLayout
...
+
++ (void)setShouldRetainSublayoutLayoutElements:(BOOL)shouldRetain
+{
+  static_retainsSublayoutLayoutElements = shouldRetain;
```
garrettmoon: This doesn't seem thread safe either.

Source/Layout/ASLayout.mm
nguyenhuy: That is because `ASLayout` also has an instance variable called `retainSublayoutLayoutElements` [here](https://github.com/facebookarchive/AsyncDisplayKit/blob/master/Source/Layout/ASLayout.mm#L57).

nguyenhuy:
@garrettmoon Ready for another round of review. Thanks!

Source/Layout/ASLayout.mm
```diff
@@ -76,6 +76,18 @@ @implementation ASLayout
 
 @dynamic frame, type;
 
+static std::atomic_bool static_retainsSublayoutLayoutElements = ATOMIC_VAR_INIT(NO);
```
nguyenhuy: Added `static` prefix because `ASLayout` also has an instance variable called `retainSublayoutLayoutElements ` [here](https://github.com/TextureGroup/Texture/blob/master/Source/Layout/ASLayout.mm#L64).

Source/ASDisplayNode.mm
```diff
@@ -82,6 +82,7 @@ @implementation ASDisplayNode
 @synthesize threadSafeBounds = _threadSafeBounds;
 
 static BOOL suppressesInvalidCollectionUpdateExceptions = NO;
+static std::atomic_bool storesUnflattenedLayouts = ATOMIC_VAR_INIT(NO);
```
garrettmoon: Oh man, this is way nicer than objc!

Source/Layout/ASLayout.mm
garrettmoon: 👍

nguyenhuy:
Awesome. Thanks, @garrettmoon!

PR #340, opened 2017-06-08
Fix internal Linter warnings #trivial

maicki:
Our internal Linter caught a couple of warnings that this PR will fix. Thanks for @jerrymarino for discovering it:

```
...
In file included from /Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/ASDisplayNode.h:27:
/Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/ASLayoutElement.h:75:1: warning: Using assign for an ObjC pointer type is not valid. Either use strong or weak for a zeroing reference in this case.
@property (nonatomic, assign, readonly) ASLayoutElementStyle *style;
^
...
In file included from /Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/AsyncDisplayKit.h:19:
/Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/ASDisplayNode.h:649:1: warning: Using a pointer for a scalar type is invalid.
@property (nonatomic, assign, nullable) CGColorRef shadowColor;                // default=opaque rgb black
^
/Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/ASDisplayNode.h:654:1: warning: Using a pointer for a scalar type is invalid.
@property (nonatomic, assign, nullable) CGColorRef borderColor;                // default=opaque rgb black
^
...
/Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/ASDisplayNode+Subclasses.h:125:1: warning: Using assign for an ObjC pointer type is not valid. Either use strong or weak for a zeroing reference in this case.
@property (nullable, nonatomic, readonly, assign) ASLayout *calculatedLayout;
^
...
/Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/ASImageProtocols.h:166:1: warning: block property does not copy the block - use copy attribute instead
@property (nonatomic, strong, readwrite) void (^coverImageReadyCallback)(UIImage *coverImage);
^
...
/Users/jerry/Projects/ios/Platform/include/AsyncDisplayKit/ASImageProtocols.h:212:1: warning: block property does not copy the block - use copy attribute instead
@property (nonatomic, strong, readwrite) dispatch_block_t playbackReadyCallback;
^
6 warnings generated.
```

### DIFF
--- a/Source/Details/ASImageProtocols.h
+++ b/Source/Details/ASImageProtocols.h
@@ -163,7 +163,7 @@ withDownloadIdentifier:(id)downloadIdentifier;
 /**
  @abstract A block which receives the cover image. Should be called when the objects cover image is ready.
  */
-@property (nonatomic, strong, readwrite) void (^coverImageReadyCallback)(UIImage *coverImage);
+@property (nonatomic, copy, readwrite) void (^coverImageReadyCallback)(UIImage *coverImage);
 
 /**
  @abstract Returns whether the supplied data contains a supported animated image format.
@@ -209,7 +209,7 @@ withDownloadIdentifier:(id)downloadIdentifier;
 /**
  @abstract Should be called when playback is ready.
  */
-@property (nonatomic, strong, readwrite) dispatch_block_t playbackReadyCallback;
+@property (nonatomic, readwrite) dispatch_block_t playbackReadyCallback;
 
 /**
  @abstract Return the image at a given index.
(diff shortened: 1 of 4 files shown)


garrettmoon:
Thank you @maicki !

Source/Details/ASImageProtocols.h
```diff
@@ -209,7 +209,7 @@ withDownloadIdentifier:(id)downloadIdentifier;
 /**
  @abstract Should be called when playback is ready.
  */
-@property (nonatomic, strong, readwrite) dispatch_block_t playbackReadyCallback;
```
appleguy: @maicki @garrettmoon  Great find! Any thoughts on why not put copy here explicitly? That would seem to match the style of most of the other changes; if anything, perhaps removing the readwrite specifier is an option too.

The copy semantics as a default are a bit less clear (since it's a special case) than the readwrite semantics (since it's always the case)

Issue #353, opened 2017-06-12
[ASPrimitiveTraitCollection] Convert to atomic to avoid unusual deadlock scenario.


appleguy:
In this case, we deadlock because a data controller operation on the main thread can't proceed, because the last data controller background layout operation is stuck on updating the trait collection, which is also what the main thread is doing.

Since ASPrimitiveTraitCollection is a struct, this should be solvable by using an atomic instead.

```
(lldb) bt all
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #0: 0x0000000116fac386 libsystem_kernel.dylib`semaphore_wait_trap + 10
    frame #1: 0x0000000116c50413 libdispatch.dylib`_os_semaphore_wait + 16
    frame #2: 0x0000000116c36434 libdispatch.dylib`_dispatch_group_wait_slow + 151
  * frame #3: 0x000000010c910d36 DebugHarness`::-[ASDataController _scheduleBlockOnMainSerialQueue:](self=0x000060800010be20, _cmd="_scheduleBlockOnMainSerialQueue:", block=0x000000010c9105e0) at ASDataController.mm:808
    frame #4: 0x000000010c91057e DebugHarness`::__40-[ASDataController environmentDidChange]_block_invoke((null)=<unavailable>) at ASDataController.mm:794
    frame #5: 0x000000010c9bf383 DebugHarness`::ASPerformBlockOnMainThread(block=0x000000010c9104d0)()) at ASInternalHelpers.m:69
    frame #6: 0x000000010c91046f DebugHarness`::-[ASDataController environmentDidChange](self=0x000060800010be20, _cmd="environmentDidChange") at ASDataController.mm:786
    frame #7: 0x000000010c7eb419 DebugHarness`::-[ASCollectionNode setPrimitiveTraitCollection:](self=0x00007fafd2502260, _cmd="setPrimitiveTraitCollection:", traitCollection=ASPrimitiveTraitCollection @ 0x00007fff5431bc80) at ASCollectionNode.mm:787
    frame #8: 0x000000010c8e74f0 DebugHarness`::-[ASViewController propagateNewTraitCollection:](self=0x00007fafd2507740, _cmd="propagateNewTraitCollection:", traitCollection=ASPrimitiveTraitCollection @ 0x00007fff5431c040) at ASViewController.mm:287
[...]

PR #355, opened 2017-06-12
[ASTraitCollection] Convert ASPrimitiveTraitCollection from lock to atomic.

appleguy:
This resolves a deadlock case, which might be more common with Yoga (unsure): https://github.com/TextureGroup/Texture/issues/353 .  Fortunately it also simplifies the code.

We should think carefully if there are any transactional locking cases that could cause this to be unsafe. However, I think it has fairly limited usage and is probably appropriate for an atomic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Add your own contributions to the next release on the line below this with your name.
+- [ASTraitCollection] Convert ASPrimitiveTraitCollection from lock to atomic. [Scott Goodson](https://github.com/appleguy)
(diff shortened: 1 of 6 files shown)


appleguy:
@Adlai-Holler oh wow, does this mean that ASLayoutElementStyle can also do that? I didn't know this worked, presumably operator overloading to call .store() for you.

Sure, I'll change it.

appleguy:
@garrettmoon yeah, that's a good point and something we should talk more about strategically. I think we need to do some high level data flow analysis and make a call such as flags and a small number of other properties maybe are OK to use atomics, but even that could potentially need value-capture at the beginning of locking scopes that use the flag multiple times (this is rare but can happen)

PR #372, opened 2017-06-19
Fix double-load issue with ASCollectionNode

### LINKED ISSUE #351: [ASCollectionNode] Node may load data twice during initial load.
I haven't narrowed the conditions down exactly, but I was building tests for some new collection node API and I noticed this issue. We get `numberOfSections, numberOfItemsInSection, nodeBlockForItem` twice for the entire data set on initial load.

I'll attach the unit test when I put up the PR in a minute. @nguyenhuy or I should look into this when there's time.

Adlai-Holler:
Resolves #351 
- Make `[ASCollectionView reloadData*]` unavailable, not deprecated. This is breaking API but the methods have been deprecated for going on 7 months.
  - We can't reliably tell the difference between clients calling this method, and UICollectionView itself calling this method, so we can't safely override it and change its behavior.
  - We now unconditionally forward this call to super. 
  - It's OK for clients to blindly call `reloadData` on a view that has been cast to `(UICollectionView *)` but it just won't do what they expect. Nothing will actually change.
- Modify `[ASCollectionNode reloadData*]` to do the real work.
- Update unit tests.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -710,7 +710,10 @@
 
 - (void)relayoutItems
 {
-  [self.view relayoutItems];
+  ASDisplayNodeAssertMainThread();
+  if (self.nodeLoaded) {
+  	[self.view relayoutItems];
+  }
 }
 
 - (void)beginUpdates
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -28,6 +28,7 @@
 #import <AsyncDisplayKit/ASDisplayNode+FrameworkPrivate.h>
 #import <AsyncDisplayKit/ASInternalHelpers.h>
 #import <AsyncDisplayKit/ASCellNode+Internal.h>
+#import <AsyncDisplayKit/_ASHierarchyChangeSet.h>
 #import <AsyncDisplayKit/AsyncDisplayKit+Debug.h>
 #import <AsyncDisplayKit/ASSectionContext.h>
 #import <AsyncDisplayKit/ASDataController.h>
@@ -682,9 +683,17 @@
 - (void)reloadDataWithCompletion:(void (^)())completion
 {
   ASDisplayNodeAssertMainThread();
-  if (self.nodeLoaded) {
-    [self.view reloadDataWithCompletion:completion];
+  if (!self.nodeLoaded) {
+    return;
   }
+  
+  [self performBatchUpdates:^{
+    [self.view.changeSet reloadData];
+  } completion:^(BOOL finished){
+    if (completion) {
+      completion();
+    }
+  }];
 }
 
 - (void)reloadData
@@ -692,14 +701,16 @@
   [self reloadDataWithCompletion:nil];
 }
 
+- (void)reloadDataImmediately
+{
+  ASDisplayNodeAssertMainThread();
+  [self reloadData];
+  [self waitUntilAllUpdatesAreCommitted];
+}
+
 - (void)relayoutItems
 {
   [self.view relayoutItems];
-}
-
-- (void)reloadDataImmediately
-{
-  [self.view reloadDataImmediately];
 }
 
 - (void)beginUpdates
--- a/Tests/ASCollectionViewTests.mm
+++ b/Tests/ASCollectionViewTests.mm
@@ -259,7 +259,8 @@
   [window setRootViewController:testController];
   [window makeKeyAndVisible];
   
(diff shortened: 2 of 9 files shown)


Source/ASCollectionNode.mm
```diff
@@ -682,24 +683,34 @@ - (void)waitUntilAllUpdatesAreCommitted
...
+  ASDisplayNodeAssertMainThread();
+  [self reloadData];
+  [self waitUntilAllUpdatesAreCommitted];
 }
 
-- (void)reloadDataImmediately
+- (void)relayoutItems
 {
-  [self.view reloadDataImmediately];
+  [self.view relayoutItems];
```
Adlai-Holler: I swapped the order of these two methods to group the `reloadData*` methods together.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASCollectionViewTests.mm:</p>
<pre><code>//
//  ASCollectionViewTests.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASCollectionNode.mm
```diff
@@ -682,24 +683,34 @@ - (void)waitUntilAllUpdatesAreCommitted
...
+  if (!self.nodeLoaded) {
+    return;
   }
+  
+  [self performBatchUpdates:^{
+    [self.view.changeSet reloadData];
+  } completion:^(BOOL finished){
+    if (completion) {
+      completion();
+    }
```
Adlai-Holler: Implementation copied from `ASCollectionView`.

Source/ASCollectionNode.mm
nguyenhuy: I know the original code doesn't assert main thread and check if this node is loaded, should we do these now?

Tests/ASCollectionViewTests.mm
```diff
@@ -752,7 +754,8 @@ - (void)testThatSectionContextsAreCorrectAfterReloadData
   updateValidationTestPrologue
   
   del.sectionGeneration++;
-  [cv reloadDataImmediately];
+  [cn reloadData];
+  [cn waitUntilAllUpdatesAreCommitted];
```
nguyenhuy: And here.

Tests/ASCollectionViewTests.mm
```diff
@@ -390,12 +391,13 @@ - (void)testThatCollectionNodeConformsToExpectedProtocols
...
   __unused ASCollectionView *cv = testController.collectionView;\
-  __unused ASCollectionNode *cn = testController.collectionNode;\
+  ASCollectionNode *cn = testController.collectionNode;\
   UIWindow *window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];\
   [window makeKeyAndVisible]; \
   window.rootViewController = testController;\
   \
-  [testController.collectionView reloadDataImmediately];\
+  [cn reloadData];\
+  [cn waitUntilAllUpdatesAreCommitted]; \
```
nguyenhuy: Same here, `[cn reloadDataImmediately]`?

Tests/ASCollectionViewTests.mm
```diff
@@ -259,7 +259,8 @@ - (void)testSelection
   [window setRootViewController:testController];
   [window makeKeyAndVisible];
   
-  [testController.collectionView reloadDataImmediately];
+  [testController.collectionNode reloadData];
+  [testController.collectionNode waitUntilAllUpdatesAreCommitted];
```
nguyenhuy: Shouldn't we call `reloadDataImmediately` directly?

Adlai-Holler: No because it's deprecated.

Tests/ASCollectionViewTests.mm
Adlai-Holler: No because it's deprecated.

Tests/ASCollectionViewTests.mm
Adlai-Holler: No because it's deprecated.

Tests/ASCollectionViewTests.mm
nguyenhuy: Got it 👍

Issue #303, opened 2017-05-24
Deadlock in ASVideoNode when playing video after creating


klivin:
Getting deadlock in ASVideoNode on main thread:
![screenshot 2017-05-23 16 17 55](https://cloud.githubusercontent.com/assets/200045/26380423/6ccfc502-3fd3-11e7-9993-8d8661891205.png)

` (void)didExitVisibleState
{
  [super didExitVisibleState];
  
  ASDN::MutexLocker l(__instanceLock__);
  
  if (_shouldBePlaying) {
    [self pause];
    if (_player != nil && CMTIME_IS_VALID(_player.currentTime)) {
      _lastPlaybackTime = _player.currentTime; /// CALL TO CURRENT TIME HERE DEADLOCKS
    }
    _shouldBePlaying = YES;
  }
}
`

This is happening when the video node is played using the following in a UICollectionViewCell.  Workaround is adding a delay when playing the video node after it is created.  
`
    func refreshVideoNode() {
        guard let videoUrl = userData?.promotion?.video?.urlString else {
            return
        }
    
        if videoNode == nil && playerState != .building {
            playerState = .building
            VideoCache.sharedVideoCache.cacheVideoToDisk(videoUrl, completion: { [weak self] (url, error) in
                    // Background Thread
                    guard let localUrl = url else { return }
                    self?.videoNode = self?.profileImageNode.addVideo(localUrl)
                    if self?.indexPath?.item == 0 {
                        // adding delay here alleviates deadlock in asyncdisplay kit
                        delay_main(0.5, closure: {
                            [weak self] in
                            self?.videoNode?.play()
                        })
                    } else {
                        self?.videoNode?.pause()
                    }

            })
        }
    }
`
Any tips on why this may be happening, and how to better avoid deadlocks when playing a video node?  Adding a delay is obviously not optimal...anything under .2 seconds, and it will deadlock every time.

maicki:
Related: #375

PR #375, opened 2017-06-20
[ASVideoNode] Improve locking in ASVideoNode didExitVisibleState

maicki:


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASVideoNode.mm
```diff
@@ -467,15 +467,24 @@ - (void)didExitVisibleState
...
   
-  ASDN::MutexLocker l(__instanceLock__);
-  
-  if (_shouldBePlaying) {
+  __instanceLock__.lock();
+  // We have to cache the shouldBePlaying state here as pause will set it to NO
+  BOOL shouldBePlaying = _shouldBePlaying;
+
+  if (shouldBePlaying) {
+    __instanceLock__.unlock();
```
appleguy: This seems pretty complicated - could we just leave it locked here? Otherwise we would lose transactionality and then the subtle need to cache the value of shouldBePlaying in a local variable comes into play.

I think that we will probably never get to a stage where __instanceLock__ is a non-recursive lock, so it might not be beneficial to avoid calling pause while locked.

appleguy:
I'm always impressed by your attention to detail, @maicki - caring about every corner of the codebase :).

Was there a bug you had in mind when fixing this, or is more for stylistic or implement-nonrecursive-lock reasons?

nguyenhuy:
@maicki Could you make a final call on this PR? If this PR fixes the deadlock reported in #303, I think we should somehow push this forward.

maicki:
@nguyenhuy I couldn't verify if this is fixing this issues in #303 as I couldn't reproduce it yet. We may should ask the @klivin if after applying this patch it will fix the issues for him and in this case push it through or otherwise close it for now.

maicki:
I will close this PR for now as unfortunately I don't have time at the moment to push it through and I cannot verify if this will fix #303 for sure.

PR #377, opened 2017-06-21
Add new workspaces for tests for different integrations #trivial

Adlai-Holler:
The first chunk of work toward having tests for the upcoming Diffing subspec. We need to have a (relatively) scalable way to add unit tests that apply to Texture+Integration environments.

- I tried very hard to have the different test suites in the same .xcworkspace but it's impossible to get it reliable.
- Make a new SubspecWorkspaces folder in root that contains the old ASDKListKit workspace.
  - Diffing and Yoga workspaces will go in here. @appleguy @nguyenhuy @maicki 
- Make the "Common" test files (helpers) accessible to both the base tests and the subspec tests.
- Update the ListKit tests for IGListKit 3.0 and deployment target 8.0.


ASDKListKit/Podfile
```diff
@@ -1,9 +0,0 @@
-source 'https://github.com/CocoaPods/Specs.git'
-
-platform :ios, '8.0'
-target 'ASDKListKitTests' do
-	pod 'Texture/IGListKit', :path => '..'
-	pod 'IGListKit', :git => 'https://github.com/Instagram/IGListKit', :commit => '357a28e'
-	pod 'JGMethodSwizzler', :git => 'https://github.com/JonasGessner/JGMethodSwizzler', :branch => 'master'
```
Adlai-Holler: Moved to `SubspecWorkspaces/ASDKListKit/Podfile`

Podfile
```diff
@@ -8,23 +8,23 @@ target :'AsyncDisplayKitTests' do
...
-  pod 'PINRemoteImage', '3.0.0-beta.7'
+  pod 'PINRemoteImage', '3.0.0-beta.10'
+end
 
-  #TODO CocoaPods plugin instead?
-  post_install do |installer|
-    require 'fileutils'
+#TODO CocoaPods plugin instead?
+post_install do |installer|
+  require 'fileutils'
```
Adlai-Holler: Since the post_install hook isn't associated with the target (i.e. you can only have one per file,) moved this to root.

Tests/ASDisplayNodeTests.mm
```diff
@@ -2266,12 +2266,10 @@ - (void)testThatItIsAllowedToRetrieveDebugDescriptionIncludingVCOffMainThread
...
   [vc.view addSubnode:node];
   dispatch_group_t g = dispatch_group_create();
-  dispatch_group_enter(g);
   __block NSString *debugDescription;
-  [NSThread detachNewThreadWithBlock:^{
+  dispatch_group_async(g, dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
     debugDescription = [node debugDescription];
-    dispatch_group_leave(g);
-  }];
+  });
```
Adlai-Holler: `detachNewThreadWithBlock:` is iOS10-only. I don't suspect we'll run our tests against iOS 8/9 too too much, but this was the only divergence so let's keep the option open.

Tests/ASXCTExtensions.h
```diff
@@ -1,36 +0,0 @@
...
- * XCTest extensions for CGGeometry.
- *
- * Prefer these to XCTAssert(CGRectEqualToRect(...)) because you get output
- * that tells you what went wrong.
- * Could use NSValue, but using strings makes the description messages shorter.
- */
-
-#import <XCTest/XCTestAssertionsImpl.h>
-
-#define ASXCTAssertEqualSizes(s0, s1, ...) \
```
Adlai-Holler: File was duplicated before. Now it, and all other Common test files, are referenced by the Subspec workspaces but stored in a common place.

PR #379, opened 2017-06-21
[Event Log] Log ASM flag when modify subnodes #trivial

nguyenhuy:
Since we're shying away from assertions in favor of event log, let's log the ASM flag in subnode modifications.


nguyenhuy:
Yes, we should!

nguyenhuy:
@maicki I've updated this PR to log all subnode modifications. Please have another look!

maicki:
@nguyenhuy Looks good!

PR #387, opened 2017-06-24
[ASCellNode] Remove unnecessary frame setting #trivial

nguyenhuy:
The same thing is (and should be) done in [`-layout`](https://github.com/TextureGroup/Texture/pull/387/files#diff-bf3128e7e068b8e2a4e939d402792365R110).


nguyenhuy:
Thanks, @maicki!

PR #395, opened 2017-06-27
[ASLayout] Revisit the flattening algorithm

nguyenhuy:
I'm facing an issue that blocks the first release of Weaver. In order to inspect the whole layout **element** tree that contains both display nodes and layout specs, I have to tell **all** display nodes to skip flattening their calculated layout immediately and only do so when the layout is applied (https://github.com/TextureGroup/Texture/pull/337). When it's time to flatten a layout tree, the tree contains **unflattened** calculated layouts of all nodes in the corresponding node tree. The flattening algorithm doesn't handle this use case well because it relies on `isFlattened` flag of each `ASLayout` object which is never turned on.

The fact that the algorithm relies on that flag also caused a problem with Yoga integration because a Yoga tree is always flattened, but we may forget to set the flag. If we take a step back, this flag is actually a "calculated" flag which means it is determined by other internal states and shouldn't be a stored property.

This PR aims to eliminate the flag completely. There was a performance concern last time @maicki tried to remove it (https://github.com/facebookarchive/AsyncDisplayKit/pull/3039). I took extra steps to ensure that we're now reusing `ASLayout` objects aggressively. 

Besides the flag elimination, there are some other changes:
- The algorithm no longer checks `context.layout` against `self`.
- The (DFS) algorithm stops diving down a branch when it hits a display node (hence the `if-else` clause [here](https://github.com/TextureGroup/Texture/pull/395/files#diff-c1c7995d5a6e8dcb6df558964ed77ffeR245)). This is to make sure that, even if a sublayout is not flattened, no subnode of that sublayout is ever considered a direct subnode of the root node.
- Re-enforce the requirement that a flattened layout tree is always 1-level deep. That means `rootLayout.sublayouts[i].sublayouts.count` is always 0. ([Change here](https://github.com/TextureGroup/Texture/compare/master...nguyenhuy:HNRemoveFlattenedFlag?expand=1#diff-c1c7995d5a6e8dcb6df558964ed77ffeR251))
- Add unit tests.

### DIFF
--- a/Source/Layout/ASLayout.mm
+++ b/Source/Layout/ASLayout.mm
@@ -134,7 +134,7 @@ static std::atomic_bool static_retainsSublayoutLayoutElements = ATOMIC_VAR_INIT(
     _size = size;
     
     if (ASPointIsNull(position) == NO) {
-      _position = CGPointMake(ASCeilPixelValue(position.x), ASCeilPixelValue(position.y));
+      _position = ASCeilPointValues(position);
     } else {
       _position = position;
     }
@@ -243,7 +243,7 @@ static std::atomic_bool static_retainsSublayoutLayoutElements = ATOMIC_VAR_INIT(
     const CGPoint absolutePosition = context.absolutePosition;
     
     if (ASLayoutIsDisplayNodeType(layout)) {
-      if (sublayoutsCount > 0 || CGPointEqualToPoint(absolutePosition, layout.position) == NO) {
+      if (sublayoutsCount > 0 || CGPointEqualToPoint(ASCeilPointValues(absolutePosition), layout.position) == NO) {
         // Only create a new layout if the existing one can't be reused, which means it has either some sublayouts or an invalid absolute position.
         layout = [ASLayout layoutWithLayoutElement:layout.layoutElement
                                               size:layout.size
--- a/Source/Layout/ASLayout.mm
+++ b/Source/Layout/ASLayout.mm
@@ -216,6 +216,8 @@ static std::atomic_bool static_retainsSublayoutLayoutElements = ATOMIC_VAR_INIT(
 - (ASLayout *)filteredNodeLayoutTree
 {
   if (ASLayoutIsFlattened(self)) {
+    // All flattened layouts must have this flag enabled
+    // to ensure sublayout elements are retained until the layouts are applied.
     self.retainSublayoutLayoutElements = YES;
     return self;
   }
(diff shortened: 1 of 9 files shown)


Source/Layout/ASLayout.mm
```diff
@@ -207,7 +215,10 @@ - (void)setRetainSublayoutLayoutElements:(BOOL)retainSublayoutLayoutElements
 
 - (ASLayout *)filteredNodeLayoutTree
 {
-  NSMutableArray *flattenedSublayouts = [NSMutableArray array];
+  if (ASLayoutIsFlattened(self)) {
+    self.retainSublayoutLayoutElements = YES;
```
garrettmoon: This seems like a strange side effect. Should we document why this needs to be the case?

Source/Layout/ASLayout.mm
```diff
@@ -216,28 +227,40 @@ - (ASLayout *)filteredNodeLayoutTree
...
-      if (sublayout.isFlattened == NO) {
+    ASLayout *layout = context.layout;
+    const NSArray<ASLayout *> *sublayouts = layout.sublayouts;
+    const NSUInteger sublayoutsCount = sublayouts.count;
+    const CGPoint absolutePosition = context.absolutePosition;
+    
+    if (ASLayoutIsDisplayNodeType(layout)) {
+      if (sublayoutsCount > 0 || CGPointEqualToPoint(absolutePosition, layout.position) == NO) {
+        // Only create a new layout if the existing one can't be reused, which means it has either some sublayouts or an invalid absolute position.
+        layout = [ASLayout layoutWithLayoutElement:layout.layoutElement
```
garrettmoon: Is this the "aggressive reuse"? Is this case skipped often enough in practice to make a difference? Not sure of the performance impact of removing the flag but do we have a good guess as to the consequences of landing this PR?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5d4ec8414da0f934259d40a3f18dfdbd/raw/71629bafbbdaebeb634630197bf6fdfc88878be8/log.txt)

nguyenhuy:
@garrettmoon Good points!

I run this diff against Pinterest master and got some numbers. It was able to reuse 14x `ASLayout` objects on app startup alone and easily more than 1000 after a short session. And that's fewer objects to dealloc later on.

Regarding performance impact of this diff, turns out it is 20-25% faster than the current implementation. Since each run is already very fast (less than 0.1ms) and can be on background, the performance impact on each run is not noticeable, but can reach 100ms after 7000 runs. That being said, I'm happy that this diff is not slower and any performance gain is just nice-to-have.

Before
screen shot 2017-06-28 at 6 29 46 pm

After
screen shot 2017-06-28 at 6 34 21 pm


Here is the code I used to measure: https://github.com/nguyenhuy/Texture/commit/b6ff4edc915c7541255da9e5ee15e103fbace4f0

garrettmoon:
That's great! Thanks for doing the perf checks!

nguyenhuy:
Landing this, thanks @garrettmoon and @maicki!

@Adlai-Holler Feel free to do a post-merge review here.

Issue #385, opened 2017-06-23
Couldn't find 'visible' supplementary element of kind UICollectionElementKindSectionHeader  assertion


mrvincenzo:
Getting the `Couldn't find 'visible' supplementary element of kind UICollectionElementKindSectionHeader at index path <> in map <>` assertion when animating section insertion with `performBatchUpdates` in `ASCollectionNode`. Please see https://github.com/facebookarchive/AsyncDisplayKit/pull/3158 for additional details.

Starting from scratch. First section insertion works well. Second section insertion gets the above assertion. Code snippet that reproduces the issue:

    import AsyncDisplayKit

    final class BatchUpdatesAssertion: ASViewController<ASCollectionNode> {
  
        fileprivate var timer: Timer!
        fileprivate var numberOfSections = 0

        init() {
            let layout = UICollectionViewFlowLayout()
            layout.scrollDirection = .vertical

            let node = ASCollectionNode(collectionViewLayout: layout)
            node.registerSupplementaryNode(ofKind: UICollectionElementKindSectionHeader)
      
            super.init(node: node)
      
            node.dataSource = self
            node.delegate = self
      
            timer = Timer.scheduledTimer(withTimeInterval: 2, repeats: true) {  _ in
                self.numberOfSections = self.numberOfSections + 1
                self.animateInsetion()
            }
        }
  
        required init?(coder aDecoder: NSCoder) {
            fatalError("init(coder:) has not been implemented")
        }
    }

    // MARK: - ASCollectionDataSource

    extension BatchUpdatesAssertion: ASCollectionDataSource {
        func animateInsetion() {
            node.performBatchUpdates({ 
                node.insertSections(IndexSet(integer: 0))
                node.insertItems(at: [IndexPath(item: 0, section: 0)])
            }, completion: { _ in
            })
        }
  
        func numberOfSections(in collectionNode: ASCollectionNode) -> Int {
            return numberOfSections
        }
  
        func collectionNode(_ collectionNode: ASCollectionNode, numberOfItemsInSection section: Int) -> Int {
            return 1
        }
  
        func collectionNode(_ collectionNode: ASCollectionNode, nodeForItemAt indexPath: IndexPath) -> ASCellNode {
            let node = ASTextCellNode()
            node.text = "item"
            return node
        }
  
        func collectionNode(_ collectionNode: ASCollectionNode, nodeForSupplementaryElementOfKind kind: String, at indexPath: IndexPath) -> ASCellNode {
[...]

PR #404, opened 2017-07-01
Fix issue where supplementary elements don't track section changes

### LINKED ISSUE #397: Crashes when removing sections from ASCollectionNode
Hello, there are two types of crashes when removing sections from ASCollectionNode:

- "no UICollectionViewLayoutAttributes instance for -layoutAttributesForSupplementaryElementOfKind: UICollectionElementKindSectionHeader at path ..."
- "Supplementary node should exist ..."

Here is example project demonstrating both crashes: [ASDKTest.zip](https://github.com/TextureGroup/Texture/files/1107917/ASDKTest.zip)

Adlai-Holler:
Unfortunately, since supplementary elements are keyed by section index, we have to migrate supplementaries when sections are inserted/deleted. So if you have section 0 with a header, and you insert a section at index 0, that previous header should now be at index path e.g. `{1, 0}`.

So we migrate all supplementary elements when there are section-level changes. This is meant to be a minimal diff to fix the crash, and it was reduced down from the larger change #405 which unifies our index change tracking.

Resolves #397 and (I believe) #385 @Pacek @mrvincenzo

### DIFF
--- a/Source/Private/ASMutableElementMap.mm
+++ b/Source/Private/ASMutableElementMap.mm
@@ -1,5 +1,5 @@
 //
-//  ASMutableElementMap.m
+//  ASMutableElementMap.mm
 //  Texture
 //
 //  Copyright (c) 2014-present, Facebook, Inc.  All rights reserved.
@@ -22,6 +22,7 @@
 #import <AsyncDisplayKit/ASElementMap.h>
 #import <AsyncDisplayKit/ASTwoDimensionalArrayUtils.h>
 #import <AsyncDisplayKit/NSIndexSet+ASHelpers.h>
+#import <AsyncDisplayKit/_ASHierarchyChangeSet.h>
 
 typedef NSMutableArray<NSMutableArray<ASCollectionElement *> *> ASMutableCollectionElementTwoDimensionalArray;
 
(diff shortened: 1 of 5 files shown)


Source/Private/ASMutableElementMap.mm
```diff
@@ -108,6 +102,34 @@ - (void)insertElement:(ASCollectionElement *)element atIndexPath:(NSIndexPath *)
...
+      NSInteger newSection = [changeSet newSectionForOldSection:oldSection];
+
+      if (oldSection == newSection) {
+        // Index path stayed the same, do nothing.
+      } else if (newSection == NSNotFound) {
+        // Section was deleted, remove the supplementary element.
+        [supps removeObjectForKey:oldIndexPath];
+      } else {
+        // Section index changed, move it.
+        NSIndexPath *newIndexPath = [NSIndexPath indexPathForItem:oldIndexPath.item inSection:newSection];
```
appleguy: I think in this case we may also need to do the removeObjectForKey:oldIndexPath ?

Adlai-Holler: Ah good catch!

mrvincenzo:
@Adlai-Holler I've verified that a5a9627 fixes #385. Thanks a lot!

appleguy:
In turn, thank you for filing the issue *and* verifying the fix so quickly!

That really helps increase our confidence to land and cut a release
rapidly, and enables us to move on to more great things :-)

Adlai-Holler:
Note: Unfortunately there's still a subtle bug here!

Imagine you have sections 0 1 2, you delete section 0, and during the dictionary enumeration you visit in the order 2 1 0. Visit 2: `remove:2` `update:1`, then visit 1: `remove:1` – now we've lost the header for section 1.

I'll update the diff to be a little less eager about performance and simply generate a new dictionary every time without erasing elements.

Adlai-Holler:
I've updated the diff and tested it again against the sample project from @Pacek in #397 to make sure it still works.

Pacek:
@appleguy and the rest of the Texture gang, you guys are the best, thanks for fixing this so fast.

Source/Private/ASMutableElementMap.mm
```diff
@@ -108,6 +102,37 @@ - (void)insertElement:(ASCollectionElement *)element atIndexPath:(NSIndexPath *)
...
+    return;
+  }
+
+  // For each element kind,
+  [_supplementaryElements enumerateKeysAndObjectsUsingBlock:^(NSString * _Nonnull key, NSMutableDictionary<NSIndexPath *,ASCollectionElement *> * _Nonnull supps, BOOL * _Nonnull stop) {
+    
+    // For each index path of that kind, move entries into a new dictionary.
+    // Note: it's tempting to update the dictionary in-place but because of the likely collision between old and new index paths,
+    // subtle bugs are possible. Note that this process is rare (only on section-level updates),
+    // that this work is done off-main, and that the typical supplementary element use case is just 1-per-section (header).
```
appleguy: @Adlai-Holler this is a very worthwhile comment, thanks!

PR #407, opened 2017-07-01
[ASCollectionView] Small improvements

nguyenhuy:
- `_ASCollectionReusableView` and `_ASCollectionViewCell` no longer expose getters for their collection element and cell node properties. This is to make sure that clients can only obtain elements from the visible map which is always the source of truth.
- Since the map can return `nil` for an element request, it's much safer to check and avoid adding/removing a nil pointer to an `NSArray`.
- Since we use a special way to check whether an object of kind of `_ASCollectionViewCell` or `_ASCollectionReusableView`, based on the assumption that these classes are subclass restricted, I added cast-or-return  macros in the header files, closer to where the restrictions are declared.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -952,7 +952,13 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout sizeForItemAtIndexPath:(NSIndexPath *)indexPath
 {
   ASDisplayNodeAssertMainThread();
-  ASCellNode *cell = [self nodeForItemAtIndexPath:indexPath];
+  ASCollectionElement *element = [_dataController.visibleMap elementForItemAtIndexPath:indexPath];
+  if (element == nil) {
+    ASDisplayNodeAssert(NO, @"Unexpected nil element for collectionView:layout:sizeForItemAtIndexPath: %@, %@, %ld", self, layout, section);
+    return CGSizeZero;
+  }
+
+  ASCellNode *cell = element.node;
   if (cell.shouldUseUIKitCell) {
     if ([_asyncDelegate respondsToSelector:@selector(collectionView:layout:sizeForItemAtIndexPath:)]) {
       CGSize size = [(id)_asyncDelegate collectionView:collectionView layout:collectionViewLayout sizeForItemAtIndexPath:indexPath];
@@ -960,8 +966,8 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
       return size;
     }
   }
-  ASCollectionElement *e = [_dataController.visibleMap elementForItemAtIndexPath:indexPath];
-  return [self sizeForElement:e];
+
+  return [self sizeForElement:element];
 }
 
 - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)layout referenceSizeForHeaderInSection:(NSInteger)section
@@ -971,6 +977,7 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   ASCollectionElement *element = [_dataController.visibleMap supplementaryElementOfKind:UICollectionElementKindSectionHeader
                                                                             atIndexPath:indexPath];
   if (element == nil) {
+    ASDisplayNodeAssert(NO, @"Unexpected nil element for collectionView:layout:referenceSizeForHeaderInSection: %@, %@, %ld", self, layout, section);
     return CGSizeZero;
   }
 
@@ -990,6 +997,7 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   ASCollectionElement *element = [_dataController.visibleMap supplementaryElementOfKind:UICollectionElementKindSectionFooter
                                                                             atIndexPath:indexPath];
   if (element == nil) {
+    ASDisplayNodeAssert(NO, @"Unexpected nil element for collectionView:layout:referenceSizeForFooterInSection: %@, %@, %ld", self, layout, section);
     return CGSizeZero;
   }
 
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -977,7 +977,6 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   ASCollectionElement *element = [_dataController.visibleMap supplementaryElementOfKind:UICollectionElementKindSectionHeader
                                                                             atIndexPath:indexPath];
   if (element == nil) {
-    ASDisplayNodeAssert(NO, @"Unexpected nil element for collectionView:layout:referenceSizeForHeaderInSection: %@, %@, %ld", self, layout, section);
     return CGSizeZero;
   }
 
@@ -997,7 +996,6 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   ASCollectionElement *element = [_dataController.visibleMap supplementaryElementOfKind:UICollectionElementKindSectionFooter
                                                                             atIndexPath:indexPath];
   if (element == nil) {
-    ASDisplayNodeAssert(NO, @"Unexpected nil element for collectionView:layout:referenceSizeForFooterInSection: %@, %@, %ld", self, layout, section);
     return CGSizeZero;
   }
 
@@ -1069,18 +1067,18 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     [(id <ASCollectionDelegateInterop>)_asyncDelegate collectionView:collectionView willDisplayCell:rawCell forItemAtIndexPath:indexPath];
   }
 
-  // Always add the element to the visible set even if it is backed by an UIKit / non-_ASCollection* view
-  ASCollectionElement *element = [_dataController.visibleMap elementForItemAtIndexPath:indexPath];
-  if (element) {
-    [_visibleElements addObject:element];
-  } else {
-    ASDisplayNodeAssert(NO, @"Unexpected nil element for willDisplayCell: %@, %@, %@", rawCell, self, indexPath);
-    return;
-  }
-
   _ASCollectionViewCell *cell = ASDynamicCastStrict(rawCell, _ASCollectionViewCell);
   if (cell == nil) {
     [_rangeController setNeedsUpdate];
+    return;
+  }
+
+  ASCollectionElement *element = cell.element;
+  if (element) {
+    ASDisplayNodeAssertTrue([_dataController.visibleMap elementForItemAtIndexPath:indexPath] == element);
+    [_visibleElements addObject:element];
+  } else {
+    ASDisplayNodeAssert(NO, @"Unexpected nil element for willDisplayCell: %@, %@, %@", rawCell, self, indexPath);
     return;
   }
 
@@ -1126,18 +1124,18 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     [(id <ASCollectionDelegateInterop>)_asyncDelegate collectionView:collectionView didEndDisplayingCell:rawCell forItemAtIndexPath:indexPath];
   }
 
-  // Always remove the element from the visible set even if it is backed by an UIKit / non-_ASCollection* view
-  ASCollectionElement *element = [_dataController.visibleMap elementForItemAtIndexPath:indexPath];
+  _ASCollectionViewCell *cell = ASDynamicCastStrict(rawCell, _ASCollectionViewCell);
+  if (cell == nil) {
+    [_rangeController setNeedsUpdate];
+    return;
+  }
+
+  // Retrieve the element from cell instead of visible map because at this point visible map could have been updated and no longer holds the element.
+  ASCollectionElement *element = cell.element;
   if (element) {
     [_visibleElements removeObject:element];
   } else {
     ASDisplayNodeAssert(NO, @"Unexpected nil element for didEndDisplayingCell: %@, %@, %@", rawCell, self, indexPath);
-    return;
-  }
-
-  _ASCollectionViewCell *cell = ASDynamicCastStrict(rawCell, _ASCollectionViewCell);
-  if (cell == nil) {
-    [_rangeController setNeedsUpdate];
     return;
   }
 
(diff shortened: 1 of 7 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0e2e8cea082be88f90115f8baffaa341/raw/f13820a5107939018d54f99cbc1ad333cfa19d68/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/a33c0886e018cf0e53e0be5be567eae8/raw/983f6496498e7063be4a7aa0ff7015edd3ca78e0/log.txt)

Source/ASCollectionView.mm
```diff
@@ -960,30 +960,38 @@ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollection
 {
   ASDisplayNodeAssertMainThread();
   NSIndexPath *indexPath = [NSIndexPath indexPathForItem:0 inSection:section];
-  ASCellNode *cell = [self supplementaryNodeForElementKind:UICollectionElementKindSectionHeader
-                                               atIndexPath:indexPath];
-  if (cell.shouldUseUIKitCell && _asyncDelegateFlags.interop) {
+  ASCollectionElement *element = [_dataController.visibleMap supplementaryElementOfKind:UICollectionElementKindSectionHeader
```
nguyenhuy: Because `-supplementaryNodeForElementKind:atIndexPath` internally calls `[_dataController.visibleMap supplementaryElementOfKind:atIndexPath].node`, we're literally asking for the element twice in this method. Let's directly reach out to the visible map.

Source/ASCollectionView.mm
```diff
@@ -960,30 +960,38 @@ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollection
...
+                                                                            atIndexPath:indexPath];
+  if (element == nil) {
+    return CGSizeZero;
```
nguyenhuy: @appleguy In case of interop, should we reach out to the delegate here? Note that since element is `nil`, `element.node.shouldUseUIKitCell` must be `NO` here, so this is not a breaking change.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ed0c820780ce738fb49df1303f668d61/raw/36cb34fa651ef0a9f1bb32f71d3069dab8f23121/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/daee02d4db293962ffe9157af07fae37/raw/8454910b62062978aa7b2f93138c29656853cbd1/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/cf443fe121a7a56adee589cab5042ba6/raw/51efece89f5df6c5727a2944a60e310093b69814/log.txt)

Source/Details/_ASCollectionReusableView.h
```diff
@@ -18,15 +18,41 @@
...
+/**
+ * Attempts to cast x to _ASCollectionReusableView, or nil if not possible.
+ *
+ * Since _ASCollectionReusableView is not available for subclassing (see below),
+ * comparing x's and _ASCollectionReusableView's classes is faster than calling -isKindOfClass: on x.
+ */
+#define ASCollectionReusableViewCast(x) ({ \
+  id __var = x; \
+  ((_ASCollectionReusableView *) (x.class == [_ASCollectionReusableView class] ? __var : nil)); \
+})
```
Adlai-Holler: I see that @appleguy started this pattern at line 1100, and that you've formalized it, but it strikes me as serious over-optimization, and I think we should simply remove it and use whatever is most concise at the call site. `-isKindOfClass:` is extremely fast. Here's the source code, from https://opensource.apple.com/source/objc4/objc4-709/runtime/NSObject.mm

```
- (BOOL)isKindOfClass:(Class)cls {
    for (Class tcls = [self class]; tcls; tcls = tcls->superclass) {
        if (tcls == cls) return YES;
    }
    return NO;
}
```

It only includes one method call – which is unavoidable – and otherwise just uses raw pointer comparison memory reads.

Source/Details/_ASCollectionReusableView.h
```diff
@@ -18,15 +18,41 @@
...
 
 AS_SUBCLASSING_RESTRICTED
+
 @interface _ASCollectionReusableView : UICollectionReusableView
-@property (nonatomic, strong, readonly, nullable) ASCellNode *node;
-@property (nonatomic, strong, nullable) ASCollectionElement *element;
+
 @property (nonatomic, strong, nullable) UICollectionViewLayoutAttributes *layoutAttributes;
+
+- (void)setElement:(ASCollectionElement *)element;
```
Adlai-Holler: I don't totally agree with removing the readwrite property. It's true that the visibleMap is the source of truth about what ought to be displayed, but it is useful to be able to read, from a view, what element it is currently displaying in reality. For example, to be able to tell in `collectionView:cellForItemAtIndexPath:` what element the cell is transitioning from, or to read for debugging purposes. And since these classes are underbarred, I think the risk of external clients leaning on the property is really minor.

appleguy: Do you have any theories on why we were seeing the misalignment between expected visibleMap and the cell.element?

Is it because layoutIfNeeded had not been flushed and somehow we are getting calls from UIKit about cell appearance with an invalid layout?

Source/Details/_ASCollectionReusableView.h
appleguy: Changing to isKindOfClass wouldn't simplify the code, because the macro is still needed to have the return-nil behavior. We could use ASDynamicCast, though.

I added the optimization after measuring its overhead. It has the most impact in the UICV compatibility case because we check superclasses' class all the way to NSObject as well. Overall because I would like to use ASCV everywhere, even when no ASCellNodes are being used, it is nice to minimize overhead where it does not add complexity.

If we're feeling generous for this use case, maybe moving it to neighbor ASDynamicCast (ASDynamicCastImmediate ?) would bring the complexity overhead to 0 for this file.

appleguy: Regarding the class hierarchy, note that the UICV case has subclasses in the app too. Something like (guessing at the hierarchy):

CustomCellClass -> UICollectionViewCell -> UICollectionReusableView -> UIView -> UIResponder -> NSObject

Source/ASCollectionView.mm
appleguy: @nguyenhuy this is a very good and probably important question to resolve some of the crashes I've been trying to get past. I think the answer will come from...in what case can we get here but have a nil element?

If that is expected to be possible / valid, then we probably do need to call the delegate. If it is believed to be invalid, then we should probably add an assertion in the return CGSizeZero case, and not worry about calling the delegate.

Source/ASCollectionView.mm
```diff
@@ -1005,7 +1013,7 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
     view = [self dequeueReusableSupplementaryViewOfKind:kind withReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
   }
   
-  if (_ASCollectionReusableView *reusableView = ASDynamicCast(view, _ASCollectionReusableView)) {
+  if (_ASCollectionReusableView *reusableView = ASCollectionReusableViewCast(view)) {
```
appleguy: IMO this is a very simple, efficient and appropriate check. This is some pretty hardcore infrastructure and we totally control _ASCollection* views, with confidence that they aren't going to have subclasses.

In fact, this check is simpler than using ASDynamicCast and having to manually pass _ASCollectionReusableView.

Source/ASCollectionView.mm
```diff
@@ -1086,7 +1096,7 @@ - (void)collectionView:(UICollectionView *)collectionView willDisplayCell:(UICol
   
   [_rangeController setNeedsUpdate];
   
-  if (ASSubclassOverridesSelector([ASCellNode class], [cellNode class], @selector(cellNodeVisibilityEvent:inScrollView:withCellFrame:))) {
+  if ([cell consumesCellNodeVisibilityEvents]) {
```
appleguy: This is potentially valuable if made into an API that can be implemented optionally on UIKit cells too.

Source/ASCollectionView.mm
```diff
@@ -1131,37 +1143,45 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
 
 - (void)collectionView:(UICollectionView *)collectionView willDisplaySupplementaryView:(UICollectionReusableView *)rawView forElementKind:(NSString *)elementKind atIndexPath:(NSIndexPath *)indexPath
 {
-  if (rawView.class != [_ASCollectionReusableView class]) {
+  ASCollectionReusableViewCastOrReturn(rawView, view, (void)0);
```
appleguy: This one seems slightly more magical - maybe worth just using the other cast methods and including the return nil check.

appleguy: Possible issue: do we really want to return here, before the element is added to the visibleMap?

It seems like for consistency, we should add the element to the visibleMap, even if it is backed by an ASCellNode that uses UIKit / non-_ASCollection* view

Source/ASCollectionView.mm
```diff
@@ -1131,37 +1143,45 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
...
+
+  ASCollectionElement *element = [_dataController.visibleMap supplementaryElementOfKind:elementKind atIndexPath:indexPath];
+  if (element == nil) {
     return;
   }
-  auto view = (_ASCollectionReusableView *)rawView;
-
-  [_visibleElements removeObject:view.element];
+  
+  [_visibleElements removeObject:element];
```
appleguy: Same question, should we allow removing the element even if the view is a non-_AS cellview?

Source/ASCollectionView.mm
```diff
@@ -1395,14 +1411,12 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
...
 {
-    for (_ASCollectionViewCell *collectionCell in _cellsForVisibilityUpdates) {
-        [[collectionCell node] cellNodeVisibilityEvent:ASCellNodeVisibilityEventDidEndDragging
-                                          inScrollView:scrollView
-                                         withCellFrame:collectionCell.frame];
-    }
-    if (_asyncDelegateFlags.scrollViewDidEndDragging) {
-        [_asyncDelegate scrollViewDidEndDragging:scrollView willDecelerate:decelerate];
-    }
+  for (_ASCollectionViewCell *cell in _cellsForVisibilityUpdates) {
```
appleguy: Is it valid to use this type when iterating over _cellsForVisibilityUpdates?

The gating factor to add things to this array is whether the cell consumes visibility events, but this might theoretically happen for non-_AS cells too.

Maybe we should make the _cellsForVisibilityUpdates collection typed to <id <ASCollectionViewCellVisibilityProtocol>> or something.

If this already works correctly because of the early-returns checking for the _AS classes that occur before the cells are added, then that's OK for now.

Source/Details/_ASCollectionViewCell.h
```diff
@@ -17,16 +17,49 @@
...
 AS_SUBCLASSING_RESTRICTED
 @interface _ASCollectionViewCell : UICollectionViewCell
-@property (nonatomic, strong, nullable) ASCollectionElement *element;
-@property (nonatomic, strong, readonly, nullable) ASCellNode *node;
+
+/**
+ * Whether or not this cell is interested in cell node visibility events.
+ * -cellNodeVisibilityEvent:inScrollView: should be called only if this property is YES.
+ */
+@property (nonatomic, readonly) BOOL consumesCellNodeVisibilityEvents;
```
appleguy: Did these events ever work for Supplementary views? It would be nice if they did, particularly because they use the same ASCellNode and the API is exposed there, so it would be unclear at a user level if they did not get called.

Source/ASCollectionView.mm
nguyenhuy: > I think the answer will come from...in what case can we get here but have a nil element?

I was assuming that clients that use interop can return an item/supplementary view that is not of type `_ASCollectionViewCell`/`_ASCollectionReusableView`, and so the element can be `nil` in these cases.

It turns out that, there is a race condition in `ASDataController` that can cause an element to be `nil`. We also got crashes reported at Pinterest with similar symptoms. More details here: https://github.com/TextureGroup/Texture/pull/420. Now that we've fixed the root cause, I think the `nil` checks in this PR is less of a concern, although I still want to keep them should code correctness.

Source/Details/_ASCollectionReusableView.h
nguyenhuy: hi, it sounds like we can get some benefits from this pattern. Would be great if you could share some numbers, @appleguy.

Since we already bite the bullet, and the changes in this diff actually make it a bit better, I'd vote for keeping it around. @Adlai-Holler Thoughts?

Source/Details/_ASCollectionReusableView.h
nguyenhuy: Indeed, there is a race condition in `ASDataController` that could cause the misalignment: https://github.com/TextureGroup/Texture/pull/420.

I'm gonna update the code to still retrieve the element from visible map whenever possible, but keep these properties around in case they are needed.

Source/Details/_ASCollectionReusableView.h
Adlai-Holler: @nguyenhuy The macro `ASCollectionReusableViewCast` just doesn't scale well. 

How about we add a variant of `ASDynamicCast` that uses `[self class] == [c class]`? `ASDynamicCastStrict` might be a good name.

nguyenhuy: I dig that. Will follow up tomorrow.

Source/ASCollectionView.mm
nguyenhuy: I'm gonna keep this as is as we do make sure only `_ASCollectionViewCell` are added to this collection. We'll update it to `id` when we support non-_AS cells.

Source/Details/_ASCollectionViewCell.h
nguyenhuy: No, this has never supported supplementary views. This API is in `_ASCollectionViewCell` so I think it's pretty clear that only _AS item cells are considered. We can of course extend it to work on other types of cells or supplementary views, but it is not in the scope of this PR.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/09a85f9c65756e71f69d3039a9563927/raw/d65e49fa878e260a625854c5a4a982d54b66be89/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1fa4a0026695e3b12d97bf9931101120/raw/ac240d5eb661da15709554151e7067c07851ae8c/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/82e9ecc989a151faea4100daac614bc2/raw/5d19f48b7a4539e48f9514b37f5766833902e381/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/dc3d93f2b0e9aa305dd767f236cc9bb8/raw/bf66dc4a1584a5ea278eb898384446f952bc8a76/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/15506ba483efb68b646507cb93fbde47/raw/52888547ab8d9e1f94072cd269eac11eed26c867/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1be589b4de60f9cb6f227d82167710f1/raw/86c419dcb76c1580d476ec021f93269f7db2ef53/log.txt)

Source/ASCollectionView.mm
nguyenhuy: It turns out that we can't track visibility of elements that are not backed by a non-_ASCollection* view.  That is because in `-didEndDisplaying*` methods, we can't retrieve the element from visible map since the map might have been updated and no longer holds the element. So we have to rely on `cell.element`, otherwise we risk having imbalanced add and remove calls on `_visibleElements`, i.e elements that aren't backed by a non-_ASCollection* view will stay visible forever.

nguyenhuy:
@Adlai-Holler This PR is ready for another round of review. Thanks!

Source/Details/_ASCollectionViewCell.h
```diff
@@ -17,16 +17,27 @@
...
 #import <AsyncDisplayKit/ASBaseDefines.h>
+#import <AsyncDisplayKit/ASCellNode.h>
 
-@class ASCellNode, ASCollectionElement;
+@class ASCollectionElement;
 
 NS_ASSUME_NONNULL_BEGIN
 
-AS_SUBCLASSING_RESTRICTED
+AS_SUBCLASSING_RESTRICTED // Note: ASDynamicCastStrict is used on instances of this class based on this restriction.
```
Adlai-Holler: 👍

PR #411, opened 2017-07-03
[ASDisplayNode] -didEnterPreloadState does not need to call -layoutIfNeeded #trivial

appleguy:
This was originally added for ASCollectionNode and ASTableNode preloading to work
as intended when nested inside of another ASRangeController-powered node. Indeed,
it is still necessary to trigger layout on these UIKit-backed components in order
for their own ASRangeControllers to start preparing content within them.

However, unlike the comment suggests, it is *not* necessary to do this for image
nodes. ASNetworkImageNode has only one .URL, and does not use the .bounds or
.calculatedLayout at all during loading. Even the ASMultiplexImageNode does not
use the .bounds, and the ASMapNode uses .calculatedLayout instead of .bounds.

This change has important performance benefits. In particular, it avoids
layouts that would occur on the main thread, often including text sizing,
and also can result in redundant layout passes (even during a layout pass that
triggers a node to enter the preload range, it may force its own layout early).

It would be great to test this change with Pinterest to confirm its safety, but
based on a full audit of the framework codebase, the only possibility that I
see for a regression is if app implementations of -didEnterPreloadState make
direct use of .bounds instead of .calculatedLayout (which is easy to fix).

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -2915,10 +2915,6 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
   ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__);
   [_interfaceStateDelegate didEnterPreloadState];
   
-  // Trigger a layout pass to ensure all subnodes have the correct size to preload their content.
-  // This is important for image nodes, as well as collection and table nodes.
-  [self layoutIfNeeded];
-  
   if (_methodOverrides & ASDisplayNodeMethodOverrideFetchData) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -213,10 +213,10 @@
 
 - (void)didEnterPreloadState
 {
-  // Intentionally allocate the view here so that super will trigger a layout pass on it which in turn will trigger the intial data load.
-  // We can get rid of this call later when ASDataController, ASRangeController and ASCollectionLayout can operate without the view.
-  [self view];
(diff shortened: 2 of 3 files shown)


appleguy:
Note that a layout pass will be forced when reaching the Display range anyway, via recursivelyEnsureDisplaySynchronously:NO which is called on the ASCellNode at that time. If the layout hasn't changed by the time it is Visible, then no additional work should be performed.

Source/ASDisplayNode.mm
```diff
@@ -2915,10 +2915,6 @@ - (void)didEnterPreloadState
   ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__);
   [_interfaceStateDelegate didEnterPreloadState];
   
-  // Trigger a layout pass to ensure all subnodes have the correct size to preload their content.
-  // This is important for image nodes, as well as collection and table nodes.
-  [self layoutIfNeeded];
```
maicki: If we remove `layoutIfNeeded` in this case would user with custom `ASDisplayNode` subclass that relied on the fact that the size was available via bounds / calculatedSize suddenly get an undefined behavior? 

They would have to basically do the same as `ASTableNode` and `ASCollectionNode` and explicitly allocating the view and trigger a layout if needed if they would like to have the old behavior?

appleguy: @maicki Since this sets it in a downward propagation, it wouldn't necessarily have set the bounds on self, unless the parent had done so first.

Assuming the parent had done so, then yes this could be a behavior change. However, using .calculatedSize actually will not change in behavior.

I think it is quite rare to override this method, and even more rare to use .bounds in it. One potential option could be to throw an assertion if a client DOES access .bounds during this call, and to tell them "this probably won't do what you want, use .calculatedLayout.size instead".

nguyenhuy:
Would love to have input from @Adlai-Holler before merging.

PR #428, opened 2017-07-09
Use a sentinel NSUInteger for node layout data #trivial

Adlai-Holler:
Instead of pushing an `invalidate()` call onto display node layouts, and having a dirty flag on them, use an atomic counter that represents a node's layout data state.

This makes it easier to control checkpoints and avoid accidentally using stale layout data.

### DIFF
--- a/Source/ASDisplayNode+Layout.mm
+++ b/Source/ASDisplayNode+Layout.mm
@@ -82,13 +82,11 @@
   }
 
   ASLayout *layout = nil;
-  if (_calculatedDisplayNodeLayout->isValidForConstrainedSizeParentSize(constrainedSize, parentSize)) {
+  NSUInteger version = _layoutVersion;
+  if (_calculatedDisplayNodeLayout != nullptr && _calculatedDisplayNodeLayout->isValid(constrainedSize, parentSize, version)) {
     ASDisplayNodeAssertNotNil(_calculatedDisplayNodeLayout->layout, @"-[ASDisplayNode layoutThatFits:parentSize:] _calculatedDisplayNodeLayout->layout should not be nil! %@", self);
-    // Our calculated layout is suitable for this constrainedSize, so keep using it and
-    // invalidate any pending layout that has been generated in the past.
-    _pendingDisplayNodeLayout = nullptr;
     layout = _calculatedDisplayNodeLayout->layout;
-  } else if (_pendingDisplayNodeLayout != nullptr && _pendingDisplayNodeLayout->isValidForConstrainedSizeParentSize(constrainedSize, parentSize)) {
+  } else if (_pendingDisplayNodeLayout != nullptr && _pendingDisplayNodeLayout->isValid(constrainedSize, parentSize, version)) {
     ASDisplayNodeAssertNotNil(_pendingDisplayNodeLayout->layout, @"-[ASDisplayNode layoutThatFits:parentSize:] _pendingDisplayNodeLayout->layout should not be nil! %@", self);
     layout = _pendingDisplayNodeLayout->layout;
   } else {
@@ -96,7 +94,7 @@
     layout = [self calculateLayoutThatFits:constrainedSize
                           restrictedToSize:self.style.size
                       relativeToParentSize:parentSize];
-    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, constrainedSize, parentSize);
+    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, constrainedSize, parentSize, version);
     ASDisplayNodeAssertNotNil(layout, @"-[ASDisplayNode layoutThatFits:parentSize:] newly calculated layout should not be nil! %@", self);
   }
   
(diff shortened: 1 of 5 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Private/ASDisplayNodeInternal.h
```diff
@@ -34,7 +34,6 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol _ASDisplayLayerDelegate;
 @class _ASDisplayLayer;
 @class _ASPendingState;
-@class ASSentinel;
```
Adlai-Holler: This was just a leftover – the ASSentinel class was removed in favor of atomic integers.

appleguy:
@Adlai-Holler I feel like we're starting to overuse #trivial...@garrettmoon, what do you think about making the changelog a persistent warning rather than an error? Since #trivial disables it anyway, it might actually get more changelogs written if it couldn't be disabled as a warning.

Source/ASDisplayNode+Layout.mm
```diff
@@ -82,21 +82,19 @@ - (ASLayout *)layoutThatFits:(ASSizeRange)constrainedSize parentSize:(CGSize)par
   }
 
   ASLayout *layout = nil;
-  if (_calculatedDisplayNodeLayout->isValidForConstrainedSizeParentSize(constrainedSize, parentSize)) {
+  NSUInteger version = _layoutVersion;
+  if (_calculatedDisplayNodeLayout != nullptr && _calculatedDisplayNodeLayout->isValid(constrainedSize, parentSize, version)) {
```
appleguy: I'm fairly sure that _calculated is *never* nullptr, and thus we leave it out of all conditionals. If so let's leave it out here because it adds length.

Hopefully we refactor this soon. I think it would be much better to use a raw ASLayout with some special instance variables that only get set for the root (this would be very efficient but also very clean, e.g. specified in ASLayout+Root.h as a private header).

Source/ASDisplayNode+Layout.mm
```diff
@@ -579,6 +579,7 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
 
       BOOL automaticallyManagesSubnodesDisabled = (self.automaticallyManagesSubnodes == NO);
       self.automaticallyManagesSubnodes = YES; // Temporary flag for 1.9.x
+      newLayoutVersion = _layoutVersion;
```
appleguy: Can you set this immediately after the lock is taken so it is more clear why it is separate from its definition?

Source/ASDisplayNode.mm
```diff
@@ -924,7 +920,6 @@ - (void)__layout
     // This method will confirm that the layout is up to date (and update if needed).
     // Importantly, it will also APPLY the layout to all of our subnodes if (unless parent is transitioning).
     [self _locked_measureNodeWithBoundsIfNecessary:bounds];
-    _pendingDisplayNodeLayout = nullptr;
```
appleguy: Why is this removed? Although it may be valid, I think we should revert this change if it's not necessary, as it adds at least a touch of risk that would be better to do in a diff focused on the _pending layout storage.

For example, I'm not sure if there is ever a case where the _pending is not copied to _calculated, such as if the _calculated matches the current bounds but _pending does not. In this case it would change the behavior of other calls that would use _pending later, compared to before this change.

Source/Private/ASDisplayNodeLayout.h
```diff
@@ -30,35 +30,26 @@ struct ASDisplayNodeLayout {
...
   BOOL requestedLayoutFromAbove;
-  BOOL _dirty;
+  NSUInteger version;
   
   /*
    * Create a new display node layout with
    * @param layout The layout to associate, usually returned from a call to -layoutThatFits:parentSize:
    * @param constrainedSize Constrained size used to create the layout
    * @param parentSize Parent size used to create the layout
+   * @param version The version of the layout data.
```
appleguy: In at least one of the comments we should describe what this means.. E.g. the layout is considered invalid when this version doesn't match the node's own state, which is guaranteed to occur after each call to invalidateCalculatedLayout (which is called by setNeedslayout).

I saw you touched on this with the instance variable definition, but this is just my observation of where exposition would be useful.

Source/ASDisplayNode+Layout.mm
Adlai-Holler: I made this change so that the redundancy between the first condition and the second condition is more explicit, and the fact that `_calculated` is never null isn't widely known (at least by this guy) and probably not ideal! Why should we have two ways of representing a null display node layout – literally `nullptr` vs this blank entry we create `ASDisplayNodeLayout()`? I'm sensitive to the length argument but I'm going to keep this change because I think the consistency of treatment outweighs the length.

That said I completely agree about the latter! Your suggestion is good or bundling those extra ivars up into a C-struct and storing two props in node `_pendingLayout: ASLayout, _pendingLayoutExtras: ASLayoutMeta` and mutating them together or something. Currently the use of shared_ptrs here is hard on the eyes.

appleguy:
That's a fine position, but I think we should change it everywhere in that
case. At the moment, it had been consistent that there were no checks for
null on that variable.

It sounds like we can jump ahead to a better clean up state anyway. I think
your suggestion is a good one, and that two instance variables is fine
(definitely preferred over another object wrapping just to contain a couple
scalars).

It's worth noting that with the plan to always layout from root, this path
will become even hotter on the main thread, so these efforts seem justified.

Source/ASDisplayNode.mm
Adlai-Holler: My aim here is to converge our handling of invalidation. If the _pendingDisplayNodeLayout is still valid, we should keep it around and use it freely. It may be a behavior change but I think it'll be for the better. If _pending and _calculated both have the same version, and it's the current version, then we should be comfortable arbitrarily using either one (minus size constraints) since neither has fresher layout data.

I believe that if I'm wrong, and I may well be, then the solution is to fix our understanding of layout freshness and perhaps find the right place to bump `_layoutVersion` to avoid using this pending layout again, rather than relying on this one-off (this is the only place where we explicitly clear pending layout). Thoughts? I'd also like to hear @maicki 's take on this if possible.

nguyenhuy:
LGTM. Merging!

PR #429, opened 2017-07-09
[ASDisplayNode] Allow setting stretchable contents on nodes; add bridged properties. #trivial

appleguy:
This makes it much easier to use the ASCoreAnimationExtras method which offers by
far the most efficient way to display a stretchable image.

In the future, we should move that function to UIImage+ASConvenience or another
header where it can be more easily found and enjoyed!

### DIFF
--- a/Source/ASDisplayNode.h
+++ b/Source/ASDisplayNode.h
@@ -621,28 +621,31 @@ extern NSInteger const ASDefaultDrawingPriority;
  */
 - (void)layoutIfNeeded;
 
-@property (nonatomic, strong, nullable) id contents;                           // default=nil
+@property (nonatomic, assign)           CGRect frame;                          // default=CGRectZero
+@property (nonatomic, assign)           CGRect bounds;                         // default=CGRectZero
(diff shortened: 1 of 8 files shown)


Source/ASDisplayNode.h
```diff
@@ -621,29 +621,32 @@ extern NSInteger const ASDefaultDrawingPriority;
  */
 - (void)layoutIfNeeded;
 
-@property (nonatomic, strong, nullable) id contents;                           // default=nil
+@property (nonatomic, assign)           CGRect frame;                          // default=CGRectZero
```
appleguy: We should definitely do additional cleanup in this header. However, I made a few improvements (no breaking changes of course) so that the most commonly used properties are more towards the top and more logically grouped.

Because this is still pretty far from ideal, let's not worry too much about optimizing it in this patch - but at least it will be slightly better until we get around to it.

maicki: 👍

appleguy:
@maicki thanks man :) Is there a UIKit method that does this, or you mean UIImageView?

Issue #430, opened 2017-07-09
[ASCollection] When using supplementary items, rotation may not invalidateLayout early enough.


appleguy:
The intention of the code in ```- (void)layer:(CALayer *)layer didChangeBoundsWithOldValue:(CGRect)oldBounds newValue:(CGRect)newBounds```, implemented in ASCollectionView, is to force a relayout of the ASCellNodes at the new size — and *then* have the UICV update its layout to reflect these new values.

However, there are two potential issues with this:
- In the case of this crash, the act of performing cell relayout somehow performs a batch update on the collection, which triggers it to perform a layout pass -- thus jumping ahead to a point where the layout should already be invalidated.
- Because invalidateLayout doesn't in itself trigger any actions, it may be safe to perform before the cell relayout
- Finally, because of recent changes to use layoutThatFits: rather than -calculatedSize when answering queries from UICollectionView for sizes, we may be able to completely eliminate the forced relayoutAllNodes-and-wait -- and instead rely on UICV's queries to synchronously fetch the layout for the cells it needs.
- If we did this latter option, we might still want to kick off an asynchronous (but not block-and-wait) relayout of all nodes, so that as the user then continues to scroll down, they don't experience synchronous layout stalls.

@nguyenhuy @Adlai-Holler @maicki 

```
2017-07-09 15:27:55.059 CollectionsStressBinary[17127:22633698] *** Assertion failure in -[UICollectionViewData validateLayoutInRect:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/UIKit_Sim/UIKit-3600.7.47/UICollectionViewData.m:445
2017-07-09 15:27:55.111 CollectionsStressBinary[17127:22633698] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'layout attributes for supplementary item at index path (<NSIndexPath: 0xc000000001600116> {length = 2, path = 1 - 11}) changed from <UICollectionViewLayoutAttributes: 0x6080001e7700> index path: (<NSIndexPath: 0xc000000001600116> {length = 2, path = 1 - 11}); element kind: (YTSeparatorKind); frame = (0 520; 320 17);  to <UICollectionViewLayoutAttributes: 0x6080001e6600> index path: (<NSIndexPath: 0xc000000001600116> {length = 2, path = 1 - 11}); element kind: (YTSeparatorKind); frame = (0 520; 480 17);  without invalidating the layout'
*** First throw call stack:
(
[...]

PR #431, opened 2017-07-09
[ASCollectionView] Improve performance and behavior of rotation / bounds changes.

appleguy:
See #430 for details.

Besides the unit tests, which do pass, is there any dedicated ASCollectionView thrash tester I should run? I see the ASTableViewStressTest in examples_extra/.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

appleguy:
@Adlai-Holler @nguyenhuy thanks for looking!

The most important question here: what caused this to start crashing? It could be something happening in my app, but it also might be one of the other refactors that has happened across the collection stack.

Perhaps there is a different way to fix the crash that would not create an issue for any layout implementations using -calculatedSize.

nguyenhuy:
@appleguy To answer your question, looking at the crash log, I believe there were some blocks scheduled on the main queue before the relayout, and one of them is the completion block of a change set update (`-dataController:didUpdateWithChangeSet:updates:`). And since `-[ASMainSerialQueue performBlockOnMainThread:]` flushes all scheduled blocks at once, the completion block might have been executed at an unexpected time.

Source/ASCollectionView.mm
```diff
@@ -2156,10 +2156,12 @@ - (void)layer:(CALayer *)layer didChangeBoundsWithOldValue:(CGRect)oldBounds new
...
 
   if (changedInNonScrollingDirection) {
-    [_dataController relayoutAllNodes];
-    [_dataController waitUntilAllUpdatesAreCommitted];
-    // We need to ensure the size requery is done before we update our layout.
+    // Because -invalidateLayout doesn't trigger any operations by itself, and we answer queries from UICollectionView using layoutThatFits:,
+    // we invalidate the layout before we have updated all of the cells. Any cells that the collection needs the size of immediately will get
+    // -layoutThatFits: with a new constraint, on the main thread, and synchronously calculate them. Meanwhile, relayoutAllNodes will update
+    // the layout of any remaining nodes on background threads (and fast-return for any nodes that the UICV got to first).
     [self.collectionViewLayout invalidateLayout];
```
nguyenhuy: I think invalidating the collection layout should be done "in the future", inside `-[ASDataController _relayoutAllNodes]` [here](https://github.com/TextureGroup/Texture/blob/master/Source/Details/ASDataController.mm#L811). The reason is that there might be new layouts calculated before the size change, but not still stuck on the main serial queue and so not yet applied yet. Those layouts are invalid. But calling invalidating here is too soon and doesn't invalidate them.

So it looks like we need to change `-[ASDataController relayotuAllNodes]` to accept a block that get called on the main serial queue.

Source/ASCollectionView.mm
```diff
@@ -2156,10 +2156,12 @@ - (void)layer:(CALayer *)layer didChangeBoundsWithOldValue:(CGRect)oldBounds new
   BOOL changedInNonScrollingDirection = (fixedHorizontally && newBounds.size.width != lastUsedSize.width) || (fixedVertically && newBounds.size.height != lastUsedSize.height);
 
   if (changedInNonScrollingDirection) {
-    [_dataController relayoutAllNodes];
-    [_dataController waitUntilAllUpdatesAreCommitted];
-    // We need to ensure the size requery is done before we update our layout.
+    // Because -invalidateLayout doesn't trigger any operations by itself, and we answer queries from UICollectionView using layoutThatFits:,
+    // we invalidate the layout before we have updated all of the cells. Any cells that the collection needs the size of immediately will get
+    // -layoutThatFits: with a new constraint, on the main thread, and synchronously calculate them. Meanwhile, relayoutAllNodes will update
+    // the layout of any remaining nodes on background threads (and fast-return for any nodes that the UICV got to first).
```
nguyenhuy: > Meanwhile, relayoutAllNodes will update the layout of any remaining nodes on background threads (and fast-return for any nodes that the UICV got to first).

I'm actually not aware of this behavior in the current codebase. The current implementation of `-relayoutAllNodes` processes all nodes in the UIKit index space on the main thread. That being said, I think this is the right approach that we should pursue. Are you planning to implement it shortly?

nguyenhuy:
@Adlai-Holler Sure, I'll prepare a change in Pinterest once this PR is merged.

Source/ASCollectionView.mm
appleguy: @nguyenhuy Interesting; that sounds about right. This seems like a very important issue to fix. Do you have any ideas on how we would test it properly?

I'm a bit worried because I don't have time to dig into this right now. In fact, I haven't synced with upstream for almost 2 months, and I have to focus on some deadlines for Q4 and 2018 planning.

If you would be able to take over this diff, I'd certainly appreciate it. I bet this change would greatly help the Pinterest app, especially on iPad rotation.

Source/ASCollectionView.mm
appleguy: Heh, you're right. I suppose it is necessary for it to be synchronous in the current codebase without some kind of virtualization support (exposing only a subset of the datasource to UIKit).

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/a44016d6619fe8229706671f5ded1731/raw/05205f7939d5a5ae4ba18ef85493d851a0367919/log.txt)

appleguy:
Hmm, this erroneous test failure is striking again.

ASCollectionViewTests
  testThatMultipleBatchFetchesDontHappenUnnecessarily, failed - Too many batch fetches!
  /Users/Shared/buildkite/builds/iosf-garrett-VMrmmvuqKsO4.dyn.pinadmin.com-1/pinterest/texture/Tests/ASCollectionViewTests.mm:894
  ```
    if (batchFetchCount > 1) {
      XCTFail(@"Too many batch fetches!");

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/7497d89ba03b952c3ca80fd1d8a32a5e/raw/9bdd47acdb2462de9ff0ad8470315a2b4bb7f835/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/2405841165c2f5761aeac2c60bb9f686/raw/f7049b824a87293537c32da4fe2bcd3efe805c66/log.txt)

appleguy:
@nguyenhuy we need to either land this one now, and then I can fix merge conflicts for https://github.com/TextureGroup/Texture/pull/616 — or we can land that one first and I'll fix this one.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/06971b337ab51b8b4b283a3a53c20c40/raw/e2d3fe36b670915c89237a69a413cd0ab680c06b/log.txt)

nguyenhuy:
CI [passed](https://buildkite.com/pinterest/texture/builds/1348).

Issue #438, opened 2017-07-14
[ASCollectionView] sizeForElement: may call out to the delegate for constraint while in a different index-space


appleguy:
I believe this is a regression since -calculatedSize was swapped for layoutThatFits:. To my knowledge, it isn't valid to call out to the delegate here, as the item being sized (satisfying request directly from UICollectionViewLayout) is referenced in the UIKit index space -- whereas the call being delivered to the application delegate would access the app index space.

![image](https://user-images.githubusercontent.com/565251/28193915-6d8bbb78-67f5-11e7-81bc-3e95c77c7925.png)

PR #440, opened 2017-07-14
[ASCollectionView] Add delegate bridging and index space translation for missing UICollectionViewLayout properties.

appleguy:
This is a first attempt at resolving #438.

I know there is already one similar method to do indexPath conversions, and that the #define for
accessing properties on the UICollectionViewFlowLayout should probably be moved somewhere else.

Looking for feedback on the general direction here. In particular, I'm a bit surprised that so much
has changed in how these calls occur, as I believe especially constrainedSizeForNodeAtIndexPath: is
now called with inconsistent index path spaces (but was not before this PR landed in March:
https://github.com/facebookarchive/AsyncDisplayKit/pull/3136/files)

Since the impact of mixing the spaces is fairly serious (can cause crashes), I'm also wondering if
I am misinterpreting some aspects of the code, or if maybe the crashing impact wasn't noticed yet.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/a875b8c65dc6be2c61d059456147f2ad/raw/63085b976ae5235c0c1453c868254de46243fd49/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/abdd63945f64f729d28f8546e2e50e6f/raw/677a87bb8b795d28f8279ca6e84dd5b62d3cdc4f/log.txt)

Source/ASCollectionView.mm
```diff
@@ -949,7 +955,8 @@ - (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSe
   return [_dataController.visibleMap numberOfItemsInSection:section];
 }
 
-- (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout sizeForItemAtIndexPath:(NSIndexPath *)indexPath
+- (CGSize)collectionView:(UICollectionView *)collectionView
+                  layout:(UICollectionViewLayout *)layout sizeForItemAtIndexPath:(NSIndexPath *)indexPath
```
nguyenhuy: Since you changed the parameter name from `collectionViewLayout` to `layout`, you need to update this [line](https://github.com/TextureGroup/Texture/pull/440/files#diff-8b82b1556d1293f6f9789beb236c9271R964).

Source/ASCollectionView.mm
```diff
@@ -622,10 +622,16 @@ - (CGSize)sizeForElement:(ASCollectionElement *)element
...
+  // Get the indexPath from the pendingMap, as we're about to query the delegate (and it will reference its current data source).
+  // Often this is a translation compared to the visibleMap where the element came from (UICollectionViewLayout requesting sizeForItem).
+  NSIndexPath *indexPath = [_dataController.pendingMap indexPathForElement:element];
   NSString *supplementaryKind = element.supplementaryElementKind;
-  NSIndexPath *indexPath = [_dataController.visibleMap indexPathForElement:element];
   ASSizeRange sizeRange;
-  if (supplementaryKind == nil) {
+  if (indexPath == nil) {
+    // In this case, the latest data no longer has the element, so we can't ask the delegate for its latest constrained size.
+    // The element is still visible, but will be deleted soon. We can reuse the last-known constrainedSize for the element.
```
nguyenhuy: > The element is still visible

"visible" is quite misleading here, and so does the "vibisbleMap" name. We should change it to "current" or something similar that indicates that the element is still recognized by UIKit objects.

appleguy:
@nguyenhuy I'm really glad you mentioned the NSIndexPath wrapper idea. This is exactly what I had in mind, too. We should make the data types incompatible to cross the index-space boundary!

@Adlai-Holler I'd like your feedback on this, but I think we should at least consider using a simple struct (exactly like ASIndexPath) for this purpose. Index path handling remains one of our biggest costs, including specifically main thread costs during large collection update application cycles that are the periods of time where we see frame drops.

There might be an easier way to distinguish between the index spaces using NSIndexPath (although I'm hesitant to use an associated value, even a non-object integer). That said, it would be interesting to consider what the "best" way is to represent these values if we weren't tied to NSIndexPath.

Source/ASCollectionView.mm
appleguy: Sounds good. I agree that if we clarified the terminology here, and used it consistently everywhere (and we could even start putting it in method names, or comments at the start of methods), it would be easier to ensure correctness.

Ideas for "UIKit" space:
- View
- UIKit
- Committed

Ideas for "Latest" space:
- Source
- App
- DataSource
- Latest

I feel like there must be a better, more parallel set of names. One trouble with "current" is that you could imagine that being the current UIKit state, or the current App state.

Adlai-Holler:
@appleguy @nguyenhuy I'm opposed to migrating to a struct because on 64-bit, NSIndexPath is a tagged pointer, and it's absolutely ubiquitous when dealing with UIKit.

To prevent illegal index space crossings, I believe the best solution is for us to modify our APIs not to use index paths so often. For example, this delegate method should probably be changed to `constrainedSizeForNode:` and you can read the current data-source-indexpath, supplementary element kind, and visible-indexpath safely on your end if it's relevant.

UIKit is restricted always to using NSIndexPaths because their API doesn't reify elements. We do. We have ASCollectionElement and ASCellNode which both serve as better identifiers for content than index paths, and from which we can always derive the appropriate index path in a desired space.

appleguy:
@Adlai-Holler - ok, good points. Keeping NSIndexPath where we need to use index-based references at all.

I think we are all supportive of using elements as widely as possible. However, we still need a way to keep the index spaces straight and physically prevent / assert against problems like the changing behavior of the constrainedSizeForNode: call (since this had been in the codebase for a while, it's an indication that some kind of assertion would've helped us catch it much sooner).

Another challenge with relying on elements is that some APIs just need to use indexPaths. This includes UICollectionView passthrough. Although we shouldn't let that influence the node-based API, the technical requirements of keeping the index spaces straight are quite limited because of how great the changeSet / elementMap infrastructure is, allowing translation between them. So just like the tweaks in this patch, making sure the delegate gets sane values is more a matter of verifying correctness than a tough implementation challenge.

Any feedback on this? Would an assertion to guard our assumptions be useful?

Source/ASCollectionView.mm
```diff
@@ -970,45 +979,118 @@ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollection
...
 }
 
-- (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)layout referenceSizeForHeaderInSection:(NSInteger)section
+- (NSIndexPath *)pendingIndexPathForSection:(NSInteger)section
+{
+  // NOTE: For now, we don't translate length = 1 NSIndexPaths between spaces. Use the 0th item until then.
+  NSIndexPath *sectionIndexPath = [NSIndexPath indexPathForItem:0 inSection:section];
+  ASCollectionElement *element = [_dataController.visibleMap elementForItemAtIndexPath:sectionIndexPath];
+  NSIndexPath *pendingIndexPath = element ? [_dataController.pendingMap indexPathForElement:element] : nil;
+  return pendingIndexPath;
```
Adlai-Holler: Should this be `- (NSInteger)dataSourceSectionForSection:(NSInteger)section`? I could see not-checking-NSNotFound on the caller side being a problem. Not checking nil is roughly as likely to occur but slightly less likely to cause issues immediately.

Currently `ASSection` objects are only created if the user implements the optional `contextForSection:` API. We should instead create `ASSections` unconditionally and use them as identifiers, similar to ASCollectionElement. That way looking up the indexes of sections between maps is easy!!

Source/ASCollectionView.mm
Adlai-Holler: `UIDataSourceTranslating` uses "Presentation." I kind of like that name.

nguyenhuy: "DataSource" and "Presentation", or "Processed" and "Presenting"?

nguyenhuy:
Regarding preventing index space crossings, although I totally understand the reasons behind, I'm not so keen on changing the API. For the time being, adding assertions sounds fine, as long as those assertions always trigger, without requiring race conditions (e.g different visible and pending maps) to occur beforehand.

appleguy:
Model and Presentation could be good! It would mirror CALayer.

appleguy:
@nguyenhuy @Adlai-Holler I found another method that needs index-space conversion (layout:sizeForItem: itself). I have a fix for this locally, but haven't updated this patch yet.

Thanks for the feedback here. I need to focus on fixing locally to address a release in flight. Feel free to continue landing diffs (although I would appreciate if we can hold off for a few days on any non-functional changes such as naming updates) and I will check which changes are still needed once rebasing.

appleguy:
@nguyenhuy @Adlai-Holler I worked on this throughout the weekend, and it's pretty heavily tested now. If it passes your review, feel free to merge it. Thanks for your help!

Source/ASCollectionView.mm
```diff
@@ -622,15 +622,17 @@ - (CGSize)sizeForElement:(ASCollectionElement *)element
...
+  BOOL useUIKitCell = node.shouldUseUIKitCell;
+  if (useUIKitCell) {
+    // In this case, we should use the exact value that was stashed earlier by calling sizeForItem:, referenceSizeFor*, etc.
+    // Although the node would use the preferredSize in layoutThatFits, we can skip this because there's no constrainedSize.
+    ASDisplayNodeAssert([node.superclass isSubclassOfClass:[ASCellNode class]] == NO,
+                        @"Placeholder cells for UIKit passthrough should be generic ASCellNodes: %@", node);
+    return node.style.preferredSize;
   } else {
-    sizeRange = [self dataController:_dataController constrainedSizeForSupplementaryNodeOfKind:supplementaryKind atIndexPath:indexPath];
+    return [node layoutThatFits:element.constrainedSize].size;
```
nguyenhuy: Ok, so we no longer reach out to the delegate for a size range but rely on `element.constrainedSize`. This seems like  a simpler approach 👍  

It also means once a size range is returned by the delegate, it's no longer mutable by clients. They have to reload the element to change it, or request for a new API which simply exposes `-[ASDataController relayoutNodes:nodesSizeChanged:]`. Until then, let's get this rolling.

PR #455, opened 2017-07-18
[ASDisplayNode] Fix infinite layout loop

nguyenhuy:
The problem was introduced by #428. It will occur if a node does either of the followings:
1. Keeps using a stale pending layout over a calculated layout
2. Doesn't update the next layout's version after calling `_setNeedsLayoutFromAbove`.

### DIFF
--- a/Source/ASDisplayNode+Layout.mm
+++ b/Source/ASDisplayNode+Layout.mm
@@ -302,10 +302,10 @@ ASPrimitiveTraitCollectionDeprecatedImplementation
   }
   
   CGSize boundsSizeForLayout = ASCeilSizeValues(bounds.size);
-  
+
   // Prefer _pendingDisplayNodeLayout over _calculatedDisplayNodeLayout (if exists, it's the newest)
   // If there is no _pending, check if _calculated is valid to reuse (avoiding recalculation below).
-  if (_pendingDisplayNodeLayout == nullptr) {
+  if (_pendingDisplayNodeLayout == nullptr || _pendingDisplayNodeLayout->version < _layoutVersion) {
     if (_calculatedDisplayNodeLayout->version >= _layoutVersion
         && (_calculatedDisplayNodeLayout->requestedLayoutFromAbove == YES
             || CGSizeEqualToSize(_calculatedDisplayNodeLayout->layout.size, boundsSizeForLayout))) {
@@ -352,8 +352,10 @@ ASPrimitiveTraitCollectionDeprecatedImplementation
     ASLayout *layout = [self calculateLayoutThatFits:constrainedSize
                                     restrictedToSize:self.style.size
                                 relativeToParentSize:boundsSizeForLayout];
-    
     nextLayout = std::make_shared<ASDisplayNodeLayout>(layout, constrainedSize, boundsSizeForLayout, version);
+    // Now that the constrained size of pending layout might have been reused, the layout is useless
+    // Release it and any orphaned subnodes it retains
+    _pendingDisplayNodeLayout = nullptr;
   }
   
   if (didCreateNewContext) {
@@ -373,6 +375,9 @@ ASPrimitiveTraitCollectionDeprecatedImplementation
     // particular ASLayout object, and shouldn't loop asking again unless we have a different ASLayout.
     nextLayout->requestedLayoutFromAbove = YES;
     [self _setNeedsLayoutFromAbove];
+    // Update the layout's version here because _setNeedsLayoutFromAbove calls __setNeedsLayout which in turn increases _layoutVersion
+    // Failing to do this will cause the layout to be invalid immediately 
+    nextLayout->version = _layoutVersion;
   }
 
   // Prepare to transition to nextLayout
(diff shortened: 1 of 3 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/09a91e3435a065ab8dacb50c8a8bafe7/raw/dfb68e245b4730ac99fd7662a04806f1a25a5fb2/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

nguyenhuy:
@Adlai-Holler Yes, it indeed sounds like a cleaner approach. However, there are a couple of problems with it:
- Calling `-setNeedsLayout` on the layer is only true for the node that originates the call. Since `_setNeedsLayoutFromAbove` is recursive, all other higher up nodes need to invalidate their layouts, so `[self setNeedsLayout]` is needed. This can be solved by passing the original node as a parameter and compare against `self`.
- Another bigger problems is that, some classes override `-setNeedsLayout` to add custom logics. Metadata node for example needs to reset its animation flags as soon as the current layout is invalid. Not calling `[self setNeedsLayout]` in this case is an API breaking change.

appleguy:
Should we revert the layoutVersion change for now? It's definitely necessary (currently) to call setNeedsLayout while ascending the tree, so we should be careful to align any changes with the current expectations of the code.

Source/ASDisplayNode+Layout.mm
```diff
@@ -352,8 +352,10 @@ - (void)_locked_measureNodeWithBoundsIfNecessary:(CGRect)bounds
     ASLayout *layout = [self calculateLayoutThatFits:constrainedSize
                                     restrictedToSize:self.style.size
                                 relativeToParentSize:boundsSizeForLayout];
-    
     nextLayout = std::make_shared<ASDisplayNodeLayout>(layout, constrainedSize, boundsSizeForLayout, version);
+    // Now that the constrained size of pending layout might have been reused, the layout is useless
+    // Release it and any orphaned subnodes it retains
+    _pendingDisplayNodeLayout = nullptr;
```
appleguy: Does this effectively replace the line I commented on here? https://github.com/TextureGroup/Texture/pull/428/files#r126296449

If so, we should probably replace the line at that same location. I think it should not have been removed originally.

nguyenhuy: Yes, more or less. The main difference is that here only stale or inapplicable pending layouts are cleared, so it can happen that at a given time, both calculated and pending layouts point to a valid layout (and thus can be interchangeable). With the other original line, all pending layouts are cleared, including valid/applied ones.

Adlai-Holler: Why is this line necessary @nguyenhuy ?

nguyenhuy: Either of this line or the new layout version check [above](https://github.com/TextureGroup/Texture/pull/455/files#diff-71ea742a29543cb7e08c3dffe2a2508eR308) fixes part of the problem. The reason I included it is that at this point, the pending layout can be stale ([here](https://github.com/TextureGroup/Texture/pull/455/files#diff-71ea742a29543cb7e08c3dffe2a2508eR338)) and it's never going to be applied. And yet, it still holds onto orphaned subnodes that were either never inserted or removed by newer layouts (check out `-[ASLayout retainSublayoutLayoutElements]`).

nguyenhuy: It is true that in case the pending layout can't be applied because of a bounds mismatch, we can keep it around. I'm fine with adding another flag [here](https://github.com/TextureGroup/Texture/pull/455/files#diff-71ea742a29543cb7e08c3dffe2a2508eR340) to handle that edge.

However, as said above, in case `_pendingDisplayNodeLayout->version < _layoutVersion`, it's quite important to get rid of it.

Adlai-Holler: Sorry I must be missing something big. Why is it important to get rid of it? What's the benefit?

Adlai-Holler: Oh there's a comment between those two that I missed. OK that makes sense, to release those nodes. We should have a common pattern that in one step checks if a given layout is valid, and discards it if not, and use that everywhere, but yes I'm on the same page about this line now.

nguyenhuy: 👍

nguyenhuy:
@Adlai-Holler Agreed!

I reckon the long-term solution would be to always layout from root.

nguyenhuy:
I'm gonna merge this PR and run an internal test. If we encounter another problem, I think we should really consider to revert the original PR as @appleguy suggested. Thanks everyone!

PR #465, opened 2017-07-23
[ASCornerRounding] Introduce .cornerRoundingType: CALayer, Precomposited, or Clip Corners.

appleguy:
For at least two years, we've gotten requests from the community to support alternatives to CALayer's .cornerRadius. Even on today's devices, this property still has a performance cost (including on early 64-bit SoCs, like the A7 and A8). In the cases where many views need to be rounded and exist on screens with movement, the overhead can become very significant.

This new ASCornerRounding property supports all of the modes described on the Texture website, with a minimal change in our API. This is achieved by using the .backgroundColor property to influence the mode where appropriate. In the future, I'll improve the logic to be able to detect certain undesired conditions, like non-opaque background with ClipCorner type.

http://texturegroup.org/docs/corner-rounding.html

This has already been tested with ASDKgram, including code that automatically advances through each of the corner rounding modes to ensure the transitions between them are smooth.

It would be great to get a review and work towards an accept soon, but before landing this, I'll add snapshot tests and look for opportunities to improve documentation. Suggestions welcome!

It is known that there is a subtle difference in rounding between the arc-based clip corners, and the iOS 7-style UIBezierPath precomposited corners. This is described here: https://www.paintcodeapp.com/blogpost/code-for-ios-7-rounded-rectangles . The difference is not noticeable for most use cases (and of these modes, Precomposited will be used the most, + already uses the newer style). Thus I feel confident revisiting this later for ClipCorner type.

Very brief search for some folks requesting guidance / samples / support for this:
https://github.com/facebookarchive/AsyncDisplayKit/issues/1446
https://github.com/facebookarchive/AsyncDisplayKit/issues/979


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -290,6 +281,91 @@ - (asyncdisplaykit_async_transaction_operation_block_t)_displayBlockWithAsynchro
...
+    
+    if (willDisplayNodeContentWithRenderingContext) {
+      willDisplayNodeContentWithRenderingContext(context, drawParameters);
+    }
+  }
+
+}
+- (void)__didDisplayNodeContentWithRenderingContext:(CGContextRef)context image:(UIImage **)image drawParameters:(id _Nullable)drawParameters backgroundColor:(UIColor *)backgroundColor borderWidth:(CGFloat)borderWidth borderColor:(CGColorRef)borderColor
+{
+  if (context == NULL && *image == NULL) {
```
Adlai-Holler: Not sure whether you're looking for `&& image == NULL` or `&& *image == nil` here.

Source/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -290,6 +281,91 @@ - (asyncdisplaykit_async_transaction_operation_block_t)_displayBlockWithAsynchro
...
+    // This works for everything from clearColor to opaque colors.
+    [backgroundColor setFill];
+    [roundedHole fillWithBlendMode:kCGBlendModeCopy alpha:1.0f];
+    
+    [roundedPath stroke];  // Won't do anything if borderWidth is 0 and roundedPath is nil.
+    
+    if (*image) {
+      *image = UIGraphicsGetImageFromCurrentImageContext();
+      UIGraphicsEndImageContext();
+    }
```
Adlai-Holler: Should this be `if (image)`? And should we ensure that we only end the image context if we created one above, and not if we were passed one? The context create-readimage-destroy behavior here seems a little unclear, for instance in the caller we will actually re-generate the image and overwrite it if the caller passed us a context.

Source/Private/ASDisplayNodeInternal.h
```diff
@@ -319,4 +326,10 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
 
 @end
 
+@interface ASDisplayNode (InternalPropertyBridge)
+
+@property (nonatomic, assign) CGFloat layerCornerRadius;
```
Adlai-Holler: May be worth dropping a comment on this property explaining its distinction from `.cornerRadius`? Seems obvious to us today but who knows tomorrow

appleguy:
@Adlai-Holler Thanks for the comments! I'll try to update this soon, although under a bit of a squeeze for time (not uncommon, just sayin' it like it is! :) ).

I'll certainly address your comments before landing. Anything else you'd be looking for prior to an accept?

Adlai-Holler:
@appleguy Nope! I've spoken now, so I'll forever hold my peace 💞

Source/ASDisplayNode.mm
```diff
@@ -1478,6 +1479,147 @@ - (void)recursivelySetNeedsDisplayAtScale:(CGFloat)contentsScale
...
+
+- (void)_setClipCornerLayersVisible:(BOOL)visible
+{
+  ASPerformBlockOnMainThread(^{
+    ASDisplayNodeAssertMainThread();
+    if (visible) {
+      for (int idx = 0; idx < 4; idx++) {
+        if (_clipCornerLayers[idx] == nil) {
+          _clipCornerLayers[idx] = [[CALayer alloc] init];
+          _clipCornerLayers[idx].zPosition = 99999;
```
muukii: I think following code will be better performance.
It reduce accessing to array.
```
CALayer *layer = [[CALayer alloc] init];
layer.zPosition = 99999;
layer.delegate = self;
_clipCornerLayers[idx] = layer;
```

appleguy:
Thanks for the suggestion!

I think this will create a retain and release call. If so, this is actually
much more expensive than the single memory indirection / pointer lookup.
Note that this is a C array for performance, not an NSArray.

Source/ASDisplayNode.mm
muukii: @appleguy 
Thanks! I understand.
Sorry, I may be misunderstanding about retain and release.
Does returning by subscript of NSArray not create a retain?

appleguy:
I don’t have the code in front of me, but is the array an NSArray? In an earlier version when I first wrote this, I was pretty sure it was a C array, but maybe I changed it before landing.

If it is an NSArray, it will be retained and autoreleased; if a C array, it won’t be retained at all.

The additional overhead comes from the variable reference. It will add another retain, and release call in addition to the retain + autorelease from the array access.

Although this isn’t a severe issue, I think the code is pretty acceptable as it is. However, we welcome any other suggestions or PRs, even just as cleanup / refactoring!!

appleguy:
@nguyenhuy @maicki @Adlai-Holler could I get an accept? Otherwise it won't let me land :(.

Source/Private/ASDisplayNodeInternal.h
```diff
@@ -173,6 +173,10 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
 
   // keeps track of nodes/subnodes that have not finished display, used with placeholders
   ASWeakSet *_pendingDisplayNodes;
+  
+  CGFloat _cornerRadius;
+  ASCornerRoundingType _cornerRoundingType;
+  CALayer *_clipCornerLayers[4];
```
bolsinga: Hi Scott! How are these ObjC objects retained in this C array? Is there ARC magic I'm clearly not aware of? -- Greg

PR #466, opened 2017-07-23
[ASImageNode] Enable .clipsToBounds by default (fix .cornerRadius, GIFs overflow).

appleguy:
We've seen a number of bugs reported over time that .cornerRadius didn't work on
ASNetworkImageNode. This wasn't much of a concern because cornerRadius is very
inefficient anyway, and there are better ways to round corners, but it should
certainly work.

It turns out that clipsToBounds has been off for images, and this ultimately
was behind another issue recently seen where decoded GIFs would spill outside
the bounds area to overlap nearby content.

Although there is some risk of behavior change from this, I think the risk
is fairly small, and in most cases it will probably fix behaviors in a way
that doesn't cause problems for the app.

We should consider if this property should be on for all ASDisplayNodes,
but for now it would be a great step to be confident it's on for all
ASImageNodes.

One of the reports of this issue:
https://github.com/facebookarchive/AsyncDisplayKit/issues/490


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Adlai-Holler:
Concur with Michael this is a reasonable move, with an assumption that clipsToBounds is pretty fast unless you're masking/rounding corners. But we should document the different default. At first blush I would support generalizing this to ASDisplayNode on that same assumption – having content spill over the bounds should be the exception not the rule.

wCodeQ:
I want to know how to load a GIF Image using ASImageNode,Thanks.

appleguy:
I believe it is necessary to use ASNetworkImageNode. You can also pass
file:/// URLs to it. Make sure you have PINRemoteImage linked.

PR #467, opened 2017-07-23
[ASCollectionView] Fix index space translation of Flow Layout Delegate methods.

appleguy:
This includes a few other cleanups, including overflow of signed integer indices.

It will take a couple PRs to fully sync both upstream and downstream in this case, but I am happy to integrate feedback along the way. Just a heads up that it may be easier to do in subsequent PRs, depending on the importance of changes (stylistic vs functional).


Source/ASCollectionView.mm
```diff
@@ -761,6 +765,28 @@ - (void)setUsesSynchronousDataLoading:(BOOL)usesSynchronousDataLoading
...
+- (void)invalidateFlowLayoutDelegateMetrics {
+  UICollectionViewLayout *layout = self.collectionViewLayout;
+  CGSize defaultSize = ({
+    UICollectionViewFlowLayout *flowLayout = ASDynamicCast(layout, UICollectionViewFlowLayout);
+    flowLayout ? flowLayout.itemSize : CGSizeZero;
+  });
+
+  for (ASCollectionElement *element in self.dataController.pendingMap.itemElements) {
+    ASCellNode *node = element.node;
+    if (node.shouldUseUIKitCell) {
```
Adlai-Holler: 1. It's probably inconsequential, but maybe we should just ignore calls where the layout isn't a flow layout, rather than using size 0?

1. Instead of building a giant array – I don't really like the `itemElements` property and may propose removing it – it's better to enumerate the map directly and `if (element.supplementaryElementKind) { continue; }` Note to self: add a `UICollectionElementCategory elementCategory` property to ASCollectionElement to simplify this distinction.

1. Shouldn't we also refetch header/footer reference sizes here? Meshes well with 2

Source/ASCollectionView.mm
```diff
@@ -1010,23 +1036,23 @@ - (UIEdgeInsets)collectionView:(UICollectionView *)cv layout:(UICollectionViewLa
...
     return [(id)_asyncDelegate collectionView:cv layout:l
-     minimumInteritemSpacingForSectionAtIndex:indexPath.section];
+               minimumInteritemSpacingForSectionAtIndex:indexPath.section];
   }
   return ASFlowLayoutDefault(l, minimumInteritemSpacing, 10.0); // Default is documented as 10.0
 }
 
 - (CGFloat)collectionView:(UICollectionView *)cv layout:(UICollectionViewLayout *)l
-minimumLineSpacingForSectionAtIndex:(NSInteger)section
+                    minimumLineSpacingForSectionAtIndex:(NSInteger)section
```
Adlai-Holler: 🙂

Source/Base/ASAssert.h
```diff
@@ -64,8 +64,8 @@
 #define ASDisplayNodeConditionalAssert(shouldTestCondition, condition, desc, ...) ASDisplayNodeAssert((!(shouldTestCondition) || (condition)), desc, ##__VA_ARGS__)
 #define ASDisplayNodeConditionalCAssert(shouldTestCondition, condition, desc, ...) ASDisplayNodeCAssert((!(shouldTestCondition) || (condition)), desc, ##__VA_ARGS__)
 
-#define ASDisplayNodeCAssertPositiveReal(description, num) ASDisplayNodeCAssert(num >= 0 && num <= CGFLOAT_MAX, @"%@ must be a real positive integer.", description)
-#define ASDisplayNodeCAssertInfOrPositiveReal(description, num) ASDisplayNodeCAssert(isinf(num) || (num >= 0 && num <= CGFLOAT_MAX), @"%@ must be infinite or a real positive integer.", description)
+#define ASDisplayNodeCAssertPositiveReal(description, num) ASDisplayNodeCAssert(num >= 0 && num <= CGFLOAT_MAX, @"%@ must be a real positive integer: %f.", description, (CGFloat)num)
+#define ASDisplayNodeCAssertInfOrPositiveReal(description, num) ASDisplayNodeCAssert(isinf(num) || (num >= 0 && num <= CGFLOAT_MAX), @"%@ must be infinite or a real positive integer: %f.", description, (CGFloat)num)
```
Adlai-Holler: 🙌

Source/ASCollectionView.mm
appleguy: @Adlai-Holler some merge conflicts have arisen, and I'm afraid my time availability is nearing a local minimum. :( Do you think this PR is acceptable to land as-is if the conflicts are fixed?

1. I think it's better to have the 0 default, but indeed maybe no one will ever know :)
2. The iteration suggestion sounds like an improvement for sure. I'm not sure I can commit to doing it only because I would have to get some test cases running to exercise the change.
3. This seems potentially important, and indeed might motivate me to do #2.

Could you give me an Accept if you are comfortable with this, and then if I do 2 or 3, I can land without waiting?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/78b10ffebadecd399c2ba8ae7a2d1102/raw/13e457e9d55478a65b4b6fa8c1c2d80063728e47/log.txt)

Source/ASCollectionView.mm
appleguy: @Adlai-Holler It took me a while, but I finally got around to implementing your advice — and went quite a bit further. Thanks for suggesting these things, as they are improvements.

In particular, I was intentionally skipping over handling the update of supplementary element sizes, which is not great. That's now fixed and the code is better factored.

If you or @nguyenhuy can give me a sanity check review soon, I'd appreciate that - would like to merge this in pretty soon.

appleguy:
Hmm, this erroneous test failure is striking again.

ASCollectionViewTests
  testThatMultipleBatchFetchesDontHappenUnnecessarily, failed - Too many batch fetches!
  /Users/Shared/buildkite/builds/iosf-garrett-VMrmmvuqKsO4.dyn.pinadmin.com-1/pinterest/texture/Tests/ASCollectionViewTests.mm:894
  ```
    if (batchFetchCount > 1) {
      XCTFail(@"Too many batch fetches!");

appleguy:
Thanks for reviewing and merging!

PR #468, opened 2017-07-23
[Accessibility] Add .isAccessibilityContainer property, allowing automatic aggregation of children's a11y labels.

appleguy:
After consulting Apple documentation and working with some a11y experts,
we've found that aggregating objects that have a11y labels but are not
themselves interactable is significantly preferred for these users.

It makes it much quicker to navigate scrolling content if VoiceOver only
stops to select entire cells, and then allows drilling down into the cell
to select individual components. This implementation achieves that behavior.

We should consider enabling isAccessibilityContainer by default on ASCellNode.
This would be an improvement for 95% of a11y use cases. Aggregation can be
enabled or disabled on any node.


Source/ASDisplayNode+Beta.h
```diff
@@ -113,6 +113,19 @@ typedef struct {
 #endif
 
 /**
+ * @abstract Whether this node acts as an accessibility container. If set to YES, then this node's accessibility label will represent
+ * an aggregation of all child nodes' accessibility labels. Nodes in this node's subtree that are also accessibility containers will
+ * not be included in this aggregation, and will be exposed as separate accessibility elements to UIKit.
+ */
+@property (nonatomic, assign) BOOL isAccessibilityContainer;
```
Adlai-Holler: 1. Just acknowledging that while this is not Objective-C convention (having `is` in the property name), this follows the UIAccessibility convention so good call.

1. Should we use objective-C atomics here? Not a big deal either way.

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -100,12 +131,64 @@ static void CollectUIAccessibilityElementsForNode(ASDisplayNode *node, ASDisplay
   });
 }
 
+static void CollectAccessibilityElementsForContainer(ASDisplayNode *container, _ASDisplayView *view, NSMutableArray *elements) {
+  UIAccessibilityElement *accessiblityElement = [ASAccessibilityElement accessibilityElementWithContainer:view node:container containerNode:container];
+
+  NSMutableArray<ASAccessibilityElement *> *labeledNodes = [NSMutableArray array];
+  NSMutableArray<ASAccessibilityCustomAction *> *actions = [NSMutableArray array];
+  std::queue<ASDisplayNode *> queue;
+  queue.push(container);
```
Adlai-Holler: Somebody should build an NSFastEnumeration-conforming class that takes a BFS/DFS option, a collection, and a `children` block. Thinking out loud.

appleguy:
Great idea on the enumerator utility!

We probably should set a standard for BOOLs. I worry it might miss some transactional atomicity, although in this case it is probably fine.

ay8s:
👍  Thanks for this one @appleguy. We're focussing a bunch on accessibility and have tweaks to our cell nodes but containers would make much more sense than the current route we've taken.

PR #470, opened 2017-07-23
[ASNodeController] Add -nodeDidLayout callback. Allow switching retain behavior at runtime.

appleguy:
With these changes, I'd also like to propose that we move ASNodeController
out of Beta (renaming the files without +Beta). Let me know what you think!

Because we don't support ASNodeController directly in ASCV / ASTV, it is still
important to allow flipping the ownership in certain cases (in particular, for
root CellNodeController objects that should follow the lifecycle of the
ASCellNode rather than owning the ASCellNode).

### DIFF
--- a/Source/ASNodeController+Beta.m
+++ b/Source/ASNodeController+Beta.m
@@ -15,34 +15,28 @@
 //      http://www.apache.org/licenses/LICENSE-2.0
 //
 
-#import "ASNodeController+Beta.h"
-#import "ASDisplayNode+FrameworkPrivate.h"
+#import <AsyncDisplayKit/ASWeakProxy.h>
+#import <AsyncDisplayKit/ASNodeController+Beta.h>
+#import <AsyncDisplayKit/ASDisplayNode+FrameworkPrivate.h>
 
-#if INVERT_NODE_CONTROLLER_OWNERSHIP
+#define _node (_shouldInvertStrongReference ? _weakNode : _strongNode)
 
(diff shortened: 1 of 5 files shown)


Source/ASNodeController+Beta.m
```diff
@@ -89,3 +108,41 @@ - (void)interfaceStateDidChange:(ASInterfaceState)newState
...
+- (void)__setNodeControllerStrong:(ASNodeController *)nodeController
+{
+  objc_setAssociatedObject(self, @selector(__nodeController), nodeController, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+}
+
+- (void)__setNodeControllerWeak:(ASNodeController *)nodeController
+{
+  // Associated objects don't support weak references. Since assign can become a dangling pointer, use ASWeakProxy.
+  ASWeakProxy *nodeControllerProxy = [ASWeakProxy weakProxyWithTarget:nodeController];
+  objc_setAssociatedObject(self, @selector(__nodeController), nodeControllerProxy, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
```
Adlai-Holler: I'm cool with this, but why not add 2 `@package` ivars to ASDisplayNode and use those instead?

appleguy:
@Adlai-Holler - @package sounds great, but what header could it be in? I'd rather not keep adding so much rob ASDIsplayNodeInternal, but maybe it's acceptable to do so.

PR #472, opened 2017-07-25
[ASStackLayoutSpec] Flex wrap fix and lineSpacing property

flovouin:
Hi guys,

I haven't opened an issue / feature request for this, but here it is, a PR that fixes a small bug in the flex wrap layout, and add a nice property to it.

- The first commit fixes a bug where spacing is not taken into account when splitting items in lines. This causes items to overflow the allowed size range in certain cases. Specifically, too many items can be added to a line because the line length doesn't keep track of spacing. See the demo project for an example.
- The second commit adds the `lineSpacing` property to the `ASStackLayoutSpec` that... well... adds space between lines. I tried to keep changes as minimal as possible, tell me what you think!

As usual, here's a small [demo project](https://github.com/flovouin/Texture-FlexWrapBugs) that demonstrates the bug and the new feature.

Cheers,

Flo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## master
 
 * Add your own contributions to the next release on the line below this with your name.
+- [ASStackLayoutSpec] Add lineSpacing property working with flex wrap. [Flo Vouin](https://github.com/flovouin)
+- [ASStackLayoutSpec] Fix flex wrap overflow in some cases using item spacing. [Flo Vouin](https://github.com/flovouin)
 - [ASNodeController] Add -nodeDidLayout callback. Allow switching retain behavior at runtime. [Scott Goodson](https://github.com/appleguy)
 - [ASCollectionView] Add delegate bridging and index space translation for missing UICollectionViewLayout properties. [Scott Goodson](https://github.com/appleguy)
 - [ASTextNode2] Add initial implementation for link handling. [Scott Goodson](https://github.com/appleguy) [#396](https://github.com/TextureGroup/Texture/pull/396)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## master
 
 * Add your own contributions to the next release on the line below this with your name.
+- [ASStackLayoutSpec] Fix flex wrap overflow and add lineSpacing property. [Flo Vouin](https://github.com/flovouin)
```
nguyenhuy: Since line spacing is a significant addition, I'd suggest to mention it on a separate line so people won't overlook it. It's your call though.

Source/Details/ASCollectionFlowLayoutDelegate.m
```diff
@@ -70,6 +70,7 @@ + (ASCollectionLayoutState *)calculateLayoutWithContext:(ASCollectionLayoutConte
                                                                       alignItems:ASStackLayoutAlignItemsStart
                                                                         flexWrap:ASStackLayoutFlexWrapWrap
                                                                     alignContent:ASStackLayoutAlignContentStart
+                                                                     lineSpacing:0
```
nguyenhuy: Since we have a convenient initializer that doesn't require `lineSpacing`, please remove this change.

Source/Details/ASCollectionGalleryLayoutDelegate.m
```diff
@@ -81,6 +81,7 @@ + (ASCollectionLayoutState *)calculateLayoutWithContext:(ASCollectionLayoutConte
                                                                       alignItems:ASStackLayoutAlignItemsStart
                                                                         flexWrap:ASStackLayoutFlexWrapWrap
                                                                     alignContent:ASStackLayoutAlignContentStart
+                                                                     lineSpacing:0
```
nguyenhuy: Same as above.

flovouin:
Hi Huy, thanks for the review!
I've made the changes you requested, so it should be all good.

I do agree that tests would be a nice addition. I'll try to address this but unfortunately I can't give you a time estimate for this, so I think it would be best to go ahead with this PR and I'll submit another one in the future.

Cheers,

nguyenhuy:
Ok, let's get this in. I created a separate issue for unit tests. Thanks for the awesome work, @flovouin!

PR #479, opened 2017-07-27
Update subclassing.md

oferRounds:
Adding the fact that `didLoad` is called only once. 

P.s. I‘m not an English native speaker, so there might be a better phrasing.


oferRounds:
Done!

oferRounds:
Sure :)

PR #482, opened 2017-07-28
Update corner-rounding.md

oferRounds:
If I understand right, both Precomposited Corners techniques uses `[path clip]`.

If so, I think it will be better to move it to the general description of the section, as currently it seems as if only the second technique uses `[path clip]`.

### DIFF
--- a/docs/_docs/corner-rounding.md
+++ b/docs/_docs/corner-rounding.md
@@ -17,7 +17,7 @@ When it comes to corner rounding, many developers stick with CALayer's `.cornerR
 
 ## CALayer's .cornerRadius is Expensive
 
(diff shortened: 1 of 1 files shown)


oferRounds:
Perhaps in the future, it’s worth adding a concrete code examples, and maybe also make this section a bit more general, to include other actions which trigger off-screen rendering, such a using the `Layer`’s `shadow` and `border` properties (while describing also the better alternatives to them).

oferRounds:
@appleguy - and just a question I was wondering about — why is iOS built in this way that when using `.cornerRadius` it triggers offscreen rendering? I.e., If Apple would like to change this behaviour, can they do things different in the way that the clipping operation would not trigger the offscreen rendering, or is this thing is inherently unavoidable?

appleguy:
@oferRounds Thanks for doing this! :)

PR #484, opened 2017-07-29
Update scroll-node.md

oferRounds:
Regarding the last sentence: “As you can see, the `scrollNode`'s underlying view is a `ASScrollNode`.”:
1. I think that `ASScrollNode` should be replaced with `UIScrollView`.
2. But — not sure how can it be seen from this example...


oferRounds:
@maicki agreed, done!

PR #485, opened 2017-07-30
iOS 11 UITableView automatic height estimation fix

christianselig:
Starting in iOS 11 `UITableView` automatically uses estimated heights. ([Source](https://twitter.com/smileyborg/status/871858815053946880))

This causes some odd behaviour in Texture, specifically with batch fetching where a lot of jumping can happen when the load occurs. 

[User djug](https://asyncdisplaykit.slack.com/archives/C0V63R86T/p1500761127076778) in the Slack encountered this issue earlier and posted their solution which is the basis for this, and is as simple as re-disabling the new default behaviour so that automatic height estimation does **not** occur.

In order to let this occur only on iOS 11 (not sure if there's any pre-iOS-11 code running on the assumption of the old default, don't want to cause a regression) I made it only occur on iOS 11 and later. This required an addition to `ASAvailability.h`, and like iOS 10 it's unfortunately not explicitly set, so I set it to what iOS 11 beta 4 currently outputs as its `kCFCoreFoundationVersionNumber`: `1438.1`.

(Interestingly Xcode 9 now allows for Objective-C to use the Swift-esque `@availability` checks, which may be worth looking into, but it obviously would break compatibility with previous versions of Xcode.)

Overall simple change that helps scrolling on iOS 11 to behave as expected.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

christianselig:
@maicki That makes sense to me, at first blush it's definitely not obvious why it's there. I added in a note!.

nguyenhuy:
Good call, @maicki.  Will merge after CI passes.

PR #450, opened 2017-07-17
fix SIMULATE_WEB_RESPONSE not imported #449

wsdwsd0829:
I could not found the correct format for "license as suggested by our bot",  what should I use exactly?

PR #509, opened 2017-08-14
ASCollectionLayout to return a zero content size if its state is unavailable

nguyenhuy:
Reported by @wsdwsd0829 [here](https://github.com/TextureGroup/Texture/pull/450#discussion_r132855870). 

Reproducible in "examples/ASCollectionView" by enabling `ASYNC_COLLECTION_LAYOUT` and popping the view controller after just 2 seconds ([this line
](https://github.com/TextureGroup/Texture/blob/master/examples/ASCollectionView/Sample/ViewController.m#L96)). The crash occurs when a containing view controller is popped out right after its collection view finished a batch update, but before an update animation settles. As soon as the view controller is deallocated, its collection view's layout is invalidated. Then when the update animation stops, it asks for the layout's content size and triggers the assertion. Fix by returning a zero size instead.

### DIFF
--- a/Source/Private/ASCollectionLayout.mm
+++ b/Source/Private/ASCollectionLayout.mm
@@ -140,6 +140,8 @@ static const ASScrollDirection kASStaticScrollDirection = (ASScrollDirectionRigh
 - (CGSize)collectionViewContentSize
 {
   ASDisplayNodeAssertMainThread();
+  // The content size can be queried right after a layout invalidation (https://github.com/TextureGroup/Texture/pull/509).
+  // In that case, return zero.
   return _layout ? _layout.contentSize : CGSizeZero;
 }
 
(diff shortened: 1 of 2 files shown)


Source/Private/ASCollectionLayout.mm
```diff
@@ -140,8 +140,7 @@ - (void)invalidateLayout
 - (CGSize)collectionViewContentSize
 {
   ASDisplayNodeAssertMainThread();
-  ASDisplayNodeAssertNotNil(_layout, @"Collection layout state should not be nil at this point");
-  return _layout.contentSize;
+  return _layout ? _layout.contentSize : CGSizeZero;
```
Adlai-Holler: So that we don't regress on this in the future, could you add a link to this PR explaining the workaround?

nguyenhuy:
Thanks, @Adlai-Holler!

PR #510, opened 2017-08-14
Avoid re-entrant call to self.view when applying initial pending state

Adlai-Holler:
We are seeing a crash in Pinterest like the one attached below.

I don't know why the re-entrant call to self.view would not return immediately (since _view is not nil as guaranteed by the `_loadViewOrLayer` method), and I don't know why we're getting a KERN_PROTECTION_FAILURE,  but either way, I believe this is the correct code for the `_locked_` method.

screen shot 2017-08-14 at 11 00 19 am


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Issue #521, opened 2017-08-20
[ASBatchContext] If a user's scroll lifts as an insertion is processing, new fetch starts early.


appleguy:
Here's the scenario:

0. ASBatchContext fetching is being used by an application.
1. ASBatchContext triggers a fetch of new data; network fetch completes, ending the batch.
2. The app processes the network fetch and triggers an insert of items in the collection / table.
3. [In the event that] a user's touch scrolling the view lifts at this moment, the check for batch fetching will reference the current .contentSize and decide that it is within the same window that 1. decided to trigger a fetch (because no fetch is in progress now).
4. Most likely before the new network request returns, the insertion finishes processing and the .contentSize is updated.
5. Batch finishes and inserts even more content, likely well beyond the intended batch interval.

This will not occur for all batch fetches, as the window of time is only as long as the concurrent layout phase of the ASCellNodes (typically less than a second, or even less than 200ms on newer devices).

The primary impact of this issue is additional server load. Bandwidth, CPU, and memory consumption are all minimal: this does not trigger additional items to be added to the preload or display ranges.

See also: upcoming PR adding -isProcessingUpdates and -addUpdateCompletionBlock: to ASCollectionNode, which may be used to implement a simple solution for this.

cc @nguyenhuy, @maicki, @Adlai-Holler

appleguy:
Refer to this PR for an easy way we can fix this issue: https://github.com/TextureGroup/Texture/pull/522

PR #522, opened 2017-08-20
[ASCollectionNode] Add -isProcessingUpdates and -onDidFinishProcessingUpdates: APIs.

appleguy:
Over time, there have actually been a lot of legitimate uses for an API like this.
In fact, I'm not quite sure what has held us back from adding one!

I believe that at least some portion of -wait calls (even if less than 50%) could be
replaced with -onDidFinishProcessingUpdates:, which could potentially improve the
performance of applications using -wait by a significant amount.

Please take a close look at implementation correctness. Although I'm in a bit of a
rush, I'm aiming to make this properly documented and added a basic test -- but it
could certainly use some more detailed testing as a followup.

### DIFF
--- a/Source/ASCollectionNode.h
+++ b/Source/ASCollectionNode.h
@@ -253,11 +253,6 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)performBatchUpdates:(nullable AS_NOESCAPE void (^)())updates completion:(nullable void (^)(BOOL finished))completion;
 
 /**
- *  Blocks execution of the main thread until all section and item updates are committed to the view. This method must be called from the main thread.
- */
-- (void)waitUntilAllUpdatesAreProcessed;
-
-/**
  *  Returns YES if the ASCollectionNode is still processing changes from performBatchUpdates:.
  *  This is typically the concurrent allocation (calling nodeBlocks) and layout of newly inserted
  *  ASCellNodes. If YES is returned, then calling -waitUntilAllUpdatesAreProcessed may take tens of
@@ -269,7 +264,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  This method will always return NO if called immediately after -waitUntilAllUpdatesAreProcessed.
  */
-- (BOOL)isProcessingUpdates;
+@property (nonatomic, readonly) BOOL isProcessingUpdates;
 
 /**
  *  Schedules a block to be performed (on the main thread) after processing of performBatchUpdates:
@@ -285,6 +280,11 @@ NS_ASSUME_NONNULL_BEGIN
  *  Calling -waitUntilAllUpdatesAreProcessed is one way to flush any pending update completion blocks.
  */
 - (void)onDidFinishProcessingUpdates:(nullable void (^)())didFinishProcessingUpdates;
+
+/**
+ *  Blocks execution of the main thread until all section and item updates are committed to the view. This method must be called from the main thread.
+ */
+- (void)waitUntilAllUpdatesAreProcessed;
 
 /**
  * Inserts one or more sections.
--- a/Source/ASCollectionNode.h
+++ b/Source/ASCollectionNode.h
@@ -534,9 +534,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)reloadDataImmediately ASDISPLAYNODE_DEPRECATED_MSG("Use -reloadData / -reloadDataWithCompletion: followed by -waitUntilAllUpdatesAreProcessed instead.");
 
-// TODO: Rename framework uses of this method and add deprecation message.
-// ASDISPLAYNODE_DEPRECATED_MSG("This method has been renamed to -waitUntilAllUpdatesAreProcessed.");
-- (void)waitUntilAllUpdatesAreCommitted;
+- (void)waitUntilAllUpdatesAreCommitted ASDISPLAYNODE_DEPRECATED_MSG("This method has been renamed to -waitUntilAllUpdatesAreProcessed.");
 
 @end
 
--- a/Source/ASCollectionNode.h
+++ b/Source/ASCollectionNode.h
@@ -255,7 +255,36 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Blocks execution of the main thread until all section and item updates are committed to the view. This method must be called from the main thread.
  */
-- (void)waitUntilAllUpdatesAreCommitted;
+- (void)waitUntilAllUpdatesAreProcessed;
+
+/**
+ *  Returns YES if the ASCollectionNode is still processing changes from performBatchUpdates:.
+ *  This is typically the concurrent allocation (calling nodeBlocks) and layout of newly inserted
+ *  ASCellNodes. If YES is returned, then calling -waitUntilAllUpdatesAreProcessed may take tens of
+ *  milliseconds to return as it blocks on these concurrent operations.
+ *
+ *  Returns NO if ASCollectionNode is fully synchronized with the underlying UICollectionView. This
+ *  means that until the next performBatchUpdates: is called, it is safe to compare UIKit values
+ *  (such as from UICollectionViewLayout) with your app's data source.
+ *
+ *  This method will always return NO if called immediately after -waitUntilAllUpdatesAreProcessed.
+ */
+- (BOOL)isProcessingUpdates;
+
+/**
+ *  Schedules a block to be performed (on the main thread) after processing of performBatchUpdates:
+ *  is finished (completely synchronized to UIKit). The blocks will be run at the moment that
+ *  -isProcessingUpdates changes from YES to NO;
+ *
+ *  When isProcessingUpdates == NO, the block is run block immediately (before the method returns).
+ *
+ *  Blocks scheduled by this mechanism are NOT guaranteed to run in the order they are scheduled.
+ *  They may also be delayed if performBatchUpdates continues to be called; the blocks will wait until
+ *  all running updates are finished.
+ *
+ *  Calling -waitUntilAllUpdatesAreProcessed is one way to flush any pending update completion blocks.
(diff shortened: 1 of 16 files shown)


appleguy:
This could be used to quite easily fix this ASBatchContext edge case: https://github.com/TextureGroup/Texture/issues/521

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/9da03b0b6493cc4128bfbec9bc0d81d0/raw/49ca1839495b42090dc5341de687f1cd3d9da9f5/log.txt)

appleguy:
I see one of the new tests is failing on the build server, although it doesn't fail on my local machine - even trying 9.3, 10.0, and 10.3.1 simulators 

I'll wait for the last build to finish and then will start debugging-by-commit-pushing :).

Source/ASCollectionNode.h
```diff
@@ -255,7 +255,36 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *  ASCellNodes. If YES is returned, then calling -waitUntilAllUpdatesAreProcessed may take tens of
+ *  milliseconds to return as it blocks on these concurrent operations.
+ *
+ *  Returns NO if ASCollectionNode is fully synchronized with the underlying UICollectionView. This
+ *  means that until the next performBatchUpdates: is called, it is safe to compare UIKit values
+ *  (such as from UICollectionViewLayout) with your app's data source.
+ *
+ *  This method will always return NO if called immediately after -waitUntilAllUpdatesAreProcessed.
+ */
+- (BOOL)isProcessingUpdates;
```
Adlai-Holler: Trivial: Make this a readonly property for better swift interop?

Source/Details/ASDataController.mm
```diff
@@ -437,6 +437,29 @@ - (void)waitUntilAllUpdatesAreCommitted
   [self _scheduleBlockOnMainSerialQueue:^{ }];
 }
 
+- (BOOL)isProcessingUpdates
+{
+  ASDisplayNodeAssertMainThread();
+  long waitResult = dispatch_group_wait(_editingTransactionGroup, DISPATCH_TIME_NOW);
+  return (waitResult == 0 ? NO : YES);
```
Adlai-Holler: Is it possible that we could return a false `NO` if they call this in between the time the editing transaction group is finished and the main queue services the data controller's UIKit flush?

We could either flush the main serial queue (but not the editing transaction queue) or add API to check if the main serial queue is empty.

Source/Details/ASDataController.mm
```diff
@@ -437,6 +437,29 @@ - (void)waitUntilAllUpdatesAreCommitted
...
+  } else {
+    dispatch_async(_editingTransactionQueue, ^{
+      // Retry the block. If we're done processing updates, it'll run immediately, otherwise
+      // wait again for updates to quiesce completely.
+      dispatch_async(dispatch_get_main_queue(), ^{
+        [self onDidFinishProcessingUpdates:completion];
+      });
+    });
+  }
+}
```
Adlai-Holler: A couple low-impact notes on this method:

- If we're asserting main, is it worth removing `ASPerformBlockOnMainThread`? Presumably an error of this kind can be reliably caught during development. If the notion here is "better safe than sorry," I'm down with that as well.
- Should we schedule the retry on the main serial queue, so that if they wait on updates to complete we can be sure this block is run before returning from the wait? I suppose it's not a big deal – there's no guarantee this block will be run **as soon as** we finish processing updates. But at the same time, this is a break from the usual pattern of `editingQueue -> main serial queue` for main-thread-injection into data controller's async work.

Tests/ASCollectionViewTests.mm
```diff
@@ -1050,8 +1050,17 @@ - (void)testInitialRangeBounds
...
+  XCTAssertTrue(cn.isProcessingUpdates, @"ASCollectionNode should still be processing updates after initial layoutIfNeeded call (reloadData)");
+  [cn onDidFinishProcessingUpdates:^{
+    XCTAssertTrue(!cn.isProcessingUpdates, @"ASCollectionNode should no longer be processing updates inside -onDidFinishProcessingUpdates: block");
+  }];
+
   // Wait for ASDK reload to finish
   [cn waitUntilAllUpdatesAreCommitted];
+
+  XCTAssertTrue(!cn.isProcessingUpdates, @"ASCollectionNode should no longer be processing updates after -wait call");
+
```
Adlai-Holler: 💕

Source/ASCollectionNode.h
appleguy: Great suggestion, I'll try to remember this in the future (so foreign to me :-P)

Source/Details/ASDataController.mm
appleguy: Yeah, this is a good catch. I think this might be possible.

If we flush the main serial queue here, you could get situations where calling isProcessingUpdates results in calls out to the delegate / dataSource, which would be pretty weird and maybe unexpected by clients. So I think I'll just check the length of the queue.

appleguy: @Adlai-Holler do you think we should perform a setNeedsLayout in the case that the queues are drained, before returning NO? This would technically be needed to ensure all information is flushed to / reflected in the UICollectionViewLayout...

Source/Details/ASDataController.mm
appleguy: Yes, the idea is some extra safety in the case the assertion fails in production. It is not likely to be necessary, but the semantics of PerformOnMain are clear enough that it seems OK.

Interested in your input on the other point. I had originally implemented it as waiting on the mainSerialQueue instead of a dispatch.

I was thinking (and now believe this may be incorrect) that doing it this way would allow us to call out to the blocks in an "intermediate quiesced" state to perform some operations before potentially more edits start getting processed.

However with the changes to isProcessingUpdates, I think this is actually invalid and we should schedule on the serial queue.

appleguy: @Adlai-Holler is it correct that we do NOT perform using the _editingTransaction*Group*? I don't want multiple scheduled blocks to get stuck waiting for each other (as they do schedule in the main serial queue, and so I think this is correct, but I'm not completely confident.

Source/Details/ASMainSerialQueue.mm
```diff
@@ -40,6 +40,11 @@ - (instancetype)init
   return self;
 }
 
+- (NSUInteger)numberOfScheduledBlocks
+{
+  return _blocks.count;
```
nguyenhuy: I think we need to grab `_serialQueueLock` lock here.

Source/Details/ASDataController.mm
```diff
@@ -437,6 +437,35 @@ - (void)waitUntilAllUpdatesAreCommitted
   [self _scheduleBlockOnMainSerialQueue:^{ }];
 }
 
+- (BOOL)isProcessingUpdates
+{
+  ASDisplayNodeAssertMainThread();
+  if (_mainSerialQueue.numberOfScheduledBlocks > 0) {
+    return YES;
```
nguyenhuy: 👍

Source/ASCollectionNode.h
```diff
@@ -501,9 +530,13 @@ NS_ASSUME_NONNULL_BEGIN
...
  *
  * @deprecated This method is deprecated in 2.0. Use @c reloadDataWithCompletion: and
- *   then @c waitUntilAllUpdatesAreCommitted instead.
+ *   then @c waitUntilAllUpdatesAreProcessed instead.
  */
-- (void)reloadDataImmediately ASDISPLAYNODE_DEPRECATED_MSG("Use -reloadData / -reloadDataWithCompletion: followed by -waitUntilAllUpdatesAreCommitted instead.");
+- (void)reloadDataImmediately ASDISPLAYNODE_DEPRECATED_MSG("Use -reloadData / -reloadDataWithCompletion: followed by -waitUntilAllUpdatesAreProcessed instead.");
+
+// TODO: Rename framework uses of this method and add deprecation message.
+// ASDISPLAYNODE_DEPRECATED_MSG("This method has been renamed to -waitUntilAllUpdatesAreProcessed.");
```
nguyenhuy: If you're not going to do this in this PR, make sure to create an issue to track it.

appleguy:
@nguyenhuy @Adlai-Holler thanks for the great feedback! I went ahead and renamed the method from Committed to Processing everywhere except the *View classes, where a deprecation warning already exists and now directs people to the Processing version (it seemed silly to add a second, also deprecated name).

This change makes me realize that we should give AS{Collection,Table}Node access to the ASDataController directly. There are a number of APIs where no participation from the View class is required. Even without breaking support for using the View classes independently (which is becoming less and less important), calls like isProcessingUpdates or waitUntil.. could be passed directly from the Node to ASDataController.

nguyenhuy:
@appleguy: Yes, agree that the nodes should have direct access to data controller. In a far future, we may even tell the data controller to do the initial load before the view is even allocated. Definitely possible with an async collection layout.

Source/Details/ASDataController.mm
Adlai-Holler: Are you maybe thinking about `layoutIfNeeded`? That would absolutely make sense 👍

Source/Details/ASDataController.mm
Adlai-Holler: That's right! The transaction group is only (and always) applied to work blocks on the editing transaction queue.

nguyenhuy:
ok, let's get this in!

appleguy:
Thanks Huy!

PR #525, opened 2017-08-22
[Table and collection views] Consider content inset when calculating (default) element size range

nguyenhuy:


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASPagerNode.m
```diff
@@ -115,6 +115,17 @@ - (NSInteger)currentPageIndex
   return (self.view.contentOffset.x / CGRectGetWidth(self.view.bounds));
 }
 
+- (CGSize)pageSize
+{
+  CGSize pageSize = self.bounds.size;
+  UIEdgeInsets contentInset = self.view.contentInset;
+  if (! UIEdgeInsetsEqualToEdgeInsets(UIEdgeInsetsZero, contentInset)) {
+    pageSize.width -= (contentInset.left + contentInset.right);
```
Adlai-Holler: Do we want to apply the horizontal inset here? It makes perfect sense to me that users would want to account for content inset in the cross direction, but I'm not so sure about the scrolling direction. For example, if they have a 50pt left inset so that the carousel starts off inset by a bit, they probably still want each page to be full-width, since the insets will only be applied at the start/end of the feed, not on every page. Thoughts?

Source/Details/ASCollectionViewLayoutInspector.m
```diff
@@ -27,6 +27,11 @@
 // of the collection view
 ASSizeRange NodeConstrainedSizeForScrollDirection(ASCollectionView *collectionView) {
   CGSize maxSize = collectionView.bounds.size;
+  UIEdgeInsets contentInset = collectionView.contentInset;
+  if (! UIEdgeInsetsEqualToEdgeInsets(UIEdgeInsetsZero, contentInset)) {
+    maxSize.width -= (contentInset.left + contentInset.right);
+    maxSize.height -= (contentInset.top + contentInset.bottom);
```
Adlai-Holler: Same principle as in the pager: I think we should not apply these in the scrolling direction.

Source/ASPagerNode.m
nguyenhuy: Oh, good point! I'll fix this.

Source/Details/ASCollectionViewLayoutInspector.m
nguyenhuy: :+1:

Source/ASCollectionView.mm
```diff
@@ -1536,10 +1536,6 @@ - (ASScrollDirection)scrollableDirections
   return [self.layoutInspector scrollableDirections];
 }
 
-- (ASScrollDirection)flowLayoutScrollableDirections:(UICollectionViewFlowLayout *)flowLayout {
-  return (flowLayout.scrollDirection == UICollectionViewScrollDirectionHorizontal) ? ASScrollDirectionHorizontalDirections : ASScrollDirectionVerticalDirections;
-}
-
```
nguyenhuy: This method isn't used anywhere.

Source/ASPagerNode.m
```diff
@@ -66,6 +66,7 @@ - (instancetype)init
 - (instancetype)initWithCollectionViewLayout:(ASPagerFlowLayout *)flowLayout;
 {
   ASDisplayNodeAssert([flowLayout isKindOfClass:[ASPagerFlowLayout class]], @"ASPagerNode requires a flow layout.");
+  ASDisplayNodeAssertTrue(flowLayout.scrollDirection == UICollectionViewScrollDirectionHorizontal);
```
nguyenhuy: The existing implementation of this class (e.g `-currentPageIndex` and `- (void)scrollToPageAtIndex: animated:`) assumes a horizontal scroll direction. Let's formalize that.

Source/ASPagerNode.m
```diff
@@ -112,7 +113,15 @@ - (void)didLoad
 
 - (NSInteger)currentPageIndex
 {
-  return (self.view.contentOffset.x / CGRectGetWidth(self.view.bounds));
+  return (self.view.contentOffset.x / [self pageSize].width);
```
nguyenhuy: The old implementation works as expected, but let's not assume that `self.view.bounds.width == self.pageSize.width`.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e4ba53843f3dec6f7861bf750135c740/raw/dfc5d6736542b80780ca8be4b9274aa67633e1ee/log.txt)

Issue #495, opened 2017-08-04
'*** -[__NSSetM addObject:]: object cannot be nil' error caused by a decoration view


edelibasi:
*I have the flow layout that contains the decoration view in the first section:*

`public class SeparatorCollectionViewFlowLayout: UICollectionViewFlowLayout {

    init(separatorWidth: CGFloat = 1, separatorColor: UIColor = UIColor.black) {
        self.separatorWidth = separatorWidth
        self.separatorColor = separatorColor
        super.init()
        register(SeparatorView.self, forDecorationViewOfKind: "TimelineSeparator")

    }
    
    required public init?(coder aDecoder: NSCoder) {
        super.init(coder: aDecoder)
    }
    
    override public func layoutAttributesForDecorationView(ofKind elementKind: String, at indexPath: IndexPath) -> UICollectionViewLayoutAttributes? {
        var frame = CGRect.zero

        if let rows = collectionView?.numberOfItems(inSection: 0), indexPath.row < rows {
            if let cellAttributes = layoutAttributesForItem(at: indexPath) {
                frame = cellAttributes.frame
            }
        }

        let layoutAttributes = ColoredViewLayoutAttributes(forDecorationViewOfKind: elementKind, with: indexPath)
        layoutAttributes.frame = CGRect(x: frame.minX, y: frame.maxY, width: frame.width, height: separatorWidth)
        layoutAttributes.zIndex = -1
        if let rows = collectionView?.numberOfItems(inSection: 0), indexPath.row >= rows {
            layoutAttributes.color = UIColor.clear
        } else {
            layoutAttributes.color = separatorColor
        }
        print(layoutAttributes)
        print(layoutAttributes.description)
        print(layoutAttributes.debugDescription)
        
        return layoutAttributes
    }
    
    override public func layoutAttributesForElements(in rect: CGRect)
        -> [UICollectionViewLayoutAttributes]? {
            guard let baseLayoutAttributes = super.layoutAttributesForElements(in: rect) else {
                return nil
            }
            
            var layoutAttributes = baseLayoutAttributes
            baseLayoutAttributes.filter { $0.representedElementCategory == .cell }.forEach {(layoutAttribute) in
                
                if let attr = layoutAttributesForDecorationView(ofKind: "TimelineSeparator", at: layoutAttribute.indexPath) {
                    layoutAttributes.append(attr)
                }
            }
            
            return layoutAttributes
    }
}

[...]

PR #527, opened 2017-08-22
Fix a crash in -[ASCollectionViewLayoutController elementsWithinRangeBounds:map:]

mlilek-nyt:
`-[ASElementMap elementForLayoutAttributes:]` returns nil for collection decoration views. The `allElementsForScrolling:rangeMode:displaySet:preloadSet:map:` method above nil checks this value. Do that here too.

The only other use of this method is in `-[ASCollectionLayout layoutAttributesForElementsInRect:]` and `ASCollectionLayoutSetSizeToElement` looks like it's OK with a nil value.

This was also reported as #495.

Not sure how to make a test for this - if there's an easy way to test this let me know and I'll add one.

### DIFF
--- a/Source/Details/ASCollectionViewLayoutController.m
+++ b/Source/Details/ASCollectionViewLayoutController.m
@@ -120,7 +120,10 @@ typedef struct ASRangeGeometry ASRangeGeometry;
     if (CATransform3DIsIdentity(la.transform3D) && CGRectIntersectsRect(la.frame, rangeBounds) == NO) {
       continue;
     }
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
@mlilek-nyt Thanks for the diff! Are you sure this is still an issue since we migrated from `NSMutableSet` to `NSHashTable`? NSHashTable should silently accept nil.

mlilek-nyt:
Bleh - I totally missed that change when looking at this (my project isn't on 2.4 yet). Made it locally in our version and it doesn't crash. Closing this.

PR #532, opened 2017-08-25
[ASDisplayNode] Notify rasterized subnodes that render pass has completed

smeis:
After the rendering pass of an `ASDisplayNode` has completed the `didDisplayAsyncLayer:` method is invoked which will trigger `displayDidFinish`. However, in a hierarchy that is rasterized the rendering of the subnodes is done by rasterizing them, hence `didDisplayAsyncLayer:` and `displayDidFinish` will only be invoked on the root node and not on the subnodes.

So subnodes that are in a rasterized hierarchy will not be notified that the rendering pass has completed. I came across this issue when working with `ASNetworkImageNode`. Both `ASNetworkImageNode` and `ASMultiplexImageNode` depend on `displayDidFinish` to notify their delegates that they finished displaying their image, which won't happen if they're in a rasterized tree.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -347,6 +349,12 @@ - (void)displayAsyncLayer:(_ASDisplayLayer *)asyncLayer asynchronously:(BOOL)asy
         layer.contents = (id)image.CGImage;
       }
       [self didDisplayAsyncLayer:self.asyncLayer];
+      
+      if (_flags.rasterizesSubtree) {
```
nguyenhuy: Since we no longer hold the instance lock by the time this `completionBlock` is executed, I think we should either grab this flag at the beginning of this method and use it here, or call `self.rasterizesSubtree` here.

smeis:
Thanks for your review @nguyenhuy ! I've added a commit to grab the flag with the instance lock held. 

I agree that we should balance the delegate calls. While updating the unit test for this I noticed that `displayWillStart` will be invoked twice on a node when the display process starts. It's called directly from [willDisplayAsyncLayer:asynchronously:](https://github.com/TextureGroup/Texture/blob/master/Source/ASDisplayNode.mm#L1556), this is followed by a call to `displayWillStartAsynchronously:` which will also call `displayWillStart` [here](https://github.com/TextureGroup/Texture/blob/master/Source/ASDisplayNode.mm#L1569). Is this intended behavior?

nguyenhuy:
Awesome, looking forward to the tests!

To answer your question, no it's not expected. I submitted a PR yesterday to address it: https://github.com/TextureGroup/Texture/pull/536.

smeis:
Ah yes I see! I've added a call to `-willDisplayAsyncLayer:` and extended the test.

PR #533, opened 2017-08-25
[Gallery layout] Include the caller in properties providing methods

nguyenhuy:
- This is the convention for delegate methods.
- Clients can use the caller to decide which value to return, for example when they have multiple layouts using the same provider.


Source/ASPagerNode.m
```diff
@@ -148,7 +148,7 @@ - (NSInteger)indexOfPageWithNode:(ASCellNode *)node
 
 #pragma mark - ASCollectionGalleryLayoutPropertiesProviding
 
-- (CGSize)sizeForElements:(ASElementMap *)elements
+- (CGSize)galleryLayoutDelegate:(nonnull ASCollectionGalleryLayoutDelegate *)delegate sizeForElements:(nonnull ASElementMap *)elements
```
maicki: Will this be a breaking API change or do we see the `ASCollectionGalleryLayout` as beta or not announced yet?

nguyenhuy: Yes, the whole async collection layout API is still in beta.

PR #536, opened 2017-08-28
[ASDisplayNode] Deprecate -displayWillStart in favor of -displayWillStartAsynchronously:

nguyenhuy:
Follow up on [#2344](https://github.com/facebookarchive/AsyncDisplayKit/pull/2344/).


Source/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -332,7 +332,7 @@ - (void)displayAsyncLayer:(_ASDisplayLayer *)asyncLayer asynchronously:(BOOL)asy
     return;
   }
   
-  ASDisplayNodeAssert(_layer, @"Expect _layer to be not nil");
+  ASDisplayNodeAssert(layer, @"Expect _layer to be not nil");
```
nguyenhuy: The instance lock is no longer held by the time this line is reached. It's safer this way.

PR #538, opened 2017-08-29
ASCollectionGalleryLayoutDelegate supplementary node support 

NikoSahradyan:


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Details/ASCollectionGalleryLayoutDelegate.mm
```diff
@@ -105,6 +110,31 @@ + (ASCollectionLayoutState *)calculateLayoutWithContext:(ASCollectionLayoutConte
   NSMutableArray<_ASGalleryLayoutItem *> *children = ASArrayByFlatMapping(elements.itemElements,
                                                                           ASCollectionElement *element,
                                                                           [[_ASGalleryLayoutItem alloc] initWithItemSize:itemSize collectionElement:element]);
+  NSInteger numberOfSupplementaryelementsBefore = 0;;
+  if ([context.elements.supplementaryElementKinds count]) {
```
nguyenhuy: `context.elements` is stored in an `elements` variable at the beginning of this method. For readability, please use it throughout this method instead of reaching out to the context every time.

Source/Details/ASCollectionGalleryLayoutDelegate.mm
```diff
@@ -105,6 +110,31 @@ + (ASCollectionLayoutState *)calculateLayoutWithContext:(ASCollectionLayoutConte
...
+    for (int sectionIndex = 0; sectionIndex< context.elements.numberOfSections; ++sectionIndex) {
+      for (NSString *kind in context.elements.supplementaryElementKinds) {
+        if ([kind isEqualToString:UICollectionElementKindSectionHeader]) {
+          ASCollectionElement *headerElement = [context.elements supplementaryElementOfKind:UICollectionElementKindSectionHeader atIndexPath:[NSIndexPath indexPathForRow:0 inSection:sectionIndex]];
+          if (headerElement) {
+            NSInteger index = sectionIndex > 0 ? [context.elements numberOfItemsInSection:sectionIndex - 1] : 0;
```
nguyenhuy: I believe this `index` won't be valid for the 3rd section and below. The reason is that `numberOfItemsInSection` returns the number in the previous section, not the total number from the 1st section to the previous one.

I understand that you're trying to inject header and footer elements at the right positions after all item elements were generated above via `ASArrayByFlatMapping`. While this approach should work, I believe it's a bit over-complicated, which will affect readability and maintainability in the long term. How about dropping the call to `ASArrayByFlatMapping` completely and generate the children array from the ground up? That is, for each section, add headers first, then items and finally footers, and only then, move on to the next section.

PR #553, opened 2017-09-06
Bugfix/text trunctaion for textnode2

wsdwsd0829:
[ASTextLayout layoutWithContainer:container text:text]; is called for same text with different container size for a complete layout pass. 

The existing caching mechanism is over confident. 

For example "A long string that happened to fit whole screen” is set as layout result when container is full screen size on first round. 

Later another narrower width constraint size is tried for same text. This should trigger an new layout calculation while previous cached is sized returned instead; 

This fix will only return cached result only when constraint size are same for different layouts. 

Another side notes: the truncations rules are remove the last word that caused truncation and apply  “…” to the previous word of that word;  (This make sense in case last word is less than 2 chars which will cause “…” overflow.) The downside however is that if the last word is long and removed, the result rendered would look much shorter then expected.


wsdwsd0829:
@appleguy

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

wsdwsd0829:
Hi I made a sample to reproduce the problem. 
https://github.com/wsdwsd0829/Texture/tree/bugfix/text-node-2-example
examples/TextNodeExample/Sample/TextCellNode.m the yoga layout works for ASTextNode but not ASTextNode2. I printed out cache related size for comparisons. 
My fix may not be best, let me know how we can better fix the problem. Thanks. 
image . 331 is the _label1 after truncation without "..." is used from cache (which we should not)

nguyenhuy:
@Adlai-Holler @wsdwsd0829 Please follow up on this. Thanks!

Adlai-Holler:
@wsdwsd0829 This diff was actually landed incidentally in #918. I've created a followup fix that retains the correctness win but reduces the performance impact. Sorry for taking so long to really dig into this!

PR #554, opened 2017-09-07
Added attributed versions of accessibilityLabel, accessibilityHint, accessibilityValue

fruitcoder:
First version of how we might add attributed string functionality to accessibility labels, hints and values. The properties are only available from iOS 11 and tv 11 and up so I added `API_AVAILABLE` to the properties in `ASDisplayNode`.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/de01254af0c2e36f041ebf643f1e3ea4/raw/5ab3148c01393d849fd04f404e3a7ae80253126e/log.txt)

Source/ASDisplayNode.h
```diff
@@ -716,8 +716,11 @@ extern NSInteger const ASDefaultDrawingPriority;
 // Accessibility support
 @property (nonatomic, assign)           BOOL isAccessibilityElement;
 @property (nonatomic, copy, nullable)   NSString *accessibilityLabel;
+@property (nullable, nonatomic, copy)   NSAttributedString *accessibilityAttributedLabel API_AVAILABLE(ios(11.0),tvos(11.0));
```
nguyenhuy: To follow the existing convention, let's put `nullable` last. So `(nonatomic, copy, nullable)`

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -169,8 +174,20 @@ static void CollectAccessibilityElementsForContainer(ASDisplayNode *container, _
...
 
   SortAccessibilityElements(labeledNodes);
-  NSArray *labels = [labeledNodes valueForKey:@"accessibilityLabel"];
-  accessiblityElement.accessibilityLabel = [labels componentsJoinedByString:@", "];
+
+  if (@available(iOS 11.0, *)) {
+    NSArray *attributedLabels = [labeledNodes valueForKey:@"accessibilityAttributedLabel"];
+    NSMutableAttributedString *attributedLabel = [NSMutableAttributedString new];
+    [attributedLabels enumerateObjectsUsingBlock:^(id  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+      if (idx != 0)
```
nguyenhuy: The convention is to use curly brackets regardless of the number of statements followed.

Source/Private/_ASPendingState.mm
```diff
@@ -586,9 +595,26 @@ - (NSString *)accessibilityLabel
 
 - (void)setAccessibilityLabel:(NSString *)newAccessibilityLabel
 {
-  _flags.setAccessibilityLabel = YES;
   if (accessibilityLabel != newAccessibilityLabel) {
```
nguyenhuy: Hm, are we comparing strings here?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -901,6 +901,22 @@ - (void)setAccessibilityLabel:(NSString *)accessibilityLabel
...
+{
+  _bridge_prologue_read;
+  return _getAccessibilityFromViewOrProperty(_accessibilityAttributedLabel, accessibilityAttributedLabel);
+}
+
+- (void)setAccessibilityAttributedLabel:(NSAttributedString *)accessibilityAttributedLabel
+{
+  _bridge_prologue_write;
+  { _setAccessibilityToViewAndProperty(_accessibilityAttributedLabel, accessibilityAttributedLabel, accessibilityAttributedLabel, accessibilityAttributedLabel); }
+  { _setAccessibilityToViewAndProperty(_accessibilityLabel, accessibilityAttributedLabel.string, accessibilityLabel, accessibilityAttributedLabel.string); }
```
nguyenhuy: This looks a bit strange. Are attributed labels supposed to overwrite non-attributed ones? Since these attributed labels are still in beta, I guess we don't know yet?

nguyenhuy:
Also, please update CHANGELOG and make sure CI builds. Thanks!

Source/Private/ASDisplayNode+UIViewBridge.mm
fruitcoder: This is in fact the behaviour I'm observing. It's like UILabel's `attributedText` and `text` attributes, which override themselves (confirmed in Playground with XC9 GM).

Source/Details/_ASDisplayViewAccessiblity.mm
fruitcoder: Sorry I thought I followed convention because I saw this in _PendingState.mm a lot for single statements (L375, L377), but I'll change my code to use curly braces.

nguyenhuy: Yeah, we need to have it written down in `CONTRIBUTING.md`. Others may disagree, but I personally follow this convention (as much as possible).

Source/Private/ASDisplayNode+UIViewBridge.mm
nguyenhuy: Cool by me then :)

Source/Details/_ASDisplayViewAccessiblity.mm
fruitcoder: I wouldn't change it for `applyToView:withSpecialPropertiesHandling:` though since every flag check seems to skip the curly braces

Source/Private/_ASPendingState.mm
fruitcoder: Yes, but this is from the original implementation: https://github.com/fruitcoder/Texture/blob/master/Source/Private/_ASPendingState.mm#L590

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/c3c1876fcd7a3c60f720884c11d67a2a/raw/395249bfcafaf018d2edbfb0e6473a12c0e11c55/log.txt)

Source/Details/UIView+ASConvenience.h
```diff
@@ -82,12 +82,15 @@ NS_ASSUME_NONNULL_BEGIN
...
  @property (nonatomic, assign)           BOOL isAccessibilityElement;
- @property (nonatomic, copy)             NSString *accessibilityLabel;
- @property (nonatomic, copy)             NSString *accessibilityHint;
- @property (nonatomic, copy)             NSString *accessibilityValue;
+ @property (nonatomic, copy, nullable)   NSString *accessibilityLabel;
+ @property (nullable, nonatomic, copy)   NSAttributedString *accessibilityAttributedLabel API_AVAILABLE(ios(11.0),tvos(11.0));
+ @property (nonatomic, copy, nullable)   NSString *accessibilityHint;
+ @property (nullable, nonatomic, copy)   NSAttributedString *accessibilityAttributedHint API_AVAILABLE(ios(11.0),tvos(11.0));
+ @property (nonatomic, copy, nullable)   NSString *accessibilityValue;
+ @property (nullable, nonatomic, copy)   NSAttributedString *accessibilityAttributedValue API_AVAILABLE(ios(11.0),tvos(11.0));
```
nguyenhuy: Please follow the `(nonatomic, copy, nullable)` convention for changes in this file.

Source/Private/_ASPendingState.mm
nguyenhuy: Yes, I see you follow this implementation throughout your changes. I should have asked better questions, sorry. 

Is there any reason to compare memory addresses here? Is the original implementation correct/acceptable?

fruitcoder: TBO I would compare those Strings with `isEqualToString:`, since if they are equal (by means of isEqualToString), we don't have to do anything.

fruitcoder:
The CI Build fails can all be categorized in 2 errors in _ASPendingState.mm:
`if (@available(iOS 11.0, *))` unexpected '@' in program -> affects all @availibility checks
`property 'accessibilityAttributedHint' not found on object of type 'UIView *'` -> affects all accesses on the new attributed versions of accessibility label/hint/value.

I don't know how to resolve those issues. Which Xcode are you using on your CI server?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/3a480b7d42a8d453b75aeb17ba2a31e6/raw/d2fa2b576b732a12ab817c1f856a1b2531f2fe6b/log.txt)

nguyenhuy:
@fruitcoder We're running Xcode 8. @garrettmoon and I agreed that we shouldn't upgrade to Xcode 9 until it's out of GM. And even so, using `@available` in the framework will force people to use Xcode 9. So this PR either needs to wait for a while, or switch to `AS_AT_LEAST_IOS11` provided in `ASAvailability.h` for the time being. There will come a time when we migrate to the new stuff, just not very soon I'm afraid.

Source/Private/_ASPendingState.mm
nguyenhuy: I agree. I'd do it the right way and not concern about the perf of `isEqualToString`, compared to `==`, especially because those labels aren't changed often I believe. Would you mind take this opportunity to do an audit in this file and switch to the formally correct compare methods?

Adlai-Holler: The existing code makes some sense, since a common use case for accessibility is to use constants for them.

The pointer compare made more sense when (1) CPUs were way slower and (2) this was an in-house framework.

I agree we should change to use `isEqualToString:`. 

Also just below here, we don't need to explicitly copy the newAccessibilityLabel before passing it to `initWithString:`. That initializer copies the contents ("initialized with the characters of `aString`").

fruitcoder:
Totally understand, I‘ll update my code to run on XCode 8 on Monday

nguyenhuy:
Great. Thank you!

fruitcoder:
I'm not too happy about the runtime code with `setValue:forKey:` and `valueForKey:` but it's the shortest I could come up with. I added two macros which use the runtime accessors/setters for the attributed versions, but you _could_ still use them without checking the API Version and using bad keys. Any alternative ideas?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -901,6 +911,28 @@ - (void)setAccessibilityLabel:(NSString *)accessibilityLabel
 {
   _bridge_prologue_write;
   _setAccessibilityToViewAndProperty(_accessibilityLabel, accessibilityLabel, accessibilityLabel, accessibilityLabel);
+  if (AS_AT_LEAST_IOS11) {
+    _accessibilityAttributedLabel = [[NSAttributedString alloc] initWithString:accessibilityLabel];
+    BOOL shouldApply = ASDisplayNodeShouldApplyBridgedWriteToView(self);
```
nguyenhuy: Use `_setAttributedAccessibilityToViewAndProperty` here?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -901,6 +911,28 @@ - (void)setAccessibilityLabel:(NSString *)accessibilityLabel
...
+{
+  _bridge_prologue_read;
+  return _getAttributedAccessibilityFromViewOrProperty(_accessibilityAttributedLabel, @"accessibilityAttributedLabel");
+}
+
+- (void)setAccessibilityAttributedLabel:(NSAttributedString *)accessibilityAttributedLabel
+{
+  _bridge_prologue_write;
+  { _setAttributedAccessibilityToViewAndProperty(_accessibilityAttributedLabel, accessibilityAttributedLabel, @"accessibilityAttributedLabel", accessibilityAttributedLabel); }
+  { _setAttributedAccessibilityToViewAndProperty(_accessibilityLabel, accessibilityAttributedLabel.string, @"accessibilityLabel", accessibilityAttributedLabel.string); }
```
nguyenhuy: Shouldn't we use `_setAccessibilityToViewAndProperty` for `_accessibilityLabel`?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -913,6 +945,22 @@ - (void)setAccessibilityHint:(NSString *)accessibilityHint
 {
   _bridge_prologue_write;
   _setAccessibilityToViewAndProperty(_accessibilityHint, accessibilityHint, accessibilityHint, accessibilityHint);
+  if (AS_AT_LEAST_IOS11) {
+    _setAttributedAccessibilityToViewAndProperty(_accessibilityAttributedHint, [[NSAttributedString alloc] initWithString:accessibilityHint], @"accessibilityAttributedHint", [[NSAttributedString alloc] initWithString:accessibilityHint]);
```
nguyenhuy: Nit: For readability and a tiny perf win, cache `[[NSAttributedString alloc] initWithString:accessibilityHint]` instead of allocating 2 identical objects.

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -913,6 +945,22 @@ - (void)setAccessibilityHint:(NSString *)accessibilityHint
...
+{
+  _bridge_prologue_read;
+  return _getAttributedAccessibilityFromViewOrProperty(_accessibilityAttributedHint, @"accessibilityAttributedHint");
+}
+
+- (void)setAccessibilityAttributedHint:(NSAttributedString *)accessibilityAttributedHint
+{
+  _bridge_prologue_write;
+  { _setAttributedAccessibilityToViewAndProperty(_accessibilityAttributedHint, accessibilityAttributedHint, @"accessibilityAttributedHint", accessibilityAttributedHint); }
+  { _setAttributedAccessibilityToViewAndProperty(_accessibilityHint, accessibilityAttributedHint.string, @"accessibilityHint", accessibilityAttributedHint.string); }
```
nguyenhuy: Same as above, use `_setAccessibilityToViewAndProperty` for `_accessibilityHint`.

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -925,6 +973,22 @@ - (void)setAccessibilityValue:(NSString *)accessibilityValue
 {
   _bridge_prologue_write;
   _setAccessibilityToViewAndProperty(_accessibilityValue, accessibilityValue, accessibilityValue, accessibilityValue);
+  if (AS_AT_LEAST_IOS11) {
+    _setAttributedAccessibilityToViewAndProperty(_accessibilityAttributedValue, [[NSAttributedString alloc] initWithString:accessibilityValue], @"accessibilityAttributedValue", [[NSAttributedString alloc] initWithString:accessibilityValue]);
```
nguyenhuy: Nit: Same here, cache the attributed string.

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -925,6 +973,22 @@ - (void)setAccessibilityValue:(NSString *)accessibilityValue
...
+{
+  _bridge_prologue_read;
+  return _getAttributedAccessibilityFromViewOrProperty(_accessibilityAttributedValue, @"accessibilityAttributedValue");
+}
+
+- (void)setAccessibilityAttributedValue:(NSAttributedString *)accessibilityAttributedValue
+{
+  _bridge_prologue_write;
+  { _setAttributedAccessibilityToViewAndProperty(_accessibilityAttributedValue, accessibilityAttributedValue, @"accessibilityAttributedValue", accessibilityAttributedValue); }
+  { _setAttributedAccessibilityToViewAndProperty(_accessibilityValue, accessibilityAttributedValue.string, @"accessibilityValue", accessibilityAttributedValue.string); }
```
nguyenhuy: Use `_setAccessibilityToViewAndProperty`.

nguyenhuy:
@fruitcoder I think it's a good approach for the time being. These new macros are private so I wouldn't worry much about them being used without version checks.

Source/Private/ASDisplayNode+UIViewBridge.mm
fruitcoder: Hm, I wonder if one of the tests should have caught that. But you're definitely right!

nguyenhuy:
Landing time 🛬 Thanks, @fruitcoder!

PR #555, opened 2017-09-07
Make ASWeakMapEntry Value Atomic

Adlai-Holler:
We are seeing a variety of memory corruption crashes around the image values in the ASImageNode display pipeline.

It could be unrelated, but there exists a race on the `value` property of ASWeakMapEntry which is implemented `nonatomic` but is read/written from different threads.


Source/Private/ASWeakMap.m
```diff
@@ -19,6 +19,7 @@
 
 @interface ASWeakMapEntry ()
 @property (nonatomic, strong) NSObject *key;
```
nguyenhuy: May worth making `key` read-only.

Adlai-Holler:
I went a little overboard and also changed the `NSObject *` declarations to `id`. Outside the scope of this diff but we almost never touch these files and using `NSObject *` when it's not needed is tragic.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Table and collection views to consider content inset when calculating (default) element size range [Huy Nguyen](https://github.com/nguyenhuy) [#525](https://github.com/TextureGroup/Texture/pull/525)
 - [ASEditableTextNode] added -editableTextNodeShouldBeginEditing to ASEditableTextNodeDelegate to mirror the corresponding method from UITextViewDelegate. [Yan S.](https://github.com/yans) [#535](https://github.com/TextureGroup/Texture/pull/535)
 - [Breaking] Remove APIs that have been deprecated since 2.0 and/or for at least 6 months [Huy Nguyen](https://github.com/nguyenhuy) [#529](https://github.com/TextureGroup/Texture/pull/529)
+- Fixed a memory corruption issue in the ASImageNode display system. [Adlai Holler](https://github.com/Adlai-Holler) [#55](https://github.com/TextureGroup/Texture/pull/555)
```
nguyenhuy: Nit: "#555"

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b5836abfbb1eb0bc4c3e2e225999b8a9/raw/42d9ed2ba4ba46f0baf94480d478ad5a12e96900/log.txt)

PR #556, opened 2017-09-07
Ensure ASWeakMapEntry's value is atomic

nguyenhuy:
It can be read and written on multiple threads. That may cause a crash that we're seeing in the Pinterest app.

screen shot 2017-09-07 at 7 37 04 pm


nguyenhuy:
@Adlai-Holler won the race! https://github.com/TextureGroup/Texture/pull/555/files

PR #558, opened 2017-09-09
Mark ASRunLoopQueue as drained if it contains only NULLs

### LINKED ISSUE #518: [ASRunLoopQueue] `processQueue` in ASRunLoopQueue method never Early-exit making CPU usage high
screen shot 2017-08-18 at 8 17 09 pm

I'm facing a problem that the `processQueue` method is never Early-exit so that the CPU usage is always 100%.

As you can see in the attachment pic, the count of `_internalQueue` is 3, and all its values are NULL. But the `[_internalQueue compact]` never worked since the `_internalQueue` has a flag of `needsCompaction = false`.

https://github.com/TextureGroup/Texture/blob/cae9517ebe48b68f4e2f5bccfe7c03c207596e07/Source/ASRunLoopQueue.mm#L385

I doubt that if it is because some object enqueue is a weak pointer? Since `NSPointerArray` has a problem that it will not handle weak pointer correctly. See [here](http://www.openradar.me/15396578).

cesteban:
I wrote a failing tests for [#518](https://github.com/TextureGroup/Texture/issues/518). Then made it pass.

Apparently, the problem is `NSPointerArray` won't be marked as `needsCompaction` when stored weak pointers become NULL. This makes subsequent calls to `compact` to return early without effectively removing NULL's. As a result `ASRunLoopQueue` is never marked as drained, event though there is no more work to do.

I added an `isEmpty` public property to `ASRunLoopQueue` which is only used in tests. That's not pretty, but I went this way because IMO the property makes sense in an API like that (if you have a queue, it's very natural to have a way to know if it's empty). If you don't want to have this property, we need to find out another way for the test to _sense_ that the queue has been marked as drained.

Finally, I believe `ASRunLoopQueue` should have a lot more tests, given the importance of the component. I would try to find the time to write a bunch of them, but that should be a different branch IMO.

Please, let me know your thoughts.

Edit (huy): Closes #518.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fixed a memory corruption issue in the ASImageNode display system. [Adlai Holler](https://github.com/Adlai-Holler) [#555](https://github.com/TextureGroup/Texture/pull/555)
 - [Breaking] Rename ASCollectionGalleryLayoutSizeProviding to ASCollectionGalleryLayoutPropertiesProviding. Besides a fixed item size, it now can provide interitem and line spacings, as well as section inset [Huy Nguyen](https://github.com/nguyenhuy) [#496](https://github.com/TextureGroup/Texture/pull/496) [#533](https://github.com/TextureGroup/Texture/pull/533)
 - Deprecate `-[ASDisplayNode displayWillStart]` in favor of `-displayWillStartAsynchronously:` [Huy Nguyen](https://github.com/nguyenhuy) [536](https://github.com/TextureGroup/Texture/pull/536)
+- Mark ASRunLoopQueue as drained if it contains only NULLs [Cesar Estebanez](https://github.com/cesteban)
 
 ##2.4
 - Fix an issue where inserting/deleting sections could lead to inconsistent supplementary element behavior. [Adlai Holler](https://github.com/Adlai-Holler)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 - Fixed a memory corruption issue in the ASImageNode display system. [Adlai Holler](https://github.com/Adlai-Holler) [#555](https://github.com/TextureGroup/Texture/pull/555)
 - [Breaking] Rename ASCollectionGalleryLayoutSizeProviding to ASCollectionGalleryLayoutPropertiesProviding. Besides a fixed item size, it now can provide interitem and line spacings, as well as section inset [Huy Nguyen](https://github.com/nguyenhuy) [#496](https://github.com/TextureGroup/Texture/pull/496) [#533](https://github.com/TextureGroup/Texture/pull/533)
 - Deprecate `-[ASDisplayNode displayWillStart]` in favor of `-displayWillStartAsynchronously:` [Huy Nguyen](https://github.com/nguyenhuy) [536](https://github.com/TextureGroup/Texture/pull/536)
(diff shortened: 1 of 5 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRunLoopQueueTests.m:</p>
<pre><code>//
//  ASRunLoopQueueTests.m
//  Texture
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

cesteban:
I finally covered ASRunLoopQueue with tests.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRunLoopQueueTests.m:</p>
<pre><code>//
//  ASRunLoopQueueTests.m
//  Texture
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Fixed a memory corruption issue in the ASImageNode display system. [Adlai Holler](https://github.com/Adlai-Holler) [#555](https://github.com/TextureGroup/Texture/pull/555)
 - [Breaking] Rename ASCollectionGalleryLayoutSizeProviding to ASCollectionGalleryLayoutPropertiesProviding. Besides a fixed item size, it now can provide interitem and line spacings, as well as section inset [Huy Nguyen](https://github.com/nguyenhuy) [#496](https://github.com/TextureGroup/Texture/pull/496) [#533](https://github.com/TextureGroup/Texture/pull/533)
 - Deprecate `-[ASDisplayNode displayWillStart]` in favor of `-displayWillStartAsynchronously:` [Huy Nguyen](https://github.com/nguyenhuy) [536](https://github.com/TextureGroup/Texture/pull/536)
+- Mark ASRunLoopQueue as drained if it contains only NULLs [Cesar Estebanez](https://github.com/cesteban)
```
nguyenhuy: Nit: Please add the PR's URL.

nguyenhuy:
Also, please update the license in `ASRunLoopQueueTests.m` as suggested by Danger above.

CHANGELOG.md
cesteban: Yup, sorry!

cesteban:
@nguyenhuy Actually, my first implementation that passed the test was something similar to this:

```objective-c
    NSInteger foundItemCount = 0;
    for (NSInteger i = 0; i < internalQueueCount && foundItemCount < maxCountToProcess; i++) {
      //...
    }

    if (foundItemCount == 0) {
      // If _internalQueue holds weak references, and all of them just become nil, then the array
      // is never marked as needsCompletion, and compact will return early, not removing the nils.
      // Inserting a nil here ensures the compaction will take place.
      [_internalQueue addPointer:nil];
    }

    [_internalQueue compact];
    if (_internalQueue.count == 0) {
      isQueueDrained = YES;
    }
```

Do you prefer this kind of approach? I can recover it, just let me know.

nguyenhuy:
@cesteban Yes, I think it's a cleaner approach. May worth adding the StackOverflow and OpenRadar links.

cesteban:
Thanks @nguyenhuy. There we go!

PR #562, opened 2017-09-11
[ASCollectionLayout] Exclude content inset on scrollable directions from viewport size

nguyenhuy:
Failing to do this will cause `ASCollectionLayout` to report a content size that spans beyond the safe scrollable area of the collection node. In case of `ASPagerNode`, that then causes the pager to bounce vertically and messes up with the pull-to-dismiss gesture in closeup.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/bc564ad7b6de95f5f07edc5ab65522b4/raw/774f959a70f2d194c14b421e3f54858ebcc562c6/log.txt)

PR #563, opened 2017-09-12
[ASDisplayNode] Assert when attempt to remove a subnode from a supernode that has ASM enabled #trivial

nguyenhuy:
If the removal proceeds, `_subnodes` and `_calculatedDisplayNodeLayout` of the supernode will be out-of-sync. Subsequence layouts applied to that node may cause range exception.

screen shot 2017-09-12 at 6 04 03 pm

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -2281,6 +2281,15 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
 {
   ASDisplayNodeLogEvent(self, @"removeFromSupernode with automaticallyManagesSubnodes: %@",
                         self.automaticallyManagesSubnodes ? @"YES" : @"NO");
+
+#if ASDISPLAYNODE_ASSERTIONS_ENABLED
+  __instanceLock__.lock();
+    __weak ASDisplayNode *supernode = _supernode;
+  __instanceLock__.unlock();
+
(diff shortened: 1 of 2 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/092c56e8b49c3a914114ab73997dc7dc/raw/ff5046ddca4c5313bc2bde37d422f039356f2867/log.txt)

Source/ASDisplayNode.mm
```diff
@@ -2281,6 +2281,15 @@ - (void)removeFromSupernode
 {
   ASDisplayNodeLogEvent(self, @"removeFromSupernode with automaticallyManagesSubnodes: %@",
                         self.automaticallyManagesSubnodes ? @"YES" : @"NO");
+
+#if ASDISPLAYNODE_ASSERTIONS_ENABLED
+  __instanceLock__.lock();
+    __weak ASDisplayNode *supernode = _supernode;
+  __instanceLock__.unlock();
+
+  ASDisplayNodeAssert(! supernode.automaticallyManagesSubnodes, @"Should not manually remove node %@ from its supernode %@ which has automaticallyManagesSubnodes enabled. Please talk to Huy if you're hitting this in closeup!", self, supernode);
```
maicki: We should maybe not add your name in here. What about either using the github name or say something like: "Open an issue etc."

nguyenhuy: The last sentence is meant for P engineers only :)

nguyenhuy:
Gonna close this for now as the problem is more complicated than this. It's probably a race condition in `ASDisplayNode`. Imagine a subnode is being inserted on a thread while another one is being removed on another thread. Since the internal insert methods grab  `__instanceLock__` multiple times, it's possible that the removing thread steals the lock in between and changes `_subnodes` array, which then causes the `subnodeIndex` to be incorrect.

PR #569, opened 2017-09-15
Update yoga version

wsdwsd0829:


wsdwsd0829:
need some of changes like 
image 
which solves problems like unnecessary layout passes caused by text rounding problem for @2x @3x

nguyenhuy:
Refer to @appleguy for a final say here :)

Issue #14, opened 2017-04-15
snapchat esque swipe ui demo sample


markmark1:
is there a snapchat like swipe based ui to swipe through videos  available thanks

markmark1:
is this asyncdisplaykit?

Adlai-Holler:
Hey @markmark1 yes this is the framework formerly known as AsyncDisplayKit but unfortunately the core team can't field questions like this. Head over to the [Texture slack](http://texturegroup.org/slack.html) and ask in the #general channel, there's a ton of people helping one another.

markmark1:
Tried slack didn't have much luck do you mind pointing me in the right direction Adlai Holler
		

		
	





On Mon, Apr 17, 2017 at 11:55 AM -0700, "Adlai Holler" <notifications@github.com> wrote:












Hey @markmark1 yes this is the framework formerly known as AsyncDisplayKit but unfortunately the core team can't field questions like this. Head over to the Texture slack and ask in the #general channel, there's a ton of people helping one another.

Adlai-Holler:
@markmark1 You should check out ASPagerNode to create a horizontal-swiping list, and then ASVideoPlayerNode to show the videos.

markmark1:
Thanks so much you rock Adlai 
		

		
	





On Mon, Apr 17, 2017 at 4:30 PM -0700, "Adlai Holler" <notifications@github.com> wrote:












@markmark1 You should check out ASPagerNode to create a horizontal-swiping list, and then ASVideoPlayerNode to show the videos.

markmark1:
What about images and overlays are there texture modules for those too?
		

		
	





On Mon, Apr 17, 2017 at 4:30 PM -0700, "Adlai Holler" <notifications@github.com> wrote:












@markmark1 You should check out ASPagerNode to create a horizontal-swiping list, and then ASVideoPlayerNode to show the videos.

markmark1:
What's a node
		

		
	





On Mon, Apr 17, 2017 at 4:30 PM -0700, "Adlai Holler" <notifications@github.com> wrote:












@markmark1 You should check out ASPagerNode to create a horizontal-swiping list, and then ASVideoPlayerNode to show the videos.

Adlai-Holler:
http://texturegroup.org/docs/getting-started.html Godspeed on your project 🙂

markmark1:
What about images and overlays 
		

		
	





[...]

PR #577, opened 2017-09-20
Fix crashes caused by failing to unlock or destroy a static mutex while the app is being terminated

nguyenhuy:
This is an interesting crash. Here is an example report in Pinterest app, happening on the static `cacheLock` mutex of `ASImageNode`:

screen shot 2017-09-20 at 7 09 29 pm
screen shot 2017-09-20 at 7 09 42 pm

There are other similar crashes related to `__staticMutex` in `ASTextKitContext` reported in https://github.com/TextureGroup/Texture/issues/136. For instance, see thread #0 and #5 in [this report](http://crashes.to/s/7d0c464960f), or thread #0 and #14 [here](http://crashes.to/s/e43ab8f7d5c).

Basically, what's happening is that all static mutexes are destroyed when an app is terminated by the system. At the same time, a static mutex is being unlocked or destroyed on another thread. The unlock or destruction yields a non-zero error code, which triggers an assertion, even in production.

A similar issue was discussed and fixed in Realm-Core (https://github.com/realm/realm-core/pull/2243). For Texture, I went with a different approach that is arguably safer, IMHO. This should fix https://github.com/TextureGroup/Texture/issues/136.

> Mutex initialization using the PTHREAD_MUTEX_INITIALIZER does not immediately initialize the mutex. Instead, on first use, pthread_mutex_timedlock_np() or pthread_mutex_lock() or pthread_mutex_trylock() branches into a slow path and causes the initialization of the mutex. Because a mutex is not just a simple memory object and requires that some resources be allocated by the system, an attempt to call pthread_mutex_destroy() or pthread_mutex_unlock() on a mutex that was statically initialized using PTHREAD_MUTEX_INITIALIZER and was not yet locked causes an EINVAL error.

(https://www.ibm.com/support/knowledgecenter/en/ssw_i5_54/apis/users_65.htm)

### DIFF
--- a/Source/Details/ASBasicImageDownloader.mm
+++ b/Source/Details/ASBasicImageDownloader.mm
@@ -46,42 +46,29 @@ NSString * const kASBasicImageDownloaderContextCompletionBlock = @"kASBasicImage
 @implementation ASBasicImageDownloaderContext
 
 static NSMutableDictionary *currentRequests = nil;
-static ASDN::StaticMutex currentRequestsLock = ASDISPLAYNODE_MUTEX_INITIALIZER;
+// Allocate currentRequestsLock on the heap to prevent destruction at app exit (https://github.com/TextureGroup/Texture/issues/136)
+static ASDN::StaticMutex& currentRequestsLock = *new ASDN::StaticMutex;
 
 + (ASBasicImageDownloaderContext *)contextForURL:(NSURL *)URL
 {
-  currentRequestsLock.lock();
-    if (!currentRequests) {
-      currentRequests = [[NSMutableDictionary alloc] init];
-    }
-    ASBasicImageDownloaderContext *context = currentRequests[URL];
-    if (!context) {
-      context = [[ASBasicImageDownloaderContext alloc] initWithURL:URL];
-      currentRequests[URL] = context;
-    }
-  int unlockCode = currentRequestsLock.unlock();
-
-  if (unlockCode != 0) {
-    // Failed to unlock currentRequestsLock static mutex.
-    // Since the mutex was locked just a few lines above, this usually means there is a race condition going on.
-    // The race condition occurs when a static mutex is being destroyed as part of an app exit on main thread,
-    // and, at the same time, being used here on another thread.
-    // See https://github.com/TextureGroup/Texture/issues/136.
-    //
-    // Return nil here to signal that the image should not be loaded.
-    return nil;
+  ASDN::StaticMutexLocker l(currentRequestsLock);
+  if (!currentRequests) {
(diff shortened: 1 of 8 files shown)


Source/Details/ASBasicImageDownloader.mm
```diff
@@ -46,28 +46,42 @@ @interface ASBasicImageDownloaderContext ()
 @implementation ASBasicImageDownloaderContext
 
 static NSMutableDictionary *currentRequests = nil;
-static ASDN::RecursiveMutex currentRequestsLock;
+static ASDN::StaticMutex currentRequestsLock = ASDISPLAYNODE_MUTEX_INITIALIZER;
```
nguyenhuy: Not sure why this is a recursive mutex in the first place.

3a4oT:
@nguyenhuy thanks for the agenda and patch, very interesting. I still not understand how it triggers the assertion. Just two questions: 1 Have you reproduced it?  And when do you plan next release?;)

nguyenhuy:
@3a4oT To answer your questions:
- If a static mutex has been destroyed on the main thread during an app exit, calling `pthread_mutex_unlock()` to unlock it on a background thread causes the method to return a non-zero value. The value then triggers the assertion [here](https://github.com/TextureGroup/Texture/blob/master/Source/Details/ASThread.h#L68).
- No, I have not been able to reproduce it.
- Soon™ :P

nguyenhuy:
@Adlai-Holler Good idea. I went ahead and convert all assertions in `ASThread` to `ASDisplayNodeCAssert`.

@garrettmoon @maicki Thinking again, my approach requires more effort on client-side and so it is not scalable. I changed to the approach used in Realm.

Source/Details/ASBasicImageDownloader.mm
```diff
@@ -46,11 +46,12 @@ @interface ASBasicImageDownloaderContext ()
 @implementation ASBasicImageDownloaderContext
 
 static NSMutableDictionary *currentRequests = nil;
-static ASDN::RecursiveMutex currentRequestsLock;
+// Allocate currentRequestsLock on the heap to prevent destruction at app exit (https://github.com/TextureGroup/Texture/issues/136)
+static ASDN::StaticMutex& currentRequestsLock = *new ASDN::StaticMutex;
```
nguyenhuy: I double-checked the code here and I'm 99% sure that this mutex doesn't need to be recursive.

nguyenhuy: ^ @garrettmoon

garrettmoon: Cool :)

nguyenhuy:
Merging. Thanks, everyone!

PR #589, opened 2017-09-27
Use Nil for "no class" instead of nil #trivial

Adlai-Holler:
I wrote this code originally, and wasn't thinking.


appleguy:
@Adlai-Holler I love this diff. I remember once upon a time, I was advocating for Nil and (I think) you were annoyed at having different representations for the same value — a totally reasonable position. 

Still, pedantic correctness can be fun! lol

Adlai-Holler:
Man old me was such an IDIOT!! Glad I’ve got everything figured out now.

PR #595, opened 2017-09-29
[PINCache] Set a default .byteLimit to reduce disk usage & startup time.

appleguy:
This default is fairly low - only 20MB - but for most apps with images
in the size range of 10-50KB, this is still 400-1000 images.

Once some optimizations land to PINCache, we'll match the PINCache
default of 50MB to ensure the default better serves users with larger
objects in the cache.

Apps should preferably set their own byteLimit to an optimal value.

@garrettmoon - one interesting question for us is the best place to
set .byteLimit as an app. Digging into the ASPINRemoteImageDownloader
and doing this type cast is a bit complicated, so a passthrough API
to get the PIN* objects directly might be worthwhile.

### DIFF
--- a/Source/Details/ASPINRemoteImageDownloader.m
+++ b/Source/Details/ASPINRemoteImageDownloader.m
@@ -76,6 +76,15 @@
 @end
 #endif
 
+// Declare two key methods on PINCache objects, avoiding a direct dependency on PINCache.h
+@protocol ASPINCache
+- (id)diskCache;
+@end
+
+@protocol ASPINDiskCache
+@property (assign) NSUInteger byteLimit;
+@end
+
 @interface ASPINRemoteImageManager : PINRemoteImageManager
 @end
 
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/a10c5d20062c9c5af618daa61742e9c2/raw/53fc347b7569f0f06dfca10684ca4f0fd5239aa4/log.txt)

garrettmoon:
@appleguy Yeah, that's a good idea, but I don't want to get all my Hacktoberfest PRs done in one day :P

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -84,7 +84,18 @@ @implementation ASPINRemoteImageManager
...
 {
-  return [[PINRemoteImageManager sharedImageManager] cache];
+  static dispatch_once_t onceToken;
+  static id <PINRemoteImageCaching> cache = nil;
+  dispatch_once(&onceToken, ^{
+    cache = [[PINRemoteImageManager sharedImageManager] cache];
+
+    // Set a default byteLimit. PINCache recently implemented a 50MB default (PR #201).
+    // Ensure that older versions of PINCache also have a byteLimit applied.
+    PINDiskCache *diskCache = [ASDynamicCast(cache, PINCache) diskCache];
```
garrettmoon: I think you need to add a <PINCache/PINCache.h> include at the top.

appleguy:
Thanks! Looking into the right way to either reference, or avoid referencing PINCache as a symbol.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/22fb047c2560be4fd3c378d80db831a7/raw/525f57bfc238b112f2b82e6c698fa759a2db8f41/log.txt)

appleguy:
@garrettmoon FYI, it looks like this test failure is spurious - it doesn't reproduce on my local machine, this patch isn't related, and it will probably pass upon retry. Had to do a fake commit change to poke the server.

    ✗ testThatMultipleBatchFetchesDontHappenUnnecessarily, failed - Too many batch fetches!

maicki:
I had the same issue in a previous PR opened a couple of days ago. I had to manually retry the CI and the next time it finished successfully.

PR #597, opened 2017-10-01
update faq toc links to match the generated html id #trivial

romankl:
Right now the faq- table of contents has links to the different headlines that are mentioned in the faqs, but some of the generated id s don't match the referencing
a tag.


romankl:
Whoops, sorry somehow a lot of whitespaces change got into this PR.

PR #604, opened 2017-10-05
[ASVideoNode] Time observer fix

flovouin:
Hi guys,

I was adding a custom seek bar to a video node, and noticed that when changing the played video, I was still getting `didPlayToTimeInterval` delegate calls for the old video.

You can easily see what I mean in [this sample project](https://github.com/flovouin/Texture-VideoNodeObserver).
The video is reloaded after 2 seconds, and after that, the progress bar goes crazy, switching back and forth between the time info of the two players.
Not releasing the time observer also means that the former video keeps playing in the background, and hence seems to increase CPU consumption (twice the load on my device).

The fix is simple: actually release the time observer when removing the player. Because the observer is linked to the player, I just moved allocation/deallocation of the observer to `addPlayerObservers`/`removePlayerObservers`.

Tell me what you think,
Cheers,

Flo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Add your own contributions to the next release on the line below this with your name.
+- [ASVideoNode] Fix unreleased time observer. [Flo Vouin](https://github.com/flovouin)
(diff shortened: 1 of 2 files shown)


nguyenhuy:
@maicki  Yeap, agreed!

PR #609, opened 2017-10-10
Avoid setting zero-length string in [ASTextNode setAttributedText:]

aimobier:
I think if I'm set to nil, I don't need to create an empty object

when l in `layoutSpecThatFits` , I can use
````objective-c
if (self.someTextNode.attributedText)
/// some things
else
/// some things
````

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -428,7 +428,7 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 {
   
   if (attributedText == nil) {
-    attributedText = [[NSAttributedString alloc] initWithString:@"" attributes:nil];
+    return;
   }
   
   // Don't hold textLock for too long.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

appleguy:
This seems like a good change, although I'm sure the empty string creation was there for a reason at one point.

Is there any other logic further down in the method that performs important manipulations to the string, or other state changes to the node itself, like invalidating its layout?

aimobier:
Yes,You're quite right. I think I'm a little **舍本逐末**.
I kept this judgement when I saw this [example](https://github.com/TextureGroup/Texture/blob/master/examples/LayoutSpecExamples/Sample/LayoutExampleNodes.m#L96). I didn't realize the problem until I had an error in the layout. Maybe we can modify the code of the example so that we can not make such mistakes

nguyenhuy:
@aimobier  I think it's safer to modify the example. Thanks!

PR #610, opened 2017-10-10
[ASTextKitComponents] Temporary components can be deallocated off main #trivial

nguyenhuy:
https://github.com/TextureGroup/Texture/pull/603 unnecessarily triggers assertions when temporary components that are created [here](https://github.com/TextureGroup/Texture/blob/master/Source/TextKit/ASTextKitComponents.mm#L84) and [here](https://github.com/TextureGroup/Texture/blob/master/Source/TextKit/ASTextKitComponents.mm#L104) are deallocated off the main thread. Fix (and close https://github.com/TextureGroup/Texture/issues/607) by not asserting on these objects.

PR #598, opened 2017-10-01
ASTextKitComponents needs to be deallocated on main

maicki:
Certain objects need to be deallocated on main. We have an automatic mechanism were it goes through each ivar and checks if the object needs to be deallocated on main or not. We have to add `ASTextKitComponents` to this list of classes now.

There are multiple ways to tackle this issue, this PR just adds the class right into the method that checks for classes who should be deallocated on main. It will have the problem as we check it with a string comparison if we change the name of the ASTextKitComponents we could miss this change in here as well as if we would like to add new classes this can easily hard to maintain.

I will follow up with a PR that is more flexible approach, but it will add some overhead.

Adresses #586

PR #612, opened 2017-10-11
[ASTextKitComponents] Make sure Main Thread Checker isn't triggered during background calculations #trivial

nguyenhuy:
Currently the checker is triggered either when the components need to get the backing text view's bounds, or when the text view's delegate is cleared out during deallocation.

In the first case, I add a new subclass of UITextView that has a thread-safe bounds property and is used exclusively by ASTextKitComponents.

The second case is handled in #598. However, a new main thread assertion added in #603 causes background deallocations of temporary components to fail. #610 attempts to fix that, but since temporary components don't have a reference to a text view, it's fine to avoid the assertion altogether.

### DIFF
--- a/Source/TextKit/ASTextKitComponents.mm
+++ b/Source/TextKit/ASTextKitComponents.mm
@@ -21,6 +21,63 @@
 
 #import <tgmath.h>
 
+@interface ASTextKitComponentsTextView ()
+@property (nonatomic, assign) CGRect threadSafeBounds;
+@end
+
+@implementation ASTextKitComponentsTextView
+{
+  ASDN::Mutex __instanceLock__;
+}
+
+@synthesize threadSafeBounds = _threadSafeBounds;
+
+- (instancetype)initWithFrame:(CGRect)frame textContainer:(NSTextContainer *)textContainer
+{
+  self = [super initWithFrame:frame textContainer:textContainer];
+  if (self) {
+    _threadSafeBounds = self.bounds;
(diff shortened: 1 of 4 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ffc04724759117c1067477de82b0fca9/raw/9067f5196b8cb2342abd04d0f1c5284537aea0a9/log.txt)

Source/TextKit/ASTextKitComponents.mm
```diff
@@ -17,19 +17,65 @@
 
 #import <AsyncDisplayKit/ASTextKitComponents.h>
 #import <AsyncDisplayKit/ASAssert.h>
+#import <AsyncDisplayKit/ASThread.h>
 
 #import <tgmath.h>
 
+@interface ASTextKitComponentsTextView ()
+@property (nonatomic, assign) CGRect threadSafeBounds;
```
maicki: Should this be `atomic`?

nguyenhuy: Good point!

PR #637, opened 2017-10-24
[ASScrollNode] Fix small bugs and add unit tests

nguyenhuy:
- Currently, ASScrollNode's calculated size is based on its parent size. However, there are cases in which the size range passed to the scroll node to calculate a layout against doesn't include the parent size. In such cases, we should clamp the parent size against the size range to make sure the range is always respected.
- When scrollable directions change, invalidate the calculated layout because these directions affect the size range that is used to calculate the content's layout.
- Add unit tests to cover these bugs and other common use cases.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Tests/ASScrollNodeTests.m
```diff
@@ -0,0 +1,137 @@
...
+  XCTAssertTrue(ASSizeRangeEqualToSizeRange(_subnode.constrainedSizeForCalculatedLayout, subnodeSizeRange));
+
+  // Same test for horizontal scrollable directions
+  _scrollNode.scrollableDirections = ASScrollDirectionHorizontalDirections;
+  [_scrollNode layoutThatFits:sizeRange parentSize:parentSize];
+
+  subnodeSizeRange = sizeRange;
+  subnodeSizeRange.max.width = CGFLOAT_MAX;
+
+  XCTAssertTrue(ASSizeRangeEqualToSizeRange(_subnode.constrainedSizeForCalculatedLayout, subnodeSizeRange));
```
Adlai-Holler: Nit: Let's use `ASXCTAssertEqualSizeRanges` and `ASXCTAssertEqualSizes` where possible, so that we get better error messages. Import `ASXCTExtensions.h` to get 'em.

Tests/ASScrollNodeTests.m
```diff
@@ -0,0 +1,137 @@
...
+- (void)testAutomaticallyManagesContentSizeUnderflow
+{
+  CGSize subnodeSize = CGSizeMake(100, 100);
+  CGSize parentSize = CGSizeMake(100, 200);
+  ASSizeRange sizeRange = ASSizeRangeUnconstrained;
+
+  _subnode.style.preferredSize = subnodeSize;
+
+  ASLayout *layout = [_scrollNode layoutThatFits:sizeRange parentSize:parentSize];
+  [_scrollNode layout];
```
Adlai-Holler: Another nit: I prefer to name ivars without underscores in unit test files. It makes things more readable and it's usually safe since a new instance is created for each test method.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/aad00ce1c506e64e36fb06f246c1f6f2/raw/5c100d08a0f68d684d53dea42a8115c642cd78f8/log.txt)

Tests/ASScrollNodeTests.m
```diff
@@ -66,13 +67,13 @@ - (void)testAutomaticallyManagesContentSizeUnderflow
   CGSize parentSize = CGSizeMake(100, 200);
   ASSizeRange sizeRange = ASSizeRangeUnconstrained;
 
-  subnode.style.preferredSize = subnodeSize;
+  self.subnode.style.preferredSize = subnodeSize;
 
-  ASLayout *layout = [scrollNode layoutThatFits:sizeRange parentSize:parentSize];
-  [scrollNode layout];
+  ASLayout *layout = [self.scrollNode layoutThatFits:sizeRange parentSize:parentSize];
+  [self.scrollNode layout];
```
appleguy: It seems better (perhaps) to call [scrollNode.layer layoutIfNeeded], to test the full layout stack as it is triggered in most apps?

Tests/ASScrollNodeTests.m
```diff
@@ -16,48 +16,49 @@
...
   ASSizeRange sizeRange = ASSizeRangeMake(parentSize);
 
-  [scrollNode layoutThatFits:sizeRange parentSize:parentSize];
+  [self.scrollNode layoutThatFits:sizeRange parentSize:parentSize];
 
   ASSizeRange subnodeSizeRange = sizeRange;
   subnodeSizeRange.max.height = CGFLOAT_MAX;
-  XCTAssertEqual(scrollNode.scrollableDirections, ASScrollDirectionVerticalDirections);
-  ASXCTAssertEqualSizeRanges(subnode.constrainedSizeForCalculatedLayout, subnodeSizeRange);
+  XCTAssertEqual(self.scrollNode.scrollableDirections, ASScrollDirectionVerticalDirections);
```
appleguy: Rather than self. everywhere, using the ivar like _scrollNode would be more readable (especially in property chains & methods that use it a lot).

Adlai-Holler: As a follow-up to this, I find that using non-underscored ivar names in test classes provides the best readability.

Since the object is destroyed after `-tearDown` is run, and since test classes usually have trivial accessors, direct ivar access is safe and convenient.

PR #638, opened 2017-10-24
Fix a layout deadlock caused by holding the lock and going up the tree.

garrettmoon:


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASDisplayNode+Layout.mm
```diff
@@ -368,8 +369,10 @@ - (void)_locked_measureNodeWithBoundsIfNecessary:(CGRect)bounds
     // In this case, we need to detect that we've already asked to be resized to match this
     // particular ASLayout object, and shouldn't loop asking again unless we have a different ASLayout.
     nextLayout->requestedLayoutFromAbove = YES;
-    [self _setNeedsLayoutFromAbove];
-    // Update the layout's version here because _setNeedsLayoutFromAbove calls __setNeedsLayout which in turn increases _layoutVersion
+    __instanceLock__.unlock();
```
garrettmoon: These 3 lines are the important bits which fix the deadlock.

Source/ASDisplayNode+Layout.mm
```diff
@@ -211,7 +211,7 @@ @implementation ASDisplayNode (ASLayoutInternal)
  * @discussion The size of a root node is determined by each subnode. Calling invalidateSize will let the root node know
  * that the intrinsic size of the receiver node is no longer valid and a resizing of the root node needs to happen.
  */
-- (void)_setNeedsLayoutFromAbove
+- (void)_u_setNeedsLayoutFromAbove
 {
   as_activity_create_for_scope("Set needs layout from above");
   ASDisplayNodeAssertThreadAffinity(self);
```
nguyenhuy: Nit: Let's add the thread ownership assertion (i.e `ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__)`). We may enable it by default in the future.

Source/ASDisplayNode+Layout.mm
```diff
@@ -287,8 +287,9 @@ - (void)displayNodeDidInvalidateSizeNewSize:(CGSize)size
   }
 }
 
-- (void)_locked_measureNodeWithBoundsIfNecessary:(CGRect)bounds
+- (void)_u_measureNodeWithBoundsIfNecessary:(CGRect)bounds
 {
```
nguyenhuy: Same, let's call `ASDisplayNodeAssertLockUnownedByCurrentThread()` before acquiring the lock.

garrettmoon:
Good callouts, thank you for the review @nguyenhuy :)

Source/ASDisplayNode+Layout.mm
appleguy: @garrettmoon We do use _calculatedDisplayNodeLayout below — it looks like this would make it possible that another thread could change _calculatedDisplayNodeLayout while this is unlocked (possibly by the main thread calling -layout).

I'm wondering if we should actually be returning after this, or something else like forcing a -layoutIfNeeded. It's a bit weird that we proceed to _completePendingLayoutTransition when, in actuality, we are not yet in a valid full-tree layout state until the parent has -layoutSublayers called by CA.

This might be OK because one could argue that the layout is correct _for this node_ and its children, even if not the full tree.

PR #639, opened 2017-10-24
Have ASNetworkImageNode report whether images were cached or not

Adlai-Holler:
This should help consumers who are interested in tracking their image cache performance.

### DIFF
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -703,6 +703,7 @@
         if (_delegateFlags.delegateDidLoadImageWithInfo) {
           ASDN::MutexUnlocker u(__instanceLock__);
           ASNetworkImageNodeDidLoadInfo info = {};
+          info.imageSource = ASNetworkImageSourceFileURL;
           [delegate imageNode:self didLoadImage:self.image info:info];
         } else if (_delegateFlags.delegateDidLoadImage) {
           ASDN::MutexUnlocker u(__instanceLock__);
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -56,6 +56,7 @@
     unsigned int delegateDidFailWithError:1;
     unsigned int delegateDidFinishDecoding:1;
     unsigned int delegateDidLoadImage:1;
+    unsigned int delegateDidLoadImageWithInfo:1;
   } _delegateFlags;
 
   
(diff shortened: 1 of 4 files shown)


Source/ASNetworkImageNode.mm
```diff
@@ -353,8 +355,18 @@ - (void)displayWillStartAsynchronously:(BOOL)asynchronously
...
+          // Call out to the delegate.
+          if (_delegateFlags.delegateDidLoadImageWithInfo) {
+            ASDN::MutexUnlocker l(__instanceLock__);
+            ASNetworkImageNodeDidLoadInfo info = {};
+            info.imageSource = ASNetworkImageSourceSynchronousCache;
+            [_delegate imageNode:self didLoadImage:result info:info];
+          } else if (_delegateFlags.delegateDidLoadImage) {
+            ASDN::MutexUnlocker l(__instanceLock__);
+            [_delegate imageNode:self didLoadImage:result];
+          }
```
Adlai-Holler: We previously weren't calling out to the delegate at all in this synchronous fetch scenario. I think we should.

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -369,5 +352,29 @@ - (id)alternateRepresentationWithData:(NSData *)data options:(PINRemoteImageMana
...
+
+  if (ASDisplayNodeThreadIsMain() && queue == dispatch_get_main_queue()) {
+    work();
+  } else if (queue == nil) {
+    ASDisplayNodeFailAssert(@"Callback queue should not be nil.");
+    work();
+  } else {
+    dispatch_async(queue, work);
+  }
+}
```
Adlai-Holler: Added this method because I added another callback path and it felt silly to keep duplicating code. Sorry, dear reviewer!

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -256,51 +252,38 @@ - (nullable id)downloadImageWithURLs:(NSArray <NSURL *> *)URLs
...
+    }];
+  };
+
+  // add "IgnoreCache" option since we have a caching API so we already checked it, not worth checking again.
+  // PINRemoteImage is responsible for coalescing downloads, and even if it wasn't, the tiny probability of
+  // extra downloads isn't worth the effort of rechecking caches every single time. In order to provide
+  // feedback to the consumer about whether images are cached, we can't simply make the cache a no-op and
+  // check the cache as part of this download.
+  return [[self sharedPINRemoteImageManager] downloadImageWithURLs:URLs
+                                                           options:PINRemoteImageManagerDownloadOptionsSkipDecode | PINRemoteImageManagerDownloadOptionsIgnoreCache
```
garrettmoon: Sadly I don't think we should skip the cache because the synchronous cache check only checks the memory cache not the disk cache.

Adlai-Holler: But we'll check the disk cache in `cachedImageWithURL:callbackQueue:completion:` so I think we're OK?

garrettmoon: I take this back, didn't see that the cache check above was implemented.

garrettmoon: For history, @Adlai-Holler and I discussed and we think this is safe. He pointed out the only case where this could cause an extra download is:
```
- request0 -> cache: miss
- request0 -> download: started
- request1 -> cache: miss
- request0 -> download: finished
- request1 -> download: started
```

This seems *very* unlikely in practice.

Source/ASNetworkImageNode.mm
garrettmoon: Good catch!

PR #647, opened 2017-10-30
[Documentation] Update Inversion Docs

GeekTree0101:
Append swift example at inversion example


docs/_docs/inversion.md
```diff
@@ -24,7 +24,9 @@ When this is enabled, developers only have to take one more step to have full in
   </pre>
 
   <pre lang="swift" class = "swiftCode hidden">
```
maicki: I think you can now remove the `hidden` class now as Swift is available.

GeekTree0101: Oh I got it, thankx

PR #660, opened 2017-11-03
Make it possible to map between sections even if they're empty

Adlai-Holler:
Previously, there was no native way to map between sections. So our best shot was to pick item 0 and translate its index path, and then read the section index. This has a couple of failure modes:

- There is no item 0 in the "from" map (empty section)
- Item 0 is deleted, reloaded, or moved in the change

To fix this and provide more direct section tracking, we:
- Always create `ASSection` objects, even if the data source provides no context for them.
- Do the mapping between these. It's fast O(N) – shouldn't be an issue.

In the future I hope we can go further with `ASSection` but this is a good start.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -677,19 +677,13 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   if (indexPath == nil) {
     return nil;
   }
-  
-  // If this is a section index path, we don't currently have a method
-  // to do a mapping.
-  if (indexPath.item == NSNotFound) {
-    return indexPath;
-  } else {
-    NSIndexPath *viewIndexPath = [_dataController.visibleMap convertIndexPath:indexPath fromMap:_dataController.pendingMap];
-    if (viewIndexPath == nil && wait) {
-      [self waitUntilAllUpdatesAreCommitted];
-      return [self convertIndexPathFromCollectionNode:indexPath waitingIfNeeded:NO];
-    }
-    return viewIndexPath;
+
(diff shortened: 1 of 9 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASCollectionView.mm
```diff
@@ -64,8 +64,6 @@
     return __val; \
   }
 
-#define ASIndexPathForSection(section) [NSIndexPath indexPathForItem:0 inSection:section]
```
Adlai-Holler: Removed this macro because using item 0 is only something that's done for flow layout headers and footers. In any other context, "index path for section" would mean a one-node index path.

Source/ASCollectionView.mm
```diff
@@ -1050,7 +1036,7 @@ - (CGSize)collectionView:(UICollectionView *)cv layout:(UICollectionViewLayout *
   ASDisplayNodeAssertMainThread();
   ASElementMap *map = _dataController.visibleMap;
   ASCollectionElement *e = [map supplementaryElementOfKind:UICollectionElementKindSectionHeader
-                                               atIndexPath:ASIndexPathForSection(section)];
+                                               atIndexPath:[NSIndexPath indexPathForItem:0 inSection:section]];
```
appleguy: Can we use an NSNotFound for the item, or even a single-number index path to represent this?

Issue #658, opened 2017-11-02
[RFC][ASDisplayNode] Provide easy access to safe area insets


ypogribnyi:
The safe area is really important for support of iPhone X screen. Having easy access to it would also be handy when designing for the classic iPhones to make sure the content is not obscured by the translucent navigation bars and toolbars.

These insets are currently available via some workarounds, but as a user, I'd like `ASDisplayNode` to have the `safeAreaInsets` property and the `- (void)safeAreaInsetsDidChange` callback. On iOS 11 they can be simply bridged from the underlying view. And it would be great to have a backward compatible calculation algorithm for older iOS versions (they can be derived from view controller's top and bottom layout guides).

I think that `automaticallyRelayoutOnSafeAreaChanges` property would be very handy too.

It would also be nice to have the `layoutMargins` and related properties on `ASDisplayNode`, but I'm not sure how many people really need them.

fruitcoder:
Accessing the underlying views `safeAreaInsets` and being informed when the insets changed seems very reasonable for `ASDisplayNode` subclasses (especially returning `ASLayoutSpec`s depending on those values). It might be hard to sync those properties if you have a layer tree and the views are only created for accessing the `safeAreaInsets`. 
It might also be possible to put them in the `ASTraitCollection` and react to `asyncTraitCollectionDidChange()` although that would be quite different from `UIKit`.

ay8s:
Think this can be closed 👍

PR #685, opened 2017-11-23
[ASDisplayNode] Provide safeAreaInsets and layoutMargins bridge

ypogribnyi:
My attempt to implement #658.

Safe area insets are quite important when designing for iPhone X or when dealing with view controller containers with overlapping bars (e.g. `UINavigationController` with a translucent navigation bar).

This PR includes:

* `safeAreaInsets`, `insetsLayoutMarginsFromSafeArea`, `layoutMargins`, `preservesSuperviewLayoutMargins` properties and `-safeAreaInsetsDidChange`, `-layoutMarginsDidChange` callbacks on `ASDisplayNode` all bridged to the underlying view.

* `automaticallyRelayoutOnSafeAreaChanges` and `automaticallyRelayoutOnLayoutMarginsChanges` flags on `ASDisplayNode` so we can automatically send `-setNeedsLayout` when the insets change.

* A compatibility `safeAreaInsets` calculation algorithm for old iOS versions (for iOS < 11 we have to observe view controller's `topLayoutGuide` and `bottomLayoutGuide` lengths, which is very inconvenient).


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/20167c19d7db5cbe27770fb936e84b79/raw/5a424ba19a143165ceea5231c3d75ad17dc0eabb/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4cec51208081620e051ee3f43215fe69/raw/47aad9a23e4749e5a6fd5cbcc5e37416e7351447/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0a9656a624d7c6dfaa31fb4670fb82ac/raw/cad382c5caaf5d7cf4b238ef94ee6d64cf713d72/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1efac2498d80169d210fabca278184d5/raw/0b64c65227c30c884f8ba922ff1560a40b3ec331/log.txt)

ypogribnyi:
@nguyenhuy, I fixed the tests that were failing because of my code. Now the log shows ticks ✓ for all tests and a strange error 'Early unexpected exit, operation never finished bootstrapping - no restart will be attempted' from xcodebuild. Is it because some tests silently fail, or is it some problem with Xcode?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e315a8633455ce65e229e55a3faee345/raw/3fcfc1fc9528e6c126b3da88a6316d4ede16e234/log.txt)

appleguy:
@ypogribnyi thanks for developing this! I think we could probably merge it soon if we can get it passing CI (apologies for the long delay over the holiday).

Perhaps you could sync to latest and make any final tweaks you had in mind and see if the build server is cooperating now?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/daa05aa71af6ac43a07dcfc552d86d84/raw/ae719fb81d751813d2fb9b9f73508cd030544c99/log.txt)

ypogribnyi:
@appleguy I have finally fixed failing tests. Have nothing else to add here.

Source/ASViewController.mm
```diff
@@ -73,10 +74,14 @@ - (void)_initializeInstance
   if (_node == nil) {
     return;
   }
+
+  _node.viewControllerRoot = YES;
```
nguyenhuy: This is fine for now, especially because it's thread-safe and faster than `-[ASDisplayNode(Convenience) closestViewController]`.

One edge case I can think of is that a root-of-view-controller node is reused and added to a different node hierarchy. It is rare, but we should document is somehow, and maybe assert in debug mode (e.g if `viewControllerRoot` is true, assert that its close view controller indeed points to itself).

Source/ASViewController.mm
```diff
@@ -159,6 +164,13 @@ - (void)viewDidLayoutSubviews
     [_node recursivelyEnsureDisplaySynchronously:YES];
   }
   [super viewDidLayoutSubviews];
+
+  UIEdgeInsets safeArea = UIEdgeInsetsMake(self.topLayoutGuide.length, 0, self.bottomLayoutGuide.length, 0);
+  UIEdgeInsets additionalInsets = self.additionalSafeAreaInsets;
+
+  safeArea = ASConcatInsets(safeArea, additionalInsets);
+
+  _node.fallbackSafeAreaInsets = safeArea;
```
nguyenhuy: This node is guaranteed to be view-backed and loaded `-viewDidLayoutSubviews`. The only reason the fallback insets is needed here is that the runtime is iOS 10 or earlier. Let's take advantage of that and avoid doing unnecessary work if possible.

Source/ASViewController.mm
```diff
@@ -264,6 +276,24 @@ - (ASInterfaceState)interfaceState
...
+  return _fallbackAdditionalSafeAreaInsets;
+}
+
+- (void)setAdditionalSafeAreaInsets:(UIEdgeInsets)additionalSafeAreaInsets
+{
+  if (AS_AVAILABLE_IOS(11.0)) {
+    [super setAdditionalSafeAreaInsets:additionalSafeAreaInsets];
+  }
+
+  _fallbackAdditionalSafeAreaInsets = additionalSafeAreaInsets;
```
nguyenhuy: Should we recalculate the root node's insets here?

Source/Private/ASDisplayNode+FrameworkPrivate.h
```diff
@@ -237,6 +237,25 @@ __unused static NSString * _Nonnull NSStringFromASHierarchyStateChange(ASHierarc
...
+
+/**
+ * @abstract Indicates if this node is a view controller's root node. Defaults to NO.
+ *
+ * @discussion Set to YES in -[ASViewController initWithNode:].
+ *
+ * YES here only means that this node is used as an ASViewController node. It doesn't mean that this node is a root of
+ * ASDisplayNode hierarchy, e.g. when its view controller is parented by another ASViewController.
+ */
+@property (nonatomic, assign, getter=isViewControllerRoot) BOOL viewControllerRoot;
```
nguyenhuy: Good call on making this property private :)

Source/Private/ASDisplayNodeInternal.h
```diff
@@ -326,6 +335,9 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
  */
 - (void)nodeViewDidAddGestureRecognizer;
 
+// Recalculates fallbackSafeAreaInsets for the subnodes
+- (void)_fallbackUpdateSafeAreaOnChildren;
```
nguyenhuy: Nit: May not needed because `_fallbackUpdateSafeAreaOnChildren` is implemented and called in `ASDisplayNode.mm` only.

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -871,6 +871,115 @@ - (void)setSemanticContentAttribute:(UISemanticContentAttribute)semanticContentA
...
+    _fallbackInsetsLayoutMarginsFromSafeArea = insetsLayoutMarginsFromSafeArea;
+
+    if (AS_AVAILABLE_IOS(11.0)) {
+      if (!_flags.layerBacked) {
+        _setToViewOnly(insetsLayoutMarginsFromSafeArea, insetsLayoutMarginsFromSafeArea);
+      }
+    }
+  }
+
+  if (__loaded(self) && (!AS_AT_LEAST_IOS11 || _flags.layerBacked)) {
```
nguyenhuy: Warning: accessing `_flags` without the lock.

Source/Private/ASDisplayNodeInternal.h
ypogribnyi: This is also used by `-[ASDisplayNode safeAreaDidChange]`, which is implemented in `ASDisplayNode+UIViewBridge.mm`.

Source/ASViewController.mm
ypogribnyi: I added an assertion to `-[_ASDisplayView didMoveToSuperview]` to check for that case. Would that be enough?

Source/ASDisplayNode.mm
```diff
@@ -829,6 +837,96 @@ - (void)nodeViewDidAddGestureRecognizer
...
+  if (needsManualUpdate) {
+    [self safeAreaInsetsDidChange];
+  }
+
+  if (updatesLayoutMargins) {
+    [self layoutMarginsDidChange];
+  }
+}
+
+- (void)_fallbackUpdateSafeAreaOnChildren
```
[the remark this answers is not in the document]
ypogribnyi: I added a shortcut that skips a child if it is a view-backed node and we run on iOS 11. And `-[ASDisplayNode setFallbackSafeAreaInsets:]` returns early when the insets don't change.

Although we could try to skip calculations for the children that use fallback safe area insets but don't really need to have them updated (e.g. neither the child's frame, the parent's frame nor the parent's safe area wasn't updated), I doubt that it would save more CPU time than would spend running the checks.

The other thing I'm thinking about is that on iOS 11 we can disable fallback logic completely if we avoid supporting the safe area for the layer-backed nodes. But I'm not sure if we should do that.

ypogribnyi:
@nguyenhuy, thank you for the detailed review!

ay8s:
This is awesome @ypogribnyi. I wonder if similar could be done to expose the readableContentGuide within cell nodes etc? Think it could even just be the layoutFrame from the readable content guide that is accessible so you can then constrain the layout to that within layoutSpecThatFits:

appleguy:
@ypogribnyi do you have any cycles to wrap up remaining requests from Huy? Let us know either way - if you need someone to take this over, I'm not sure who might be able to, but then we'd be progressing towards landing it.

This is a great contribution overall, so thanks for taking the time to work with us!

philtre:
It would be super helpful to have a `ASSafeAreaLayoutSpec` that abstracts this functionality. Looking forward to the merge!

Source/ASDisplayNode.mm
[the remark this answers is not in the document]
nguyenhuy: Thanks for the note. I think this implementation and your trade-offs are reasonable. Let's roll with this.

Source/ASViewController.mm
nguyenhuy: 👍

PR #689, opened 2017-11-26
Fix typo in docs: "constraints" -> "layout spec" in ASTableNode / Table Row Height

richardhenry:
I think this is what is actually intended?


richardhenry:
Oh I see - that makes more sense

PR #424, opened 2017-07-07
Add unit tests for the layout engine

Adlai-Holler:
This adds a testing platform and some unit tests to the layout engine. Right now there are flags to toggle between the current behavior and the desired behavior, and they're set to "current behavior."

Issues reproduced (need to merge with GitHub issues):
- Extraneous `calculatedLayoutDidChange` calls. Not a biggie.
- If a layout pass comes in mid-transition, we will calculateLayout: for some nodes against their current bounds on main, but discard the result. https://github.com/TextureGroup/Texture/pull/424/files#diff-b1441b7e61c34b6f132da6c94bb10c0cR125
- If we do `setNeedsLayout` and get a layout pass while a transition is measuring, the transition's "to-layout" will win, but the setNeedsLayout+layoutPass should actually win since its data is newer. https://github.com/TextureGroup/Texture/pull/424/files#diff-b1441b7e61c34b6f132da6c94bb10c0cR204
- If we setNeedsLayout on a leaf node that is subject to multipass measurement (e.g. flexShrink) we will not update the layout for the leaf or the stack. We will remeasure it only against the fixed (second-pass) constrained size, and inevitably get a result confirming that the current size is correct, but we should have offered the node another unconstrained measurement pass. In my estimation this is the most serious issue. https://github.com/TextureGroup/Texture/pull/424/files#diff-b1441b7e61c34b6f132da6c94bb10c0cR284

I'd like to reproduce the orphan-node issue but I'm not sure what conditions would trigger it @nguyenhuy. We've got enough flexibility in the testing platform – I believe – to test an incredible variety of different scenarios.

PROTIP: You can take any expectation and add `.andDebugBreak()` to get a breakpoint when that call happens. So if you expect an async layout: `OCMExpect([nodeA.mock calculateLayoutThatFits:]).offMainThread()` you can change that to `OCMExpect([nodeA.mock calculateLayoutThatFits:]).offMainThread().andDebugBreak()` and when that calculateLayoutThatFits invocation happens, the debugger will pause for you. ✨Magic✨

PR #695, opened 2017-11-30
[ASDisplayNode+Layout] Ensure a pending layout is applied once

nguyenhuy:
Before:
  - Even if a pending layout was applied before, it'll be unnecessarily applied again in next layout passes and cause `-calculatedLayoutDidChange` being called multiple times.

After:
  - If a pending layout was applied, the calculated layout will not be ignored but reused, if possible, in next layout passes.

Test plan: `testSetNeedsLayoutAndNormalLayoutPass` in #424.

### DIFF
--- a/Source/ASDisplayNode+Layout.mm
+++ b/Source/ASDisplayNode+Layout.mm
@@ -306,7 +306,7 @@ ASLayoutElementStyleExtensibilityForwarding
                                    && _pendingDisplayNodeLayout->version >= _layoutVersion
                                    && _pendingDisplayNodeLayout->version > _calculatedDisplayNodeLayout->version); // _pending is not yet applied
   BOOL calculatedLayoutIsReusable = (_calculatedDisplayNodeLayout->version >= _layoutVersion
-                                     && (_calculatedDisplayNodeLayout->requestedLayoutFromAbove == YES
+                                     && (_calculatedDisplayNodeLayout->requestedLayoutFromAbove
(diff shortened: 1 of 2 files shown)


Source/ASDisplayNode+Layout.mm
```diff
@@ -300,14 +300,16 @@ - (void)_u_measureNodeWithBoundsIfNecessary:(CGRect)bounds
...
-    }
+  // Prefer a newer and not yet applied _pendingDisplayNodeLayout over _calculatedDisplayNodeLayout
+  // If there is no such _pending, check if _calculated is valid to reuse (avoiding recalculation below).
+  BOOL pendingLayoutIsPreferred = (_pendingDisplayNodeLayout != nullptr
+                                   && _pendingDisplayNodeLayout->version >= _layoutVersion
+                                   && _pendingDisplayNodeLayout->version > _calculatedDisplayNodeLayout->version); // _pending is not yet applied
+  BOOL calculatedLayoutIsReusable = (_calculatedDisplayNodeLayout->version >= _layoutVersion
+                                     && (_calculatedDisplayNodeLayout->requestedLayoutFromAbove == YES
+                                         || CGSizeEqualToSize(_calculatedDisplayNodeLayout->layout.size, boundsSizeForLayout)));
+  if (!pendingLayoutIsPreferred && calculatedLayoutIsReusable) {
```
Adlai-Holler: Nit: let's take the opportunity to remove `== YES` and maybe add a comment explaining the `requestedLayoutFromAbove + sizeCheck` logic.

PR #705, opened 2017-12-07
fix bug for ASVideoPlayNode

FengDeng:
fix bug.ASVideoNodePlayerStateReadyToPlay always called before ASVideoNodePlayerStateReadyToPlay when autoplay。cause duration and timelable couldn't update.

### DIFF
--- a/Source/ASVideoPlayerNode.mm
+++ b/Source/ASVideoPlayerNode.mm
@@ -499,7 +499,7 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
     [_delegate videoPlayerNode:self willChangeVideoNodeState:state toVideoNodeState:toState];
   }
 
-  if (toState == ASVideoNodePlayerStateReadyToPlay) {
+  if (toState == ASVideoNodePlayerStateReadyToPlay || toState == ASVideoNodePlayerStatePlaying) {
     _duration = _videoNode.currentItem.duration;
     [self updateDurationTimeLabel];
   }


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

flovouin:
Hi @FengDeng,

I took a brief look at the diff, however it’s been a while since I fiddled with the `ASVideoNode`, and I’ve never even used the `ASVideoPlayerNode`, so take everything I say with a grain of salt.
From what I understand from your comment, `ReadyToPlay` is already set when it should actually be set, therefore the `ASVideoPlayerNode` does not receive the proper call to the delegate method enabling it to update the duration label. I’ve never noticed such behaviour but I don’t exclude the possibility of a bug lying somewhere in the `ASVideoNode`’s state management.
Also, I don’t think we explicitly ask the `AVAsset` to load the `duration` property before playing the media, hence I don’t think there’s a guaranty that `duration` is accessible when the video node changes its state to `ReadyToPlay`. It’s just very likely that the duration can be obtained when the media is ready to play (but again, it’s been a while so I could be wrong).
To sum up, there’s either:
- A bug in the `ASVideoNode` state management logic (and the `ASVideoPlayerNode` cannot reliably use it to update its interface)
- A misuse of said logic in the `ASVideoPlayerNode` (and `duration` should be updated elsewhere, or with a different condition)

In both cases, updating the duration more often can indeed solve the problem, but is probably not the cleanest way to do so. Therefore, I’m not entirely sure this PR is the way to go.
As always, as sample project would be ideal to see what’s actually going on. The PR description is rather short and I’m not entirely sure of what is happening here.

Could you please provide us with a sample project that demonstrates the problem?

Cheers,

Flo

flovouin:
OK so after looking at Apple documentation, `duration` should indeed be available when the `AVPlayerItem` is `ReadyToPlay`. This means that the bug is likely to be in the `ASVideoNode` state management logic. I think that the fix should therefore be in this logic, rather than working around it in the `ASVideoPlayerNode`.
But then again, a sample project would make it easier to understand what's going on. :-)

nguyenhuy:
@flovouin Thanks for the inputs. @FengDeng  Let us know your thoughts!

FengDeng:
@flovouin  You are right.This bug is in ASVideoNode.AVPlayer rate observer is return before AVPlayerItem's Status. And rate is 1.0 calls jump ASVideoNodePlayerStateReadyToPlay.Then in ASVideoPlayerNode always receive  ASVideoNodePlayerStateReadyToPlay.

FengDeng:
@nguyenhuy move to #756

nguyenhuy:
Let's close this one then.

PR #706, opened 2017-12-07
Ensure an ASM enabled node applies its pending layout when enters preload state

### LINKED ISSUE #693: Automatic subnode management (ASM) and node interface states
@maicki  and I are facing a problem in which subnodes of an ASM-enabled node have not been inserted into the supernode by the time the supernode enters the visible state. That is because ASM inserts new nodes when the pending layout is applied. Once the nodes are inserted, they'll also get the interface states of the supernode, all in one go, and start playing catch-up.

One solution is to call `-[ASDisplayNode layoutIfNeeded]` in `-didEnterPreloadState` of `ASCellNode`, if the cell node uses ASM. Since cell nodes are guaranteed to have their calculated layout by the time they enter the preload range, running a layout pass there shouldn't introduce significant overheads. On the other hand, doing so ensures that all subnodes are inserted at the right time and can start preloading their contents much sooner than the current implementation.

@appleguy @maicki @Adlai-Holler @garrettmoon Any thoughts?

nguyenhuy:
This makes sure subnodes are inserted and start preloading right away, instead of waiting until the next layout pass of the supernode. Fixes #693.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -3087,8 +3087,9 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
     // If this node has an up-to-date layout (and subnodes), calling layoutIfNeeded will be fast.
     //
     // If this node doesn't have a calculated or pending layout that fits its current bounds, a measurement pass will occur
-    // (see __layout and _u_measureNodeWithBoundsIfNecessary:). This scenario should be uncommon, and forcing is necessary
-    // to kick off preloading on subnodes since it's already a bit late.
+    // (see __layout and _u_measureNodeWithBoundsIfNecessary:).
+    // This scenario should be uncommon, and running a measurement pass here is a fine trade-off because preloading
+    // any time after this point would be late.
     [self layoutIfNeeded];
   }
 
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -3081,11 +3081,14 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
   ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__);
 
   if (self.automaticallyManagesSubnodes) {
-    // Tell the node to apply its (applicable) pending layout, if any, so that its subnodes are inserted/deleted and enter preload state as well.
+    // Tell the node to apply its applicable pending layout, if any, so that its subnodes are inserted/deleted
+    // and start preloading right away.
+    //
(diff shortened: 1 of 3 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASDisplayNodeImplicitHierarchyTests.m:</p>
<pre><code>//
//  ASDisplayNodeImplicitHierarchyTests.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

Tests/ASDisplayNodeImplicitHierarchyTests.m
```diff
@@ -2,8 +2,6 @@
 //  ASDisplayNodeImplicitHierarchyTests.m
 //  Texture
 //
-//  Created by Levi McCallum on 2/1/16.
```
nguyenhuy: Sorry Levi, Danger forces me to do this :)

Source/ASDisplayNode.mm
```diff
@@ -3079,6 +3079,20 @@ - (void)didEnterPreloadState
...
+    // Tell the node to apply its applicable pending layout, if any, so that its subnodes are inserted/deleted
+    // and start preloading right away.
+    //
+    // If this node has an up-to-date layout (and subnodes), calling layoutIfNeeded will be fast.
+    //
+    // If this node doesn't have a calculated or pending layout that fits its current bounds, a measurement pass will occur
+    // (see __layout and _u_measureNodeWithBoundsIfNecessary:).
+    // This scenario should be uncommon, and running a measurement pass here is a fine trade-off because preloading
+    // any time after this point would be late.
+    [self layoutIfNeeded];
```
appleguy: Does this work in the case of a node that doesn't have self.isNodeLoaded yet?

Ideally, we would be able to flush the layout / ASM process without relying on a CoreAnimation call at all. For example, a codepath that could call __layout without relying on -layoutSubviews.

nguyenhuy: I agree. That's why I'm calling the bridged `layoutIfNeeded` implementation of `ASDisplayNode` which is thread-safe and doesn't allocate the backing store if the node is not yet loaded (code [here](https://github.com/TextureGroup/Texture/blob/master/Source/Private/ASDisplayNode%2BUIViewBridge.mm#L448)). I also have an [assertion](https://github.com/TextureGroup/Texture/pull/706/files#diff-4bc6f1a7ed5eaad134d9d60d0f5f1c10R134) in the test specifically for this behavior.

PR #719, opened 2017-12-18
Reimplement ASRectTable using unordered_map to avoid obscure NSMapTable exception.

appleguy:
The new class is called ASRectMap, which patterns alongside ASIntegerMap in both name and implementation.

Here's the exception, which occurs about 50 times a day in my app, when invoked from the only current caller (ASLayout):

"*** -[NSMapTable initWithKeyPointerFunctions:valuePointerFunctions:capacity:] Requested configuration not supported."

After some pretty detailed investigation, including study of open-source reimplementations
of Foundation, the best lead I've found on the NSMapTable exception is that
some NSPointerFunction types are not fully supported. Strangely, the ones being used
do seem to work fine almost all of the time.

The main concern is the Struct memory type, which is not officially re-declared in
NSMapTable, and as such the documentation claims that there may exist some
combinations of NSPointerFunction that are not supported. One substantiation for this is that crashes are disproportionately more common since 11.1.2, suggesting there might be a framework change to the concrete subclass in how it handles these flags. Unfortunately, I have not been able to reproduce, but I have ruled out many possibilities such as devices low on memory.

Because the exception is occurring frequently enough to be a concern (in the hundreds
to low thousands, though only 50 a day) - I decided to replace NSMapTable entirely
in order to ensure full correctness.

This does lose some features of the original - namely, strong retaining of keys. However, this feature was unused and the class is private. Since the previous implementation is operating in a grey area of NSMapTable's API, switching to unordered_map seemed like the right tradeoff.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRectMap.mm:</p>
<pre><code>//
//  ASRectMap.mm
//  Texture
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b9dee871564f8a063420fe49348b6d50/raw/8854ac70896c01efbf5129d0c3eea62baea9b3a4/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRectMap.mm:</p>
<pre><code>//
//  ASRectMap.mm
//  Texture
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/2f5353494092aed0419ef366cc6898dc/raw/aa5eff33e75df9c23048b7be096ccec6d960ac63/log.txt)

appleguy:
This builds fine on my local machine, and the tests execute. I think the build server has some kind of caching problem where the file rename is not being picked up correctly. Any suggestions?

Source/Private/ASRectMap.mm
```diff
@@ -0,0 +1,83 @@
...
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+
+#import "ASRectMap.h"
+#import "ASObjectDescriptionHelpers.h"
+#import <UIKit/UIGeometry.h>
+#import <unordered_map>
+
+@implementation ASRectMap {
+  std::unordered_map<void *, CGRect> _map;
```
maicki: We may should add a comment that we use a `void *` key here as arc would not be able to determine the ownership qualification.

Source/Private/ASRectMap.mm
```diff
@@ -0,0 +1,83 @@
...
+  return copy;
+}
+
+- (NSMutableArray<NSDictionary *> *)propertiesForDescription
+{
+  NSMutableArray *result = [NSMutableArray array];
+
+  // { ptr1->rect1 ptr2->rect2 ptr3->rect3 }
+  NSMutableString *str = [NSMutableString string];
+  for (auto it = _map.begin(); it != _map.end(); it++) {
```
maicki: You could also use of a bit of a smaller code:
```
for ( auto const& entry : _map ) {
    [str appendFormat:@" %@->%@", entry->first, NSStringFromCGRect(entry->second)];
}
```

or fancy via a lambda:

Source/Private/ASRectMap.mm
Adlai-Holler: Could we not use `__unsafe_unretained id` as the key type here?

Source/Private/ASRectMap.mm
Adlai-Holler: Is this method dangerous, now that the keys are stored unretained? Could we be calling `-description` on a garbage pointer? We could fall back to `%p`.

Source/Private/ASRectMap.mm
```diff
@@ -0,0 +1,83 @@
...
+    [str appendFormat:@" %@->%@", it->first, NSStringFromCGRect(it->second)];
+  }
+  [result addObject:@{ @"ASRectMap": str }];
+
+  return result;
```
Adlai-Holler: The class name will already be in the description by default, so I think what we want here is `@{ (id)kCFNull, str }` which will just dump the list of pointers inside e.g.: `<ASRectMap: 0xFFFFFFFF; "a->b, c->d">`

Source/Private/ASRectMap.mm
maicki: Unfortunately we cannot. The `unordered_map` template cannot figure out the hash for the id key. We would have to write our own key like:

```
struct ASRectMapMapKey {
    __unsafe_unretained id obj;
    
    bool operator==(const ASRectMapMapKey &other) const
    {
        return [obj isEqual:other.obj];
    }
};

namespace std {
    template<> struct hash<ASRectMapMapKey>
    {
        size_t operator()(const ASRectMapMapKey &k) const
        {
            return [k.obj hash];
        }
    };
}
```

And we could use it via:
```
std::unordered_map<ASRectMapMapKey, CGRect> map;
map.insert({{@"key"}, CGRectMake(0, 0, 100, 100)});
NSLog(@"%@", NSStringFromRect(map.at({@"key"})));
```

Source/Private/ASRectMap.mm
nguyenhuy: Agree with @Adlai-Holler. Let's pay extra attention here.

Source/Private/ASRectMap.mm
```diff
@@ -0,0 +1,83 @@
...
+//  LICENSE file in the /ASDK-Licenses directory of this source tree. An additional
+//  grant of patent rights can be found in the PATENTS file in the same directory.
+//
+//  Modifications to this file made after 4/13/2017 are: Copyright (c) 2017-present,
+//  Pinterest, Inc.  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
```
nguyenhuy: Please update this file license per Danger's [comment](https://github.com/TextureGroup/Texture/pull/719#issuecomment-352317509).

```
//
//  ASRectMap.mm
//  Texture
//
//
//
```

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRectMap.mm:</p>
<pre><code>//
//  ASRectMap.mm
//  Texture
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1b4c15785c66c09cbb0cef00abdbb3c4/raw/7f14e6d61b6511cad8ece8189d6dfd58d0009b34/log.txt)

appleguy:
Thanks for all the feedback, will do some updates! Any ideas on why the build system is failing?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/595daac22557df06c6c19aa239cf5415/raw/aa220e6c5325695176dfc769bb63ea7da8e819fd/log.txt)

appleguy:
OK, ready for another review - if there are any ways to improve the safety of the accesses, let me know (I didn't quite understand if there is).

I considered using __unsafe_unretained id as the argument type for the methods, but this wouldn't affect any behavior, and would just be a form of API documentation — I can make that change if it seems better.

Main thing for now is getting the build to pass, which since I don't have admin access to the system, I need help to do.

nguyenhuy:
OK, tests are all green now. Let's have an approval from @Adlai-Holler before merging.

appleguy:
Thanks everyone!

PR #731, opened 2018-01-01
Fixed: completeBatchFetching is called on a background thread

aaronr93:
When there are no additions to the collection, `numberOfAdditionsCompletion` (and therefore `completeBatchFetching`) is called on a background thread instead of the main thread.

This issue has not come up in this example because the user would have to scroll literally 1000 pages in order to experience it. This is important to fix to maintain consistency in the examples.

### DIFF
--- a/examples_extra/ASDKgram-Swift/ASDKgram-Swift/PhotoFeedModel.swift
+++ b/examples_extra/ASDKgram-Swift/ASDKgram-Swift/PhotoFeedModel.swift
@@ -67,9 +67,9 @@ final class PhotoFeedModel {
 	private func fetchNextPageOfPopularPhotos(replaceData: Bool, numberOfAdditionsCompletion: @escaping (Int, NetworkingErrors?) -> ()) {
 
 		if currentPage == totalPages, currentPage != 0 {
-            DispatchQueue.main.async {
-                return numberOfAdditionsCompletion(0, .customError("No pages left to parse"))
-            }
+			DispatchQueue.main.async {
+				return numberOfAdditionsCompletion(0, .customError("No pages left to parse"))
+			}
 		}
 
 		var newPhotos: [PhotoModel] = []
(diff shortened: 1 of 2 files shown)


examples_extra/ASDKgram-Swift/ASDKgram-Swift/PhotoFeedModel.swift
```diff
@@ -67,7 +67,9 @@ final class PhotoFeedModel {
 	private func fetchNextPageOfPopularPhotos(replaceData: Bool, numberOfAdditionsCompletion: @escaping (Int, NetworkingErrors?) -> ()) {
 
 		if currentPage == totalPages, currentPage != 0 {
-			return numberOfAdditionsCompletion(0, .customError("No pages left to parse"))
+			DispatchQueue.main.async {
+				return numberOfAdditionsCompletion(0, .customError("No pages left to parse"))
```
maicki: You can remove the "return" here.

examples_extra/ASDKgram-Swift/ASDKgram-Swift/PhotoFeedModel.swift
```diff
@@ -67,7 +67,9 @@ final class PhotoFeedModel {
...
+			}
```
maicki: You need to add a "return" here otherwise it would execute this path although it should return like it did before.

appleguy:
Does this bug also exist in the Objective-C version?

Also, are all the ASBatch calls on ASCollectionNode behaving correctly, and only communicating with the delegate on main?

aaronr93:
@maicki I made those changes. Sorry about the delay; I need to tweak my Github notifications.

I found the same issue in the Objective-C version (95% sure...I'm not very experienced with Objective-C) and fixed it.

@appleguy I'm not sure if all the examples are correctly communicating with the delegate, if that's what you mean.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for PhotoFeedModel.m:</p>
<pre><code>//
//  PhotoFeedModel.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

PR #735, opened 2018-01-05
Enable collection node interactive moves

Adlai-Holler:
This patch enables interactive movement for collection node items and updates the `ASCollectionView` example to showcase it.

Caveats:
- Item movement is disabled when using async collection layout. See the comment at the top of `- (BOOL)collectionView:(UICollectionView *)collectionView canMoveItemAtIndexPath:(NSIndexPath *)indexPath` for details.
- When the item is dropped, a new node will be created for it, so it may flash while it's rendered. This is because our change set currently represents moves as delete/insert so the old node is discarded.

To use:
- Have your data source implement `collectionNode:moveItemAtIndexPath:toIndexPath:`.
- Optionally, implement `collectionNode:canMoveItemWithNode:`. As with UIKit, this defaults to `YES` as long as you implement the move method.
- Use the normal `UICollectionView` interactive movement methods, such as `beginInteractiveMovementForItemAtIndexPath:`. We will need to bridge or intercept these at this time.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -1517,6 +1517,10 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   // In iOS >= 11, this is made much easier by the UIDataSourceTranslating API. In previous versions of iOS our best bet
   // would be to capture the invalidation contexts that are sent during interactive moves and make our own data source translator.
   if ([self.collectionViewLayout isKindOfClass:[ASCollectionLayout class]]) {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+      as_log_debug(ASCollectionLog(), "Collection node item interactive movement is not supported when using a layout delegate. This message will only be logged once. Node: %@", ASObjectDescriptionMakeTiny(self));
+    });
     return NO;
   }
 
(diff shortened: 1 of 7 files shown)


Source/ASCollectionView.mm
```diff
@@ -1492,6 +1500,62 @@ - (void)collectionView:(UICollectionView *)collectionView performAction:(nonnull
...
+  // Currently we do not support interactive moves when using async layout. The reason is, we do not have a mechanism
+  // to propagate the "presentation data" element map (containing the speculative in-progress moves) to the layout delegate,
+  // and this can cause exceptions to be thrown from UICV. For example, if you drag an item out of a section,
+  // the element map will still contain N items in that section, even though there's only N-1 shown, and UICV will
+  // throw an exception that you specified an element that doesn't exist.
+  //
+  // In iOS >= 11, this is made much easier by the UIDataSourceTranslating API. In previous versions of iOS our best bet
+  // would be to capture the invalidation contexts that are sent during interactive moves and make our own data source translator.
+  if ([self.collectionViewLayout isKindOfClass:[ASCollectionLayout class]]) {
+    return NO;
```
maicki: Should we print out some log here to let user know about it?

nguyenhuy:
#141.

PR #737, opened 2018-01-09
2.4

sudhanshutil:


Adlai-Holler:
@sudhanshutil What's the motivation for this diff? The changes being pulled are already reflected in the current master, for instance `canUpdateToViewModel: -> canUpdateToNodeModel:` is already in the current master.

garrettmoon:
Going to close for now, lemme know if we should re-open!

PR #748, opened 2018-01-16
Faster collection operations

Adlai-Holler:
The idea here is, standardize on `ASArrayByFlatMapping` and make that macro faster and return an immutable array.

Immutable arrays mean (1) free copying, and (2) free zero-arrays and optimized singleObjectArrays, which means less retaining/releasing. Yay!

I haven't profiled this yet. Just want to put it out there.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/xcshareddata/xcschemes/AsyncDisplayKit.xcscheme
+++ b/AsyncDisplayKit.xcodeproj/xcshareddata/xcschemes/AsyncDisplayKit.xcscheme
@@ -60,6 +60,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      disableMainThreadChecker = "YES"
       launchStyle = "0"
       useCustomWorkingDirectory = "NO"
       ignoresPersistentStateOnLaunch = "NO"
--- a/Source/Base/ASBaseDefines.h
+++ b/Source/Base/ASBaseDefines.h
@@ -132,22 +132,6 @@
 #define __has_attribute(x) 0 // Compatibility with non-clang compilers.
 #endif
 
-#ifndef NS_CONSUMED
-#if __has_feature(attribute_ns_consumed)
-#define NS_CONSUMED __attribute__((ns_consumed))
-#else
-#define NS_CONSUMED
-#endif
-#endif
-
-#ifndef NS_RETURNS_RETAINED
-#if __has_feature(attribute_ns_returns_retained)
-#define NS_RETURNS_RETAINED __attribute__((ns_returns_retained))
-#else
-#define NS_RETURNS_RETAINED
-#endif
-#endif
-
 #ifndef CF_RETURNS_RETAINED
 #if __has_feature(attribute_cf_returns_retained)
 #define CF_RETURNS_RETAINED __attribute__((cf_returns_retained))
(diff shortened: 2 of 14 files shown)


Source/Layout/ASLayout.mm
```diff
@@ -265,7 +261,8 @@ - (ASLayout *)filteredNodeLayoutTree
     }
   }
   
-  ASLayout *layout = [ASLayout layoutWithLayoutElement:_layoutElement size:_size sublayouts:flattenedSublayouts];
+  NSArray *sublayoutsArray = [NSArray arrayWithObjects:flattenedSublayouts.data() count:flattenedSublayouts.size()];
+  ASLayout *layout = [ASLayout layoutWithLayoutElement:_layoutElement size:_size sublayouts:sublayoutsArray];
```
Adlai-Holler: IIRC this method is one of our hottest methods, and we copy the resulting array so I'm hopeful this will be a nice bump.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f85214e18ced692482b54e254a139ae7/raw/2f69f6f99c2bfe9cdfcb3d23b970a3d416d3c72f/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/33ea9d8b4215bb75ef62510d5d91e6d2/raw/20686be94b1f87c65f558dc20eac293b7c2654bf/log.txt)

AsyncDisplayKit.xcodeproj/xcshareddata/xcschemes/AsyncDisplayKit.xcscheme
```diff
@@ -60,6 +60,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      disableMainThreadChecker = "YES"
```
Adlai-Holler: This disables the main thread checker for unit tests and there are cases where we expect to call UIKit methods off-main during unit tests.

Source/Details/ASElementMap.h
```diff
@@ -31,6 +31,11 @@ NS_ASSUME_NONNULL_BEGIN
 AS_SUBCLASSING_RESTRICTED
 @interface ASElementMap : NSObject <NSCopying, NSFastEnumeration>
 
+/**
+ * The total number of elements in this map.
+ */
+@property (readonly) NSUInteger count;
```
Adlai-Holler: This is required by the new implementation of `ASArrayByFlatMapping`, in addition to NSFastEnumeration conformance.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f499c7282fab1d96b81f53f18a390019/raw/b66282956bab857d908b95629822b79d9c7b6dc1/log.txt)

Source/Base/ASBaseDefines.h
```diff
@@ -273,12 +269,37 @@
...
+  } else if (_i == 1) { \
+    /** NSSingleObjectArray is fast. Create one and release. */ \
+    CFTypeRef val = _cArray[0]; \
+    result = [NSArray arrayWithObject:(__bridge id)val]; \
+    CFBridgingRelease(val); \
+  } else { \
+    CFArrayCallBacks cb = kCFTypeArrayCallBacks; \
+    cb.retain = NULL; \
+    result = (__bridge NSArray *)CFArrayCreate(kCFAllocatorDefault, _cArray, _i, &cb); \
+  } \
```
Adlai-Holler: I'm going to factor this out a bit and clean it up, but here's the approach:
- We need to safely retain the result of the work.
- We want to transfer that retain into the array, instead of having it retain and then we release.
- The way is, use `__bridge_retained` to tell ARC "make sure this is a +1, and I'll take responsibility for it."
- Then when you create the array, use `CFArray` and provide `NULL` for the retain callback. Boom, CFArray won't retain, but it will release when the array goes away.
- Fast paths for 0- or 1- element collections e.g. `__NSArray0` and `__NSSingleObjectArray`.
- Not possible with `id []` because ARC will always release the elements when the symbol goes out of scope – there is no `CF_CONSUMED` for c-arrays.

Adlai-Holler:
Note that, as much as I love this diff, copy-on-write support was added to CoreFoundation for arrays, sets, and dictionaries. So copying a mutable array gives you an `__NSFrozenArrayM`, and copying that gives you the same instance back. The mechanics of the copy-on-write are unclear because it's written in Objective-C (and thus not open source) but it's definitely (attempted to be) running.

AsyncDisplayKit.xcodeproj/xcshareddata/xcschemes/AsyncDisplayKit.xcscheme
appleguy: Interesting, I'm a bit curious which things we perform off-main intentionally in tests? I wonder if there is a way to do that in-situ, since the general existence of this checker does have value (e.g. if you were to accidentally call UIKit on main in a regression sense, then a unit test could catch it)

appleguy:
This is extremely cool.

It could be enough to avoid 1 frame drop in several common scenarios, either faster devices that are near the threshold or dialing back the severity of an inevitable drop on a slower device.

There is an awful lot of runtime traffic in common codepaths that run on main — just thousands of very short method calls, which do contend with background threads, especially when any part of the system is using dynamically-implemented methods. This should reduce the surface area for contention (and thus risk for the longer kinds of sporadic stalls / priority inversion and rapid priority donation thrash), and also reduce total CPU time even when un-contended.

I'd like to study the code and profile this in more detail, but if you develop confidence in it, then don't wait on me to land.

Source/Base/ASBaseDefines.h
```diff
@@ -245,26 +229,38 @@
...
+    cb.retain = NULL; \
+    result = (__bridge NSSet *)CFSetCreate(kCFAllocatorDefault, _cArray, _i, &cb); \
+  } \
+  result; \
 })
 
 /**
  * Create a new ObjectPointerPersonality NSHashTable by mapping `collection` over `work`, ignoring nil.
  */
 #define ASPointerTableByFlatMapping(collection, decl, work) ({ \
```
appleguy: Is this one not yet used? Just wanted to say, I think this #define-based approach is very powerful as an optimization, in its ability to avoid wrapping and retains that most other mechanisms lead to. We should use it judiciously, but I bet there are a few more cases where it is justified and impactful.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASCollectionGalleryLayoutDelegate.mm:</p>
<pre><code>//
//  ASCollectionGalleryLayoutDelegate.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

PR #749, opened 2018-01-16
[ASDisplayNode] Always return the thread-safe cornerRadius property, even in slow CALayer rounding mode

nguyenhuy:
- If the node's property was updated on a background thread and main thread has not executed the block that updates `cornerRadius` of the node's layer, the layer's property is out-of-date and shouldn't be used.
- After this change, ASDisplayNode's cornerRadius is the only source of truth and users must always use it instead of CALayer's.

Relevant PR: #465.
Recorded bug [here](https://cl.ly/osDG) - Pay attention to the "Save" button on the top right corner.

### DIFF
--- a/Source/Private/ASDisplayNode+UIViewBridge.mm
+++ b/Source/Private/ASDisplayNode+UIViewBridge.mm
@@ -186,16 +186,11 @@ if (shouldApply) { _layer.layerProperty = (layerValueExpr); } else { ASDisplayNo
 - (CGFloat)cornerRadius
 {
   ASDN::MutexLocker l(__instanceLock__);
-  if (_cornerRoundingType == ASCornerRoundingTypeDefaultSlowCALayer) {
-    return self.layerCornerRadius;
-  } else {
-    return _cornerRadius;
-  }
+  return _cornerRadius;
 }
 
 - (void)setCornerRadius:(CGFloat)newCornerRadius
 {
-  ASDN::MutexLocker l(__instanceLock__);
   [self updateCornerRoundingWithType:_cornerRoundingType cornerRadius:newCornerRadius];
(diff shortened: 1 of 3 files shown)


Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -186,16 +186,11 @@ - (void)setAlpha:(CGFloat)newAlpha
...
-    return self.layerCornerRadius;
-  } else {
-    return _cornerRadius;
-  }
+  return _cornerRadius;
 }
 
 - (void)setCornerRadius:(CGFloat)newCornerRadius
 {
-  ASDN::MutexLocker l(__instanceLock__);
```
nguyenhuy: `-updateCornerRoundingWithType:cornerRadius:` acquires the lock itself ([here](https://github.com/TextureGroup/Texture/pull/465/files#diff-fb6dadff65468c53595c164828199c83R1549)).

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -207,7 +202,6 @@ - (ASCornerRoundingType)cornerRoundingType
 
 - (void)setCornerRoundingType:(ASCornerRoundingType)newRoundingType
 {
-  ASDN::MutexLocker l(__instanceLock__);
   [self updateCornerRoundingWithType:newRoundingType cornerRadius:_cornerRadius];
```
Adlai-Holler: Isn't this public? Think we should read `self.cornerRadius`?

nguyenhuy: yeap, nice catch!

PR #751, opened 2018-01-16
[ASDisplayNode] Don't force a layout pass on a visible node that enters preload state

nguyenhuy:
After #706, a layout pass is forced on an ASM-enabled node that enters preload state to make sure that its subnodes can start preloading as well. However, when the node is visible, a (coalesced, thus more efficient) layout pass will be triggered by CA soon anyways, so rely on it instead.


nguyenhuy:
Thanks for the review, @appleguy.

PR #757, opened 2018-01-19
[ASPrimitiveTraitCollection] Always treat preferredContentSize as a potential nil #trivial

ypogribnyi:
As `ASPrimitiveTraitCollection` is a struct, the application can always initialize it in its own way, without using Texture-provided initialization functions. This may lead to `preferredContentSize` field being `nil`, which, in turn, may lead to frustrating crashes when the struct is given to Texture.

While simply advising the user to always use the provided initialization functions should solve the problem, it's better to always treat this structure as possibly improperly initialized and guard against `nil` value in this field.

Additionally this PR explicitly marks all `UIContentSizeCategory` usage in `ASTraitCollection` class as non-null.

### DIFF
--- a/Source/Details/ASTraitCollection.h
+++ b/Source/Details/ASTraitCollection.h
@@ -58,6 +58,12 @@ extern ASPrimitiveContentSizeCategory ASPrimitiveContentSizeCategoryMake(UIConte
 
 #pragma mark - ASPrimitiveTraitCollection
 
+/**
+ * @abstract This is a struct equivalent of ASTraitCollection.
+ *
+ * @discussion It may be not safe to use this structure uninitialized. Please initialize it with ASPrimitiveTraitCollectionMakeDefault()
+ * or ASPrimitiveTraitCollectionFromUITraitCollection(UITraitCollection*).
+ */
 typedef struct ASPrimitiveTraitCollection {
   UIUserInterfaceSizeClass horizontalSizeClass;
   UIUserInterfaceSizeClass verticalSizeClass;
@@ -175,7 +181,7 @@ AS_SUBCLASSING_RESTRICTED
 @property (nonatomic, assign, readonly) UIUserInterfaceStyle userInterfaceStyle;
 #endif
 
-@property (nonatomic, assign, readonly) UIContentSizeCategory preferredContentSizeCategory;
+@property (nonatomic, assign, readonly) UIContentSizeCategory _Nonnull preferredContentSizeCategory;
 
 @property (nonatomic, assign, readonly) CGSize containerSize;
 
@@ -186,7 +192,7 @@ AS_SUBCLASSING_RESTRICTED
 
 + (ASTraitCollection *)traitCollectionWithUITraitCollection:(UITraitCollection *)traitCollection
                                               containerSize:(CGSize)windowSize
-                                fallbackContentSizeCategory:(UIContentSizeCategory)fallbackContentSizeCategory;
+                                fallbackContentSizeCategory:(UIContentSizeCategory _Nonnull)fallbackContentSizeCategory;
 
 #if TARGET_OS_TV
 + (ASTraitCollection *)traitCollectionWithHorizontalSizeClass:(UIUserInterfaceSizeClass)horizontalSizeClass
@@ -197,7 +203,7 @@ AS_SUBCLASSING_RESTRICTED
                                          forceTouchCapability:(UIForceTouchCapability)forceTouchCapability
                                               layoutDirection:(UITraitEnvironmentLayoutDirection)layoutDirection
                                            userInterfaceStyle:(UIUserInterfaceStyle)userInterfaceStyle
-                                 preferredContentSizeCategory:(UIContentSizeCategory)preferredContentSizeCategory
+                                 preferredContentSizeCategory:(UIContentSizeCategory _Nonnull)preferredContentSizeCategory
                                                 containerSize:(CGSize)windowSize;
 #else
 + (ASTraitCollection *)traitCollectionWithHorizontalSizeClass:(UIUserInterfaceSizeClass)horizontalSizeClass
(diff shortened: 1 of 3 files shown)


ay8s:
It seems better now when changing device orientation @ypogribnyi. Is there documentation around initializing ASPrimitiveTraitCollection? Might be missing something. 

We were seeing this on iOS 10 & 11.

ypogribnyi:
The preferred way to create `ASPrimitiveTraitCollection` is to use `ASPrimitiveTraitCollectionMakeDefault()` or `ASPrimitiveTraitCollectionFromUITraitCollection(UITraitCollection*)`. There wasn't any documentation about this, but I added a comment to `ASPrimitiveTraitCollection` definition in this PR (not sure if that's enough).

By the way, why do we expose two trait collections interfaces (`ASPrimitiveTraitCollection` and `ASTraitCollection`) to the user? Might it be better to hide one of them?

nguyenhuy:
> By the way, why do we expose two trait collections interfaces (ASPrimitiveTraitCollection and ASTraitCollection) to the user? Might it be better to hide one of them?

Good questions!  Based on https://github.com/facebookarchive/AsyncDisplayKit/pull/1592, I think the original reason is to have ASTraitCollection as a friendly object-based API and ASPrimitiveTraitCollection as the internal actual struct that is used throughout the framework. We should at least document this reasoning, and/or move ASPrimitiveTraitCollection to a framework-private interface.

Source/Details/ASTraitCollection.h
```diff
@@ -175,7 +181,7 @@ AS_SUBCLASSING_RESTRICTED
 @property (nonatomic, assign, readonly) UIUserInterfaceStyle userInterfaceStyle;
 #endif
 
-@property (nonatomic, assign, readonly) UIContentSizeCategory preferredContentSizeCategory;
+@property (nonatomic, assign, readonly) UIContentSizeCategory _Nonnull preferredContentSizeCategory;
```
nguyenhuy: I don't think we need `_Nonnull` annotations since we already use `NS_ASSUME_NONNULL_BEGIN` at the beginning of this file ([here](https://github.com/TextureGroup/Texture/pull/757/files#diff-d9819ce020f3052ee72d0d87446ba684R27)). It's always good to be defensive and guard against unexpected values in the implementation though.

Source/Details/ASTraitCollection.m
```diff
@@ -32,6 +32,14 @@ ASDISPLAYNODE_INLINE UIContentSizeCategory AS_UIContentSizeCategoryUnspecified()
   }
 }
 
+ASDISPLAYNODE_INLINE ASPrimitiveContentSizeCategory safePrimitiveContentSizeCategory(ASPrimitiveContentSizeCategory sizeCategory) {
+  if (sizeCategory) {
+    return sizeCategory;
+  } else {
+    return AS_UIContentSizeCategoryUnspecified();
+  }
+}
```
nguyenhuy: Nit, feel free to ignore this: It's a bit shorter to use ternary operator here, i.e:

 `return sizeCategory ? sizeCategory : AS_UIContentSizeCategoryUnspecified();`

Source/Details/ASTraitCollection.m
```diff
@@ -92,7 +100,7 @@ ASPrimitiveTraitCollection ASPrimitiveTraitCollectionMakeDefault() {
     .displayGamut = UIDisplayGamutUnspecified,
     .userInterfaceIdiom = UIUserInterfaceIdiomUnspecified,
     .layoutDirection = UITraitEnvironmentLayoutDirectionUnspecified,
-    .preferredContentSizeCategory = ASPrimitiveContentSizeCategoryMake(AS_UIContentSizeCategoryUnspecified()),
+    .preferredContentSizeCategory = AS_UIContentSizeCategoryUnspecified(),
```
nguyenhuy: 👍

ay8s:
Thanks @ypogribnyi. We don't actually initialize either ASPrimitiveCollectionTraits or ASTraitCollection ourselves.

I wonder if this might help with digging into the cause http://crashes.to/s/058c9ea3e54

ypogribnyi:
@ay8s, thank you for your input! I finally managed to reproduce the crash. Looks like it occurs in `ASViewController`s that are used without a node (i.e. not using `-[ASViewController initWithNode:]`). Could you please confirm if that is your case?

I'm reproducing it on `master` in `ASDKgram`s `UIKit` tab. It crashes after device rotation when trying to log the trait collection fetched with
```
ASPrimitiveTraitCollection traitCollection = _node.primitiveTraitCollection;
```
while having `nil` in `_node`.

Although having a `nil` node in a view controller is perfectly valid and fully documented behaviour, I  didn't expect that at all.

Anyways, this PR should fix the problem.

ay8s:
Yup that'd be it! Thanks @ypogribnyi!

Source/Details/ASTraitCollection.m
```diff
@@ -32,6 +32,14 @@ ASDISPLAYNODE_INLINE UIContentSizeCategory AS_UIContentSizeCategoryUnspecified()
   }
 }
 
+ASDISPLAYNODE_INLINE ASPrimitiveContentSizeCategory safePrimitiveContentSizeCategory(ASPrimitiveContentSizeCategory sizeCategory) {
```
maicki: I think we still should prefix this functions with AS

Source/Details/ASTraitCollection.m
maicki: Why do we remove the `ASPrimitiveContentSizeCategoryMake` in here. If you read the comment above `ASPrimitiveContentSizeCategory` we have to use `ASPrimitiveContentSizeCategoryMake`

Source/Details/ASTraitCollection.m
```diff
@@ -32,6 +32,14 @@ ASDISPLAYNODE_INLINE UIContentSizeCategory AS_UIContentSizeCategoryUnspecified()
...
+  return sizeCategory ? sizeCategory : AS_UIContentSizeCategoryUnspecified();
+}
```
maicki: I think in both cases we have to use `ASPrimitiveContentSizeCategoryMake(AS_UIContentSizeCategoryUnspecified())`

PR #771, opened 2018-01-28
Improve no-copy rendering experiment, remove +load method

Adlai-Holler:
This PR achieves the following:

- *Critical* Fixes an issue with no-copy rendering where images with bytes-per-row not aligned to 32 have to be copied by CoreAnimation.
- Removes our `+load` method by changing `ASScreenScale` to use `UIGraphics` which is thread-safe.
- Prevents leaks if users fail to close their ASGraphics contexts.
- ASGraphics creates images in the device-specific (wide color) space, same as plain UIGraphics. This doesn't seem to matter as far as I can tell – CA will not copy the image in either case, but still best to copy UIGraphics.
- Always read fields from "reference" UIGraphics contexts so that if they change we are covered.

### DIFF
--- a/Source/Private/ASInternalHelpers.m
+++ b/Source/Private/ASInternalHelpers.m
@@ -143,8 +143,9 @@ CGFloat ASScreenScale()
   static CGFloat __scale = 0.0;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
-    ASDisplayNodeCAssertMainThread();
-    __scale = [[UIScreen mainScreen] scale];
+    UIGraphicsBeginImageContextWithOptions(CGSizeMake(1, 1), YES, 0);
+    __scale = CGContextGetCTM(UIGraphicsGetCurrentContext()).a;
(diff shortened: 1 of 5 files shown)


Source/Private/ASInternalHelpers.m
```diff
@@ -143,8 +143,9 @@ CGFloat ASScreenScale()
   static CGFloat __scale = 0.0;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
-    ASDisplayNodeCAssertMainThread();
-    __scale = [[UIScreen mainScreen] scale];
+    UIGraphicsBeginImageContextWithOptions(CGSizeMake(1, 1), YES, 0);
```
appleguy: @Adlai-Holler as an extra measure of safety, could you add a unit test that verifies the equality of [[UIScreen mainScreen] scale] with ASScreenScale()?

Adlai-Holler: Good call, done.

PR #773, opened 2018-01-29
Expose asyncdisplaykit_node in _ASDisplayView same as in _ASDisplayLayer #trivial

maicki:
For an upcoming diff I need to have access to the node of the `_ASDisplayView`. To be aligned with `_ASDisplayLayer` we should expose it the same way. We exposed it within https://github.com/facebookarchive/AsyncDisplayKit/pull/2346/files for `_ASDisplayLayer`

### DIFF
--- a/Source/Details/_ASDisplayView.h
+++ b/Source/Details/_ASDisplayView.h
@@ -25,8 +25,8 @@
 @interface _ASDisplayView : UIView
 
 /**
- @discussion This property overrides the CALayer category method which implements this via associated objects.
- This should result in much better performance for _ASDisplayLayers.
+ @discussion This property overrides the UIView category method which implements this via associated objects.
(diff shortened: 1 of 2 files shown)


Source/Details/_ASDisplayView.h
```diff
@@ -20,8 +20,16 @@
...
 
+@class ASDisplayNode;
+
 @interface _ASDisplayView : UIView
 
+/**
+ @discussion This property overrides the CALayer category method which implements this via associated objects.
+ This should result in much better performance for _ASDisplayLayers.
+ */
+@property (nonatomic, weak) ASDisplayNode *asyncdisplaykit_node;
```
maicki: We may should add proper nullable annotations for this and in `_ASDisplayLayer` too as both could be nil.

PR #776, opened 2018-01-30
Fix synchronous state of node if +viewClass or +layerClass is overwritten #trivial

maicki:


Source/ASDisplayNode.mm
```diff
@@ -258,7 +258,10 @@ - (void)_initializeInstance
 #endif
   
   _viewClass = [self.class viewClass];
+  setFlag(Synchronous, ![_viewClass isSubclassOfClass:[_ASDisplayView class]]);
+
   _layerClass = [self.class layerClass];
+
```
nguyenhuy: I think we should do the same for `+layerClass`, in case this node is layer-backed.

PR #777, opened 2018-01-30
Fix capturing self in the block while loading image in ASNetworkImageNode

morozkin:


Source/ASNetworkImageNode.mm
```diff
@@ -719,11 +719,6 @@ - (void)_lazilyLoadImageIfNecessary
           strongSelf->_cacheUUID = nil;
           
           ASPerformBlockOnMainThread(^{
-            __typeof__(self) strongSelf = weakSelf;
-            if (strongSelf == nil) {
-              return;
-            }
-
```
Adlai-Holler: I believe this code should stay – we want to "re-strongify" when we reach the main thread.

morozkin: I can't understand you intention for doing such thing because `strongSelf` is already captured and can't pass away

Adlai-Holler: `strongSelf` will be released at the end of the ASPerformBlockOnBackgroundThread block, which may happen before the main thread services our ASPerformBlockOnMainThread call.

Adlai-Holler: So its imporrant that the main thread block not capture strongSelf, and instead capture weakSelf and then attempt to retain it

morozkin: Thanks for explanation. I’ve understood your intention. I supposed that it isn’t critical to capture strongSelf in this case because there are only delegate calls inside the block. I will revert my change

Adlai-Holler: Cool, thanks.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Adlai-Holler:
@morozkin If you'll add an entry to the changelog, we'll be ready to merge this.

PR #775, opened 2018-01-30
Add support for providing additional info to network image node delegate

Adlai-Holler:
`PINRemoteImage` recently gained support for retrieving the raw network response from image downloads.

In order to pipe this data to the application layer, network image nodes need to retrieve it and pass it through.

This diff is **minor breaking API** (see changelog) but since the `didLoadImage:info:` method was only added recently https://github.com/TextureGroup/Texture/pull/639 and since the migration pathway is trivial, I think we can release it in the next minor version.

Introduces class `ASNetworkImageLoadInfo` which includes source type, URL, download identifier, and a new `id userInfo` field.

Basic image downloader sends `nil`.
PIN image downloader sends `PINRemoteImageManagerResult`.

The URL is useful as well since, given that we unlock before calling out to the delegate, the image node URL could change before the user reads it in `didLoadImage:`.

PR #778, opened 2018-01-30
Improve ASNetworkImageNode delegate callout behavior

Adlai-Holler:
- Don't dispatch to main thread if there's no callout to run.
- Don't lock and then immediately unlock when we reach the main thread.
- Don't call out with `nil` image in the case of animated images.
- Call out using the exact image we loaded, instead of re-reading `self.image` at callout time, when it may have changed. If we loaded another image in the mean time, we will have scheduled that callout and it's important to preserve the serial nature here.

As I commented in #775, the docs say these callouts are off-main, but I decided to keep the on-main behavior for this diff.


Adlai-Holler:
Use side-by-side for this one imo 😞

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #788, opened 2018-02-05
New runloop queue to coalesce Interface state update calls.

wsdwsd0829:
Coalesce Interface state update calls. 

This PR introduce a new run loop queue that will execute before CATransaction commit. 
Each node will only be called once per transaction commit to reflect interface change.
This will help reducing calls to update to interfaceState like didEnterPreload, didEnterVisible during controller transitions like push and help apps that use these api to accurately log view visible events. 

For example: 
Before changes, ASCollectionView example will call didEnterVisibleState 47 times when “Push Another Copy” clicked, while only 27 times with new queue.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/331bbe92733aa7f1c463e8b8389b2826/raw/27480f65fb89832f842c45868f30478b7081bef4/log.txt)

appleguy:
Thanks for posting this, Max! @maicki, @Adlai-Holler, @nguyenhuy — this is a significant enough change that it would be great to have multiple eyes on it.

We are running with this internally for the past few weeks. Still, it is possible it can break some unit tests (e.g. at the app layer) where they are not very realistic in simulating the UIKit environment, where a CATransaction commit does not occur. This can be most easily fixed by disabling interfaceState coalescing in the test setup method, or for more accurate tests, by ensuring a key & visible window is used and the runloop is allowed to turn before test expectations are verified.

Source/ASDisplayNode.mm
```diff
@@ -2741,7 +2741,9 @@ - (void)setHierarchyState:(ASHierarchyState)newState
   // Entered or exited range managed state.
   if ((newState & ASHierarchyStateRangeManaged) != (oldState & ASHierarchyStateRangeManaged)) {
     if (newState & ASHierarchyStateRangeManaged) {
-      [self enterInterfaceState:self.supernode.interfaceState];
+      if (self.supernode) {
+        [self enterInterfaceState:self.supernode->_pendingInterfaceState];
```
appleguy: In order to do this, we would need to grab the __instanceLock__ of the supernode, which shouldn't be done manually.

To correctly fix this case, we'll need to add an accessor method for - (ASInterfaceState)pendingInterfaceState which uses a MutexLocker and returns the _pendingInterfaceState.

This method can be declared in ASDisplayNode+FrameworkPrivate.h, so it can be used by various internal subclasses without being made public right away (we may eventually want to make this public, but not until there is a use case — I'm optimistic that there will never be a need to do this).

Source/ASDisplayNode.mm
```diff
@@ -2784,30 +2786,34 @@ - (void)didExitHierarchy
...
+        if (![self supportsRangeManagedInterfaceState]) {
+          newState = ASInterfaceStateNone;
         }
-      });
+        self.interfaceState = newState;
+      }
+    };
+
+    if ([[ASCATransactionQueue sharedQueue] disabled]) {
+      dispatch_async(dispatch_get_main_queue(), exitVisibleInterfaceState);
```
appleguy: Note that this change intentionally alters the behavior of the -disabled state as well. After the change, the -didExitVisibleState call is held for a dispatch_async to confirm the node is still invisible, for both range-managed and not range-managed nodes (previously there was a check distinguishing these).

This will decrease the accuracy of the call slightly, compared to before the change and also compared to interfaceState coalescing which guarantees the call is done on the *same* runloop / implicit transaction.

However it will also address a very common issue where visibility calls thrash to invisible and visible twice during the UIViewController transition process, and this feels like a better / safer behavior to prefer rather than the utmost accuracy of the invisibility call.

The visibility call remains un-delayed, and that one is much more important. Gating the invisibility is sufficient to avoid the thrashing since the visibility call won't re-fire if the interfaceState remains visible.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/8a4b65437f80374d9bd99685988de0f9/raw/8523746e9c2ac5a68c08e5ac76cec3944a4a9554/log.txt)

Source/ASDisplayNode.mm
```diff
@@ -2868,25 +2872,54 @@ - (ASInterfaceState)interfaceState
...
+  ASDN::MutexLocker l(__instanceLock__);
+  if ([[ASCATransactionQueue sharedQueue] disabled]) {
+    [self applyPendingInterfaceState:newState];
+  } else {
+    ASDN::MutexLocker l(__instanceLock__);
+    if (_pendingInterfaceState != newState) {
+      _pendingInterfaceState = newState;
+      [[ASCATransactionQueue sharedQueue] enqueue:self];
+    }
+  }
```
Adlai-Holler: Remove the lock at the top of this method. It isn't needed and `applyPendingInterfaceState` asserts that we aren't locked. (`_unlocked_applyPendingInterfaceState:`?)

Source/ASRunLoopQueue.mm
```diff
@@ -469,3 +489,230 @@ - (void)unlock
...
+  NSPointerArray *_internalQueue; // Use NSPointerArray so we can decide __strong or __weak per-instance.
+  ASDN::RecursiveMutex _internalQueueLock;
+  BOOL _disableInterfaceStateCoalesce;
+
+  // In order to not pollute the top-level activities, each queue has 1 root activity.
+  os_activity_t _rootActivity;
+
+#if ASRunLoopQueueLoggingEnabled
+  NSTimer *_runloopQueueLoggingTimer;
+#endif
```
Adlai-Holler: Can we not move the duplicated ivars up into ASAbstractRunLoopQueue?

Adlai-Holler: The comment on `_internalQueue` doesn't apply to this class; we always use it in strong mode.

Source/ASRunLoopQueue.mm
```diff
@@ -469,3 +489,230 @@ - (void)unlock
...
+  }
+  return self;
+}
+
+- (void)dealloc
+{
+  if (CFRunLoopContainsSource(_runLoop, _runLoopSource, kCFRunLoopCommonModes)) {
+    CFRunLoopRemoveSource(_runLoop, _runLoopSource, kCFRunLoopCommonModes);
+  }
+  CFRelease(_runLoopSource);
```
Adlai-Holler: We can unconditionally remove the source – documentation says that if the source isn't present the function does nothing.

Source/ASRunLoopQueue.h
```diff
@@ -48,6 +55,30 @@ AS_SUBCLASSING_RESTRICTED
...
+@property (nonatomic, readonly) BOOL isEmpty;
+@property (nonatomic, readonly) BOOL disabled;
+/**
+ * The queue to run on main run loop before CATransaction commit.
+ *
+ * @discussion this queue will run after ASRunLoopQueue and before CATransaction commit
+ * to get last chance of updating/coalesce info like interface state.
+ * Each node will only be called once per transaction commit to reflect interface change.
+ */
++ (ASCATransactionQueue *)sharedQueue;
```
Adlai-Holler: Declare this as an atomic readonly class property.

Source/ASRunLoopQueue.h
```diff
@@ -48,6 +55,30 @@ AS_SUBCLASSING_RESTRICTED
 
 @end
 
+AS_SUBCLASSING_RESTRICTED
+@interface ASCATransactionQueue : ASAbstractRunLoopQueue
+
+@property (nonatomic, readonly) BOOL isEmpty;
+@property (nonatomic, readonly) BOOL disabled;
```
Adlai-Holler: These properties need to be atomic.

Source/ASRunLoopQueue.mm
```diff
@@ -469,3 +489,230 @@ - (void)unlock
...
+#endif
+}
+
+@end
+
+@implementation ASCATransactionQueue
+
+// CoreAnimation commit order is 2000000, the goal of this is to process shortly beforehand
+// but after most other scheduled work on the runloop has processed.
+static int const kASASCATransactionQueueOrder = 1000000;
```
Adlai-Holler: We can always update this later, since the current version of the diff brings obvious benefits and has been productionized, but in an ideal world would we not want to do this work at the end of the layout phase of the commit?

For example, collection views will not update their visible cells until the layout pass, so we'd have to wait for the next go-round for the interface state to be updated to reflect the truth. For a ridiculous case, imagine I have a cell node that is red whenever it's invisible and blue whenever it's visible. The current implementation would show red for 1 frame and then jump to blue.

One weird but viable option is for us to do something like this in the observer:

```objc
while (true) {
  // Layout visible windows (CA is about to do this anyway)
  for (window in app.windows) {
    if (!window.hidden) {
      [window layoutIfNeeded];
    }
  }

  // If we have no work to do, break.
  if ([self isEmpty]) {
    break;
  }

  // Otherwise notify observers and loop again.
  [self processQueue];
}
```

That way any layout invalidations that happen due to interface states, and any interface state changes due to layout, are all settled. We would also use a loop counter to limit the maximum recursive dirtying (just like CA).

Source/ASRunLoopQueue.mm
```diff
@@ -469,3 +489,230 @@ - (void)unlock
...
+  ASSignpostEnd(ASSignpostRunLoopQueueBatch);
+}
+
+- (void)enqueue:(id<ASCATransactionQueueObserving>)object
+{
+  if (!object) {
+    return;
+  }
+
+  if (_disableInterfaceStateCoalesce == YES) {
```
Adlai-Holler: Nit: remove `== YES`, I know it's used in other places but we can stop the madness.

Source/ASRunLoopQueue.h
```diff
@@ -48,6 +55,30 @@ AS_SUBCLASSING_RESTRICTED
...
+ * Each node will only be called once per transaction commit to reflect interface change.
+ */
++ (ASCATransactionQueue *)sharedQueue;
+
+- (void)enqueue:(id<ASCATransactionQueueObserving>)object;
+
+/**
+ * @abstract Apply a node's interfaceState immediately rather than adding to the queue.
+ */
+- (void)disableInterfaceStateCoalesce;
```
Adlai-Holler: I would just call this method `-disable`, since this queue isn't supposed to know anything about interface states per-se.

appleguy:
Thanks for the feedback @Adlai-Holler !  BTW @nguyenhuy, regarding the commit message in https://github.com/TextureGroup/Texture/commit/ea547270f2d05e015167ebdaaac48c95318fe9d5 -- something in the direction of this change could help.

However the trouble is that there are two use cases, one that may require layout to be complete, and the other which doesn't:
- Normal use of the visibility call to create an expensive child and add it, catching the layout pass (before layout).
- The impression logging case where it needs to look at layout information.

One thing I wonder is whether the impression logging case could be changed to read .calculatedLayout instead. If not, we may need two calls: willEnterVisibleState and didEnterVisibleState (but that would be a breaking semantic change), or didEnterVisibleState[...with layout complete].

Source/ASDisplayNode.mm
wsdwsd0829: Yes, this is a mis-merge, thanks for spotting that.

wsdwsd0829:
Thanks for reviewing. @Adlai-Holler 
I changed most suggests.  I am thinking to do a new pr on extracting common behavior to AbstractQueue because it may need to take a closer look. The two queues are different in that the ASRunloopQueue is call back/block based  while ASCATransaction queue is delegate based. 
Also a test fail in CI should be resolved here https://github.com/TextureGroup/Texture/pull/790.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/cc3244e019ac1fd19abf25fd5b44319d/raw/343b9342304c16cdee76532fe3c29f795d4a0205/log.txt)

Adlai-Holler:
@appleguy @wsdwsd0829 Good points, I don't want to drop frames but I wonder if we're prematurely optimizing by trying to avoid triggering the layout pass ourselves.

In the common case, the first layout pass will just be applying layouts that have already been generated (cells measured in background). If the user adds a node in response to the visibility notification, the synchronous remeasure will be the vast majority of the work and it's unavoidable either way.

Further, I find it semantically incorrect to notify the user about visibility before layout is complete. In CA, nothing ever becomes visible until layout finishes.

In practical terms, again consider that in the collection node case, any changes that the user makes to the node in response to visibility notifications will be delayed. If they add a subnode for example, we will still see the cell missing that subnode for a frame, since the interface states won't be updated until the collection view `layoutSubviews`. And this is how it should be, so it's inappropriate to try and hack around that specific case.

If I can find time soon I will profile the collection view example with both approaches and test out things like adding a subnode on visible. 

I feel pretty strongly about this because our handling of the visible interface state has needed correction for a while and caused Pinterest no end of headaches because of the mismatch between the CA state and the ASDK state. Let's take the time to get this as robust as possible, while still respecting the production-validation that has gone into this patch as it is today.

nguyenhuy:
@wsdwsd0829 Thanks for the diff. Very nice work!

I have to agree with @Adlai-Holler that it'd be much better to fire off visibility events after the phase of the CA transaction. I have first-hand experience dealing with Pinterest problems, and this is the best approach I can think of to fix them once and for all.

I think the most important agreement we need to make now is that our definition of "visible" should be consistent with CA's, that is "after the layout pass". And with that, `didEnterVisibleState` should be called afterward.

I do understand the use case for handling expensive child upon visibility. It wouldn't be terrible to add another event type, say "pre visible" or "will be visible", and provide a new API that will be called *before* the layout pass for any changes that should be performed in response to such event.

Having said that, it seems to me that such API will add complexity without much benefits. If an expensive child needs to be prepared on main, it will cause frame drops regardless of whether the preparation happens before or after the layout phase. Perhaps, users should minimize the work on main and/or split it up across multiple interface states (preload and render).

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e2267874cc79288a88829ee22b39b454/raw/c94ac0a5025bee28f22813759ba70ed2c64f78e3/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/cb4b4256b3a04bc7b553807e583ad9cb/raw/8bae402f64b041a9fdd40919ed167a17e52f44d4/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f8186ee87857b74c37b1f57c959baefa/raw/4a23fc6c05a391a1a8aac6dd523c4fe2ec6357f7/log.txt)

wsdwsd0829:
Hi, I made some final changes by applying interface immediately for any interface updates during processing of queue. Please talk another look at and then we should ready to ship. 
thanks.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRunLoopQueueTests.m:</p>
<pre><code>//
//  ASRunLoopQueueTests.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/800ef06b6ffcff498c88a62825f70bb0/raw/14fa3985467100e4e5167107a3d439004d40b136/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRunLoopQueueTests.m:</p>
<pre><code>//
//  ASRunLoopQueueTests.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/21ac9d4550d2e74ebc7714ded7345510/raw/a257630007188887f8d4d8e91f815323197c9b2b/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRunLoopQueueTests.m:</p>
<pre><code>//
//  ASRunLoopQueueTests.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1faf1dae34d4d09e8d958ba7b2e16732/raw/535fc786c654225b647358613ff5ffe447c7e06e/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRunLoopQueueTests.m:</p>
<pre><code>//
//  ASRunLoopQueueTests.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASRunLoopQueue.mm
```diff
@@ -464,3 +484,240 @@ - (void)unlock
...
+      }
+    }
+
+    if (foundItemCount == 0) {
+      // If _internalQueue holds weak references, and all of them just become NULL, then the array
+      // is never marked as needsCompletion, and compact will return early, not removing the NULL's.
+      // Inserting a NULL here ensures the compaction will take place.
+      // See http://www.openradar.me/15396578 and https://stackoverflow.com/a/40274426/1136669
+      [_internalQueue addPointer:NULL];
+    }
```
cesteban: If the queue only holds strong references, this trick shouldn't be necessary.

appleguy: @cesteban awesome catch, thanks for reviewing! @wsdwsd0829 this would be a good change to make, since we expect this queue to always be strong (otherwise the node could dealloc before getting the invisible state) - so I think we can just delete these few lines.

wsdwsd0829: Thanks, fixed.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASRunLoopQueueTests.m:</p>
<pre><code>//
//  ASRunLoopQueueTests.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

Adlai-Holler:
@wsdwsd0829 One more thing is just to update the license header. It should skip CI and we can land today.

PR #789, opened 2018-02-05
Fix UIResponder handling with view backing ASDisplayNode

maicki:
### What is the problem?
If certain `UIResponder` methods are overwritten in a `ASDisplayNode` subclass they are not considered. The only two methods that are currently properly forwarded are: `canBecomeFirstResponder` and `canResignFirstResponder` in _ASDisplayView.mm. Every other UIResponder method will never reach any `_ASDisplayView`. This especially is problem for `ASEditableTextNode` as we overwrite these methods in there and expect to be considered.

Example repo: https://github.com/maicki/TextureResponderChain

### Tests
I added a test that show's this invalid behavior. You can find them under `ASDisplayNodeTests.mm` and `ASTableViewTests.mm`

###  Some more details
The way I actually stumbled upon the issue is that as soon as an `ASEditableTextNode` is a subnode of a `ASCellNode` as well as a first responder, calling `reloadData` on an `ASTableNode` will **not** reload the `UITableView`. UIKit will bail out within the process.
Internally UIKit's `UITableView` is trying to resign the first responder from the current first responder on an `reloadData` what is the `textView` of the `ASEditableTextNode` and afterwards expects no view in the `UITableView` hierarchy is the first responder anymore. That will not be case if `UITableView` is calling `resignFirstResponder` on the `ASEditableTextNode.textView` as we push only the first responder to the `ASEditableTextNode.view`. `UITableView` has an ivar called `_firstResponderView` where it determines if the first responder view is a view in the `UITableView` hierarchy. If this is ivar != null it will bail out if `reloadData` is called, before actually reloading the data (I verified that within Hopper).
And so if the `UITableView` is calling `resignFirstResponder` on the `ASEditableTextNode.textView` the `_firstResponderView` will just be `ASEditableTextNode.view`, but it should be nil. If we properly forward all responder methods `_firstResponderView` will be nil as the `ASEditableTextNode.view` will not become the first responder if the table view calls `resignFirstResponder` on the `ASEditableTextNode.textView`.

### Current solution
```
// Called canBecomeFirstResponder on _ASDisplayView
-> [_ASDisplayView canBecomeFirstResponder]
  -> if subclass of _ASDisplayView overwrites canBecomeFirstResponder
      // Prevent an infinite loop in here if [super canBecomeFirstResponder] was called
      -> if subclass of ASDisplayNode overwrites canBecomeFirstResponder
        -> [ASDisplayNode canBecomeFirstResponder];
      -> else
        // Call through to views superclass as we expect super was called from the
        // _ASDisplayView subclass and node does not implement canBecomeFirstResponder
        -> [_ASDisplayView __canBecomeFirstResponder]
          -> [super canBecomeFirstResponder]
  -> else
      // Call through to the display node to give a chance to dive in
    -> [ASDisplayNode __canBecomeFirstResponder]

// Called canBecomeFirstResponder on ASDisplayNode
-> [ASDisplayNode+UIViewBridge canBecomeFirstResponder]
  // Just call through to our internal method
  -> [ASDisplayNode (self) __canBecomeFirstResponder]

// Called form _ASDisplayView and ASDisplayNode from ASDisplayNode+UIViewBridge
-> [ASDisplayNode __canBecomeFirstResponder]
  -> if view it is a subclass of _ASDisplayView (synchronous == TRUE)
    // If it's synchronous just call through to the view as we expect it's
    // a non _ASDisplayView subclass that will respond
    -> [_view canBecomeFirstResponder]
  -> else
      -> if subclass of _ASDisplayView overwrites canBecomeFirstResponder
        // If the subclass overwrites canBecomeFirstResponder just call through
        // to it as we expect it will handle it
        [_view canBecomeFirstResponder];
      -> else
        // Call through to _ASDisplayView to get it handled
        -> [_ASDisplayView __canBecomeFirstResponder]
          -> [super canBecomeFirstResponder]
```

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -856,7 +856,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     return NO;
   }
   
-  if (checkFlag(Synchronous)) {
+  if (checkFlag(Synchronous)
+      || ASSubclassOverridesSelector([_ASDisplayView class], _viewClass, @selector(canBecomeFirstResponder))) {
     return [_view canBecomeFirstResponder];
   } else {
     return [(_ASDisplayView *)_view __canBecomeFirstResponder];
@@ -865,15 +866,19 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (BOOL)__becomeFirstResponder
 {
-  if (self.isLayerBacked || ![self __canBecomeFirstResponder]) {
+  
+  if (self.isLayerBacked || ![self canBecomeFirstResponder]) {
     return NO;
   }
-
+  
   // We explicitly create the view in here
-  if (checkFlag(Synchronous)) {
+  [self view];
+
+  if (checkFlag(Synchronous)
+      || ASSubclassOverridesSelector([_ASDisplayView class], _viewClass, @selector(becomeFirstResponder))) {
     return [_view becomeFirstResponder];
   } else {
-    return [(_ASDisplayView *)_view __canBecomeFirstResponder];
+    return [(_ASDisplayView *)_view __becomeFirstResponder];
   }
 }
 
@@ -884,7 +889,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     return YES;
   }
 
-  if (checkFlag(Synchronous)) {
+  if (checkFlag(Synchronous)
+      || ASSubclassOverridesSelector([_ASDisplayView class], _viewClass, @selector(canResignFirstResponder))) {
     return [_view canResignFirstResponder];
   } else {
     return [(_ASDisplayView *)_view __canResignFirstResponder];
@@ -897,7 +903,11 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     return NO;
   }
   
-  if (checkFlag(Synchronous)) {
+  // We explicitly create the view in here
+  [self view];
+  
+  if (checkFlag(Synchronous)
+      || ASSubclassOverridesSelector([_ASDisplayView class], _viewClass, @selector(resignFirstResponder))) {
     return [_view resignFirstResponder];
   } else {
     return [(_ASDisplayView *)_view __resignFirstResponder];
@@ -911,7 +921,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     return NO;
   }
   
-  if (checkFlag(Synchronous)) {
+  if (checkFlag(Synchronous)
+      || ASSubclassOverridesSelector([_ASDisplayView class], _viewClass, @selector(isFirstResponder))) {
     return [_view isFirstResponder];
   } else {
     return [(_ASDisplayView *)_view __isFirstResponder];
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -874,7 +874,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (BOOL)__becomeFirstResponder
 {
-  
   if (self.isLayerBacked || ![self canBecomeFirstResponder]) {
     return NO;
   }
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -540,11 +540,14 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     _viewBlock = nil;
     _viewClass = [view class];
   } else {
+    
     view = [[_viewClass alloc] init];
   }
   
   // Special handling of wrapping UIKit components
   if (checkFlag(Synchronous)) {
+    [self checkResponderCompatibility];
+    
     // UIImageView layers. More details on the flags
     if ([_viewClass isSubclassOfClass:[UIImageView class]]) {
       _flags.canClearContentsOfLayer = NO;
@@ -826,6 +829,93 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 {
   ASDN::MutexLocker l(__instanceLock__);
   _flags.viewEverHadAGestureRecognizerAttached = YES;
+}
+
+#pragma mark UIResponder
+
+- (void)checkResponderCompatibility
+{
+#if ASDISPLAYNODE_ASSERTIONS_ENABLED
+  // There are certain cases we cannot handle and are not supported:
+  // 1. If the _view class is not a subclass of _ASDisplayView
+  if (checkFlag(Synchronous)) {
+    // 2. At least one UIResponder methods are overwritten in the node subclass
+    ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self.class, @selector(canBecomeFirstResponder)), @"TBD");
+    ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self.class, @selector(becomeFirstResponder)), @"TBD");
+    ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self.class, @selector(canResignFirstResponder)), @"TBD");
+    ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self.class, @selector(resignFirstResponder)), @"TBD");
+    ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self.class, @selector(isFirstResponder)), @"TBD");
+  }
+#endif
+}
+
+- (BOOL)__canBecomeFirstResponder
+{
+  if (_view == nil) {
+    // By default we return NO if not view is created yet
+    return NO;
+  }
+  
+  if (checkFlag(Synchronous)) {
+    return [_view canBecomeFirstResponder];
+  } else {
+    return [(_ASDisplayView *)_view __canBecomeFirstResponder];
+  }
+}
+
+- (BOOL)__becomeFirstResponder
+{
+  if (self.isLayerBacked || ![self __canBecomeFirstResponder]) {
+    return NO;
+  }
+
+  // We explicitly create the view in here
+  if (checkFlag(Synchronous)) {
+    return [_view becomeFirstResponder];
+  } else {
+    return [(_ASDisplayView *)_view __canBecomeFirstResponder];
+  }
+}
+
+- (BOOL)__canResignFirstResponder
+{
+  if (_view == nil) {
+    // By default we return YES if no view is created yet
+    return YES;
+  }
+
+  if (checkFlag(Synchronous)) {
+    return [_view canResignFirstResponder];
+  } else {
+    return [(_ASDisplayView *)_view __canResignFirstResponder];
+  }
+}
+
+- (BOOL)__resignFirstResponder
+{
+  if (self.isLayerBacked || ![self canResignFirstResponder]) {
+    return NO;
+  }
+  
+  if (checkFlag(Synchronous)) {
+    return [_view resignFirstResponder];
+  } else {
+    return [(_ASDisplayView *)_view __resignFirstResponder];
+  }
+}
+
+- (BOOL)__isFirstResponder
+{
+  if (_view == nil) {
+    // If no view is created yet we can just return NO as it's unlikely it's the first responder
+    return NO;
+  }
+  
+  if (checkFlag(Synchronous)) {
+    return [_view isFirstResponder];
+  } else {
+    return [(_ASDisplayView *)_view __isFirstResponder];
+  }
 }
 
 #pragma mark <ASDebugNameProvider>
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -856,11 +856,19 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     return NO;
   }
   
-  if (checkFlag(Synchronous)
-      || ASSubclassOverridesSelector([_ASDisplayView class], _viewClass, @selector(canBecomeFirstResponder))) {
+  if (checkFlag(Synchronous)) {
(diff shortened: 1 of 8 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/fc517a3b816f9ce2e572b36d21884efc/raw/a8ae6ad46b6858bba6c7a1e26d23c82f5f005974/log.txt)

Source/ASDisplayNode.mm
```diff
@@ -828,6 +831,104 @@ - (void)nodeViewDidAddGestureRecognizer
...
+
+- (BOOL)__canBecomeFirstResponder
+{
+  if (_view == nil) {
+    // By default we return NO if not view is created yet
+    return NO;
+  }
+  
+  if (checkFlag(Synchronous)
+      || ASSubclassOverridesSelector([_ASDisplayView class], _viewClass, @selector(canBecomeFirstResponder))) {
```
maicki: We may should catch all of them in the initialization phase like `ASDisplayNodeMethodOverrides`. That said at that point we may not know the actual view class so we only could do it after the view is created.

Source/ASDisplayNode.mm
```diff
@@ -828,6 +831,123 @@ - (void)nodeViewDidAddGestureRecognizer
...
+- (BOOL)__becomeFirstResponder
+{
+  if (self.isLayerBacked || ![self canBecomeFirstResponder]) {
+    return NO;
+  }
+  
+  // We explicitly create the view in here as it's supposed to become a first responder
+  [self view];
+
+  if (checkFlag(Synchronous)) {
```
maicki: This should probably extracted to a macro as the only change is the selector

Source/Details/_ASDisplayView.mm
```diff
@@ -358,14 +359,110 @@ - (void)tintColorDidChange
...
+    return [node __canBecomeFirstResponder];
+  }
 }
 
-- (BOOL)canResignFirstResponder {
-  ASDisplayNode *node = _asyncdisplaykit_node; // Create strong reference to weak ivar.
-  return [node canResignFirstResponder];
+- (BOOL)becomeFirstResponder
+{
+  ASDisplayNode *node = _asyncdisplaykit_node;
```
maicki: This parts also be extracted to a macro.

timonus:
Thank you so much for fixing this Michael!

appleguy:
Amazing analysis and fix @maicki !

I never knew about this behavior, but it totally makes sense to avoid text fields being cleared by accidental app behaviors when a user is actively editing them.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b9957e9d915e36413c7cbee354a4f7f2/raw/f647cd0487fb5c840efc7742e87e4ef880159526/log.txt)

appleguy:
@maicki thanks for the ping, I'll take a look soon! Thanks for the really awesome description, and analysis underlying it :).

Source/ASDisplayNode.mm
```diff
@@ -828,6 +830,94 @@ - (void)nodeViewDidAddGestureRecognizer
...
+    } \
+  } \
+
+- (void)checkResponderCompatibility
+{
+#if ASDISPLAYNODE_ASSERTIONS_ENABLED
+  // There are certain cases we cannot handle and are not supported:
+  // 1. If the _view class is not a subclass of _ASDisplayView
+  if (checkFlag(Synchronous)) {
+    // 2. At least one UIResponder methods are overwritten in the node subclass
```
appleguy: Any concerns with touchesBegan:, touchesMoved:, touchesEnded:, or touchesCancelled: ?

Source/ASDisplayNode.mm
```diff
@@ -828,6 +830,94 @@ - (void)nodeViewDidAddGestureRecognizer
   _flags.viewEverHadAGestureRecognizerAttached = YES;
 }
 
+#pragma mark UIResponder
+
+#define HANDLE_RESPONDER_METHOD(__sel) \
+  if (checkFlag(Synchronous)) { \
```
appleguy: This should probably assert main thread. At least -becomeFirstResponder might be reasonable for a developer to expect to be a UIViewBridge property and threadsafe to call, but this implementation triggers view creation.

Maybe worth even a special assertion in -becomeFirstResponder that tells developers to call that method on the main thread or in onDidLoad / -didLoad.

Source/ASDisplayNode.mm
```diff
@@ -828,6 +830,94 @@ - (void)nodeViewDidAddGestureRecognizer
...
+    // By default we return NO if not view is created yet
+    return NO;
+  }
+  
+  HANDLE_RESPONDER_METHOD(canBecomeFirstResponder);
+}
+
+- (BOOL)__becomeFirstResponder
+{
+  if (self.isLayerBacked || ![self canBecomeFirstResponder]) {
```
appleguy: Check for isLayerBacked is probably redundant, and also grabs lock. Could check _view if a short circuit is needed, as this method is main thread only then lock isn't needed there.

Source/ASDisplayNode.mm
```diff
@@ -828,6 +830,94 @@ - (void)nodeViewDidAddGestureRecognizer
   _flags.viewEverHadAGestureRecognizerAttached = YES;
 }
 
+#pragma mark UIResponder
```
appleguy: IMO this should go in another file, since ASDisplayNode.mm is quite large - it seems very appropriate for UIViewBridge or perhaps we could introduce another one.

Not a blocking change (feel free to land this if you're just eager to get it in!)

Source/Details/_ASDisplayView.h
```diff
@@ -39,6 +39,14 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)__forwardTouchesEnded:(NSSet *)touches withEvent:(UIEvent *)event;
 - (void)__forwardTouchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event;
 
+// These methods expose a way for ASDisplayNode responder methods to let the view call super responder methods
```
appleguy: This first sentence isn't quite clear -- "for ASDisplayNode responder methods to let the view call super responder methods".

Clarify if possible, otherwise not exactly sure how to improve it.

Source/Details/_ASDisplayView.mm
```diff
@@ -358,14 +359,79 @@ - (void)tintColorDidChange
   [node tintColorDidChange];
 }
 
-- (BOOL)canBecomeFirstResponder {
-  ASDisplayNode *node = _asyncdisplaykit_node; // Create strong reference to weak ivar.
-  return [node canBecomeFirstResponder];
+#pragma mark UIResponder Handling
+
+#define HANDLE_RESPONDER_METHOD(__sel) \
```
appleguy: Confusing to have both macros named the same - maybe call this one HANDLE_VIEW_RESPONDER_METHOD and the other one HANDLE_NODE_RESPONDER_METHOD?

Source/Details/_ASDisplayView.mm
```diff
@@ -358,14 +359,79 @@ - (void)tintColorDidChange
...
+  ASDisplayNode *node = _asyncdisplaykit_node; \
+  SEL sel = @selector(__sel); \
+  /* Prevent an infinite loop in here if [super canBecomeFirstResponder] was called on a
+  / _ASDisplayView subclass */ \
+  if (ASSubclassOverridesSelector([_ASDisplayView class], [self class], sel)) { \
```
appleguy: I think these checks are going to become costly; we should verify in a profiler. Since only a handful of selectors are involved, ideally these should be performed at +initialize like other subclass implementation checks are.

Source/Details/_ASDisplayView.mm
```diff
@@ -358,14 +359,79 @@ - (void)tintColorDidChange
...
+    } \
+  } else { \
+    /* Call through to internal node __canBecomeFirstResponder that will consider the view in responding */ \
+    return [node __##__sel]; \
+  } \
+
+
+- (BOOL)canBecomeFirstResponder
+{
+  HANDLE_RESPONDER_METHOD(canBecomeFirstResponder);
```
appleguy: I think you could use _cmd for each of these. They could also be made a macro for IMPLEMENT_RESPONDER_METHOD instead of HANDLE_, and define the method as well as its contents.

Source/Details/_ASDisplayView.mm
```diff
@@ -358,14 +359,79 @@ - (void)tintColorDidChange
   [node tintColorDidChange];
 }
 
-- (BOOL)canBecomeFirstResponder {
-  ASDisplayNode *node = _asyncdisplaykit_node; // Create strong reference to weak ivar.
-  return [node canBecomeFirstResponder];
+#pragma mark UIResponder Handling
+
+#define HANDLE_RESPONDER_METHOD(__sel) \
+  ASDisplayNode *node = _asyncdisplaykit_node; \
```
appleguy: Can you keep the comment // Create a strong reference to weak ivar, from the original code? It would be easy to think this could be dropped and accessed directly on the ivar, but I assume the implication is that the ivar might change or become invalid.

Source/Details/_ASDisplayView.mm
```diff
@@ -358,14 +359,79 @@ - (void)tintColorDidChange
...
+
+- (BOOL)isFirstResponder
+{
+  HANDLE_RESPONDER_METHOD(isFirstResponder);
+}
+
+// This methods are called from ASDisplayNode to let the view decide in what UIResponder state they are not overwritten
+// by a ASDisplayNode subclass
+
+- (BOOL)__canBecomeFirstResponder
```
appleguy: Technically these could be macro-ized as well, generated in a pair with the above ones :). Optional of course.

Nits: 2 spaces instead of 4 on the returns. Also, "overridden" instead of "overwritten" in the comment.

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -146,23 +136,34 @@ - (UIView *)preferredFocusedView
...
+}
+
 - (BOOL)isFirstResponder
 {
   ASDisplayNodeAssertMainThread();
-  return _view != nil && [_view isFirstResponder];
+  return [self __isFirstResponder];
 }
 
-// Note: this implicitly loads the view if it hasn't been loaded yet.
```
appleguy: Seems like a good comment to keep.

Tests/ASDisplayNodeTests.mm
```diff
@@ -256,6 +261,25 @@ - (BOOL)resignFirstResponder {
...
+}
+@end
+
+@interface UIResponderNodeTestViewCallingSuper : UIView
+@end
+@implementation UIResponderNodeTestViewCallingSuper
+- (BOOL)canBecomeFirstResponder { return YES; }
+- (BOOL)becomeFirstResponder {
+  return [super becomeFirstResponder];
+
```
appleguy: Extra return

Source/ASDisplayNode.mm
maicki: I didn't look into them or saw any issues related to them.

Source/Details/_ASDisplayView.h
maicki: The sentence is kind of the same as the one for all of the forwarding touch events. In that case we should find a better way for both.

Source/Private/ASDisplayNode+UIViewBridge.mm
maicki: We have this kind of comment in `__becomeFirstResponder` now:
```
// Note: this implicitly loads the view if it hasn't been loaded yet.
```

Let me know if I should also add it to this place back again.

Source/ASDisplayNode.mm
maicki: Agree, I added a main thread assertion for all responder methods.

Source/ASDisplayNode.mm
maicki: We actually should be able to get rid of the `isLayerBacked` call at all. We check for the `_view` in `canBecomeFirstResponder` anyway that should cover that case.

Source/Details/_ASDisplayView.mm
maicki: Renamed them.

Source/ASDisplayNode.mm
maicki: We are currently have the event handling `touchesBegan:` etc. in `ASDisplayNode.mm`, eventually we should move both of them out too in a follow up diff.

Source/Details/_ASDisplayView.mm
maicki: Added in back

Source/Details/_ASDisplayView.mm
maicki: Moved them into `+initialize` same way as `ASDisplayNode` does.

Source/Details/_ASDisplayView.mm
maicki: Moved all into one macros and called it `IMPLEMENT_RESPONDER_METHOD`

Source/Details/_ASDisplayView.mm
maicki: Also moved this into the macro and to 2 spaces.

maicki:
@appleguy Thank you for reviewing. Can you take another look over it please. Thanks!

Tests/ASTableViewTests.mm
```diff
@@ -833,6 +835,52 @@ - (void)testAutomaticallyAdjustingContentOffset
...
+  [editableTextNode.textView becomeFirstResponder];
+  
+  // Change data source count and try to reload a second time
+  dataSource.rowsPerSection = 2;
+  [node reloadData];
+  [node waitUntilAllUpdatesAreProcessed];
+  
+  // Check that numberOfRows in section 0 is 2
+  XCTAssertTrue([node numberOfRowsInSection:0] == 2, @"Number of rows in section 0 should be 2 after reload");
+  XCTAssertTrue([node.view numberOfRowsInSection:0] == 2, @"Number of rows in section 0 should be 2 after reload");
```
Adlai-Holler: Nit: use XCTAssertEqual so that we can see the failing value if this fails on CI in the future.

PR #792, opened 2018-02-06
[ASDisplayNode layout] Fix an issue that sometimes causes a node's pending layout to not be applied

nguyenhuy:
- Since the implementation of layout version (#428), if a node's pending and calculated layouts have the same current version as well as the same constrained size, the 2 layouts are considered equal and can be used interchangeably. A layout version check between the 2 layouts was added in #695. This PR adds a missing constrained size check.
- If the pending layout has the same version but a different constrained size compare to the calculated layout's, we can assume that the pending layout is newer and should be preferred over the calculated one. That is because layout operations always register their new layout as pending, which then (immediately or eventually) get applied to the node as calculated layout.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASDisplayNode+Layout.mm:</p>
<pre><code>//
//  ASDisplayNode+Layout.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

nguyenhuy:
@Adlai-Holler You're right that we can use whichever layout fits the constrained size/bounds. However, currently `_layoutSublayouts` only picks up `_calculated` layout, so we need to fix that as well.

Another concern is that, such behaviour will break the current implications for `_pending` and `_calculated`. We'll end up with 2 (or more) calculated layouts (and 0 pending layout) from which the node can choose and apply. It's a better behaviour that fits our vision of the new layout system. However, since it'll require more time and effort to implement and test, I'd prefer to land this diff as is, because it makes the current system to behave correctly, and follow up later. Thoughts?

nguyenhuy:
Awesome. Thank you!

PR #798, opened 2018-02-10
[ASDisplayNode] Consolidate main thread initialization and allow apps to invoke it manually instead of +load.

appleguy:
Additionally this has a few minor fixes for Yoga support, and adds some basic but universally valuable callbacks like -nodeDidLoad to ASNodeController.

### DIFF
--- a/Source/ASDisplayNode+Yoga.mm
+++ b/Source/ASDisplayNode+Yoga.mm
@@ -157,6 +157,8 @@
 
 - (void)setupYogaCalculatedLayout
 {
+  ASDN::MutexLocker l(__instanceLock__);
+
   YGNodeRef yogaNode = self.style.yogaNode;
   uint32_t childCount = YGNodeGetChildCount(yogaNode);
   ASDisplayNodeAssert(childCount == self.yogaChildren.count,
@@ -213,7 +215,7 @@
     // For the root node in a Yoga tree, make sure to preserve the constrainedSize originally provided.
     // This will be used for all relayouts triggered by children, since they escalate to root.
     ASSizeRange range = parentNode ? ASSizeRangeUnconstrained : self.constrainedSizeForCalculatedLayout;
-    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, range, parentSize);
+    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, range, parentSize, _layoutVersion);
   }
 }
 
--- a/Source/ASDisplayNode+Yoga.mm
+++ b/Source/ASDisplayNode+Yoga.mm
@@ -210,7 +210,10 @@
       parentSize.width = YGNodeLayoutGetWidth(parentNode);
       parentSize.height = YGNodeLayoutGetHeight(parentNode);
     }
-    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, ASSizeRangeUnconstrained, parentSize, 0);
+    // For the root node in a Yoga tree, make sure to preserve the constrainedSize originally provided.
+    // This will be used for all relayouts triggered by children, since they escalate to root.
+    ASSizeRange range = parentNode ? ASSizeRangeUnconstrained : self.constrainedSizeForCalculatedLayout;
+    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, range, parentSize);
   }
 }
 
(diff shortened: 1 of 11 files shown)


Source/ASDisplayNode+Yoga.mm
```diff
@@ -210,7 +210,10 @@ - (void)setupYogaCalculatedLayout
       parentSize.width = YGNodeLayoutGetWidth(parentNode);
       parentSize.height = YGNodeLayoutGetHeight(parentNode);
     }
-    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, ASSizeRangeUnconstrained, parentSize, 0);
+    // For the root node in a Yoga tree, make sure to preserve the constrainedSize originally provided.
+    // This will be used for all relayouts triggered by children, since they escalate to root.
+    ASSizeRange range = parentNode ? ASSizeRangeUnconstrained : self.constrainedSizeForCalculatedLayout;
```
appleguy: Without this fix, we had a pretty serious issue where a simple relayout would cause a Yoga tree to jump to a different layout, because it used Unconstrained instead of its original constrained size.

By calling through to self.constrainedSizeForCalculatedLayout, the behavior should be a pretty compatible integration with the Texture layout system.

My app has been running with these fixes to Yoga for several months in production, so at least for the limited scope of apps using Yoga, I'm pretty confident they are an improvement over prior behavior.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b0aa1e2513b1242cf38ef5b16c2d8ee5/raw/9762cc6c39831588eba01d8301bd135666cd12f2/log.txt)

Source/ASDisplayNode+Yoga.mm
```diff
@@ -235,9 +238,18 @@ - (void)updateYogaMeasureFuncIfNeeded
 - (void)invalidateCalculatedYogaLayout
 {
   YGNodeRef yogaNode = self.style.yogaNode;
-  if (yogaNode && YGNodeGetMeasureFunc(yogaNode)) {
+  if (yogaNode && [self shouldHaveYogaMeasureFunc]) {
     // Yoga internally asserts that MarkDirty() may only be called on nodes with a measurement function.
```
appleguy: This is really quite ridiculous. In any case, this behavior was required after an earlier change that un-set the measurement function unless we are about to call into Yoga to do a measurement. This is beneficial for memory management and automaticallyManagesSubnodes, but means that we need to slightly abuse the API in order to ensure we are allowed to invalidate the layout on the long-lived YGNodeRef.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e0a0be5c2eb621c1f85ee0018b38d448/raw/cebbdd28d63989f2aaef37d1d11b2ecf2fd40866/log.txt)

Adlai-Holler:
Kicked the CI. New build: https://buildkite.com/pinterest/texture/builds/1765#66cf2eec-b8f4-4938-b547-2d1d4ffed4fd

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e34e5ac43e85630bbfa4a94edfb38e2c/raw/bbb64a4bb3a72a790d290b54eeea7ebe1d7e40fa/log.txt)

Source/ASDisplayNode+Yoga.mm
```diff
@@ -210,7 +212,10 @@ - (void)setupYogaCalculatedLayout
       parentSize.width = YGNodeLayoutGetWidth(parentNode);
       parentSize.height = YGNodeLayoutGetHeight(parentNode);
     }
-    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, ASSizeRangeUnconstrained, parentSize, 0);
+    // For the root node in a Yoga tree, make sure to preserve the constrainedSize originally provided.
+    // This will be used for all relayouts triggered by children, since they escalate to root.
+    ASSizeRange range = parentNode ? ASSizeRangeUnconstrained : self.constrainedSizeForCalculatedLayout;
+    _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(layout, range, parentSize, _layoutVersion);
```
appleguy: @nguyenhuy is using _layoutVersion the correct way to handle this? I have actually not merged _layoutVersion into my branch and so was running this change without it, so this would be the biggest risk in the PR in my estimation.

Source/Private/ASInternalHelpers.m
```diff
@@ -25,6 +25,29 @@
...
 #import <AsyncDisplayKit/ASThread.h>
 
+static BOOL defaultAllowsGroupOpacity = YES;
+static BOOL defaultAllowsEdgeAntialiasing = NO;
+
+void ASInitializeFrameworkMainThread(void)
+{
+  ASDisplayNodeThreadIsMain();
+  // Ensure this value is cached on the main thread before needed in the background.
+  ASScreenScale();
```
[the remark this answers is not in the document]
Adlai-Holler: Let's remove this. I don't think it's likely the change will ever be reverted.

appleguy:
@Adlai-Holler cool, ready for a final review!

appleguy:
@maicki @Adlai-Holler Thanks guys! Could you hit merge for me? I'm not able to get past the CI being broken...

PR #802, opened 2018-02-13
Update IGListKit dependency to allow for updated versions

johntmcintosh:
The IGListKit subspec currently has the dependency hardcoded to version `3.0.0`. I've come across a need to use a later version of IGListKit, so in the process I thought it would be nice for the podspec to allow a little more flexibility in versioning of the dependency. This PR will default to the latest 3.x version, but would also allow consumers to override the IGListKit dependency in their Podfiles to be anything in the 3.x.x range.

### DIFF
--- a/Texture.podspec
+++ b/Texture.podspec
@@ -51,7 +51,7 @@ Pod::Spec.new do |spec|
   end
 
   spec.subspec 'IGListKit' do |igl|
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
This is reasonable, although we may need to restrict it again later if incompatibilities crop up. Thanks @johntmcintosh!

PR #793, opened 2018-02-07
[ASTableNode & ASCollectionNode] Keepalive reference for node if their view is necessarily alive (has a superview).

wsdwsd0829:
Some uses will try to access view and maintain it for later use. 
This PR will try to keep collection/tableNode alive as long as corresponding view still in use (may not in hierarchy), similar to ASDisplayNode & ASDisplayView.

PR #803, opened 2018-02-13
Avoid triggering main thread assertions in collection/table dealloc

Adlai-Holler:
In #793 a new assertion was introduced that inadvertently triggered main thread checker failures. 

This avoids triggering the same thing. It also improves the definition of `ASDISPLAYNODE_ASSERTIONS_ENABLED` to cover cases where `NS_BLOCK_ASSERTIONS` is defined to be 0, as some build environments do.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -172,12 +172,17 @@
   return self;
 }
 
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Alright it's time to go big. Landing!

PR #816, opened 2018-03-02
[ASDisplayNode] Add unit tests for layout z-order changes (with an open issue to fix).

appleguy:
Right now, automatic subnode management is not fixing up the z-order for nodes and their views, whenever existing subnodes are moved to a different place in the layout.

This can result in particularly weird issues when an existing layout changes with some additional nodes and some removed nodes that may displace or separate some existing nodes that also need to change order.

Once we fix, it will be easy to improve the Overlay and Background tests that are below this to also check the reversal / relayout case (I don't think they would pass currently). cc @maicki @nguyenhuy @Adlai-Holler 

```
ASDisplayNodeTests.mm:2410: error: -[ASDisplayNodeTests testThatStackSpecOrdersSubnodesCorrectly] : (([node.subnodes isEqualToArray:nodesReverseOrder]) is true) failed - subnodes: (
    "<ASDisplayNode: 0x7f8281101f90; a>",
    "<ASDisplayNode: 0x7f8281105b40; b>",
    "<ASDisplayNode: 0x7f8281105e80; c>",
    "<ASDisplayNode: 0x7f82811061c0; d>"
), reference array: (
    "<ASDisplayNode: 0x7f82811061c0; d>",
    "<ASDisplayNode: 0x7f8281105e80; c>",
    "<ASDisplayNode: 0x7f8281105b40; b>",
    "<ASDisplayNode: 0x7f8281101f90; a>"
)
Test Case '-[ASDisplayNodeTests testThatStackSpecOrdersSubnodesCorrectly]' failed (0.006 seconds).
```

### DIFF
--- a/Tests/ArrayDiffingTests.m
+++ b/Tests/ArrayDiffingTests.m
@@ -65,10 +65,14 @@
     NSIndexSet *indexSet = [test[0] _asdk_commonIndexesWithArray:test[1] compareBlock:^BOOL(id lhs, id rhs) {
       return [lhs isEqual:rhs];
     }];
+    NSMutableIndexSet *mutableIndexSet = [indexSet mutableCopy];
     
     for (NSNumber *index in (NSArray *)test[2]) {
       XCTAssert([indexSet containsIndex:[index integerValue]]);
+      [mutableIndexSet removeIndex:[index integerValue]];
     }
+
+    XCTAssert([mutableIndexSet count] == 0, @"Unaccounted deletions: %@", mutableIndexSet);
   }
 }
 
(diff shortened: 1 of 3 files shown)


Tests/ArrayDiffingTests.m
```diff
@@ -121,12 +131,20 @@ - (void)testDiffingInsertionsAndDeletions {
...
+      XCTAssert([mutableInsertions containsIndex:[index integerValue]]);
+      [mutableInsertions removeIndex:[index integerValue]];
     }
     for (NSNumber *index in (NSArray *)test[3]) {
-      XCTAssert([deletions containsIndex:[index integerValue]]);
+      XCTAssert([mutableDeletions containsIndex:[index integerValue]]);
+      [mutableDeletions removeIndex:[index integerValue]];
     }
+
+    XCTAssert([mutableInsertions count] == 0, @"Unaccounted insertions: %@", mutableInsertions);
```
appleguy: Without these, the test actually didn't fail when I tried deleting one of the supposed-to-be-present index values from the test set.

So although everything seems to behave correctly, this was important to ensure the validation checks both necessity and sufficiency of the test set.

PR #819, opened 2018-03-05
[ASTextNode] Fix an race condition on ASTextKitAttributes

hebertialmeida:
This problem was reported on #552 by @Kaspik some time ago.

Running the project with thread sanitizers enable show a bunch of errors due to a race condition on `ASTextKitAttributes`, I spent some time looking at that and I found out that the race condition is happening on `ASTextNode` when `rendererForAttributes` tries to get and set the object to cache.

This fixes the problem for me, I don't see any downsides to this implementation but of course, I would like to hear from your guys.

This could be false impressions, but I've noticed some performance improvement while rendering my app feed that has couple textNodes and I tried to measure that.

This is the average time (6 loads) to load a heavy cell with this code change: 
with: **0.119627775**
without: **0.141625549**

Issue #552, opened 2017-09-06
[Core / ASTextKitAttributes] Thread sanitizers


Kaspik:
Hey team! i know you have been looking at some main-thread access problems. Have you also been able to take a look on sanitizers? It looks like there are some issues with text attributes.

```
WARNING: ThreadSanitizer: data race (pid=58487)
  Read of size 8 at 0x7b20000f9e98 by thread T34:
  * #0 ASTextKitAttributes::operator==(ASTextKitAttributes const&) const ASTextKitAttributes.h:124 (AsyncDisplayKit:x86_64+0x5da4e8)
    #1 -[ASTextNodeRendererKey isEqual:] ASTextNode.mm:88 (AsyncDisplayKit:x86_64+0x5d9cff)
    #2 CFEqual <null>:8571360 (CoreFoundation:x86_64+0x1a33c)
    #3 -[ASTextNodeDrawParameter rendererForBounds:] ASTextNode.mm:155 (AsyncDisplayKit:x86_64+0x5db0d2)
    #4 +[ASTextNode drawRect:withParameters:isCancelled:isRasterizing:] ASTextNode.mm:509 (AsyncDisplayKit:x86_64+0x5e31dd)
    #5 __90-[ASDisplayNode(AsyncDisplay) _displayBlockWithAsynchronous:isCancelledBlock:rasterizing:]_block_invoke.74 ASDisplayNode+AsyncDisplay.mm:258 (AsyncDisplayKit:x86_64+0x1baa1d)
    #6 __71-[_ASAsyncTransaction addOperationWithBlock:priority:queue:completion:]_block_invoke _ASAsyncTransaction.mm:422 (AsyncDisplayKit:x86_64+0x1d95a)
    #7 invocation function for block in ASAsyncTransactionQueue::GroupImpl::schedule(long, NSObject<OS_dispatch_queue>*, void () block_pointer) _ASAsyncTransaction.mm:255 (AsyncDisplayKit:x86_64+0x13825)
    #8 __tsan::invoke_and_release_block(void*) <null>:8571360 (libclang_rt.tsan_iossim_dynamic.dylib:x86_64+0x637fb)
    #9 _dispatch_client_callout <null>:8571360 (libdispatch.dylib:x86_64+0x343b)

  Previous write of size 8 at 0x7b20000f9e98 by thread T32:
[...]

PR #830, opened 2018-03-11
Fix Text Node Thread Sanitizer Warning

Adlai-Holler:
I made this before I saw that #819 exists, and while I think @hebertialmeida's solution may possibly be faster (though it could lead to thread explosion), I think this solution is more precise.

The race condition is in the C++ struct `ASTextKitAttributes` itself. Since there is no synchronization around the contents of that struct, the sanitizer is correctly warning us.

This has led me down a heck of a rabbit-hole with memory safety. In lieu of a more general fix for the problem of creating very fast, but flexible and thread-safe C++ structures that interoperate well with Objective-C, this fix solves the thread sanitizer warnings in an appropriate way, by protecting access to that structure, at least in the `hash` and `isEqual` methods, which seems sufficient since the class is specifically termed a `Key`.

Note that it's not sufficient to make the Objective-C property accessor atomic, since that seems only to protect the pointer. I don't know much about how synthesized atomic accessors treat c++ objects, but I wouldn't be surprised if they're not actually atomic.

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -63,10 +63,13 @@ static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncation
 @property (assign, nonatomic) CGSize constrainedSize;
 @end
 
-@implementation ASTextNodeRendererKey
+@implementation ASTextNodeRendererKey {
+  std::mutex _m;
+}
 
 - (NSUInteger)hash
 {
+  std::lock_guard<std::mutex> _l(_m);
 #pragma clang diagnostic push
 #pragma clang diagnostic warning "-Wpadded"
   struct {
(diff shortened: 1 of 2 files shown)


Kaspik:
Problem reported in this issue https://github.com/TextureGroup/Texture/issues/552
This issue can be then probably closed. (I'll verify that after release)

Source/ASTextNode.mm
```diff
@@ -85,8 +88,13 @@ - (BOOL)isEqual:(ASTextNodeRendererKey *)object
   if (self == object) {
     return YES;
   }
+
+  // Lock both objects, avoiding deadlock.
+  std::lock(_m, object->_m);
```
garrettmoon: Unrelated to you changes, should there be class checks in here? It may not matter in since this is an internal object, but should we ensure object is an ASTextRendererKey for correctness?

Adlai-Holler: Good call. I added a comment explaining it.

Adlai-Holler:
Only un-CI'd changes are a comment I added. Landing!

PR #832, opened 2018-03-12
[ASTextNode2] Simplify Caching

Adlai-Holler:
This moves the bucketing behavior into the NSCache itself.

The cache which is currently `NSAttributedString -> ASTextCacheValue (a.k.a. [ASTextLayout])` becomes `ASTextCacheKey -> ASTextLayout` so that, instead of us doing the bucketing and scan, NSCache does it.

`ASTextCacheKey` is a pair of `(NSAttributedString, ASTextContainer)`, similar to `ASTextKitAttributes` in the ASTextNode1 world.

The fact is, a given pair of `(container, attributed string)` generates one layout, which can be used against multiple `(container, attributed string)` pairs, and this diff makes the `-hash` and `-isEqual` implementations reflect that.

### DIFF
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -310,7 +310,7 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   container.size = constrainedSize;
   [self _ensureTruncationText];
   
-  NSMutableAttributedString *mutableText = [attributedText mutableCopy];
+  NSMutableAttributedString *mutableText = [attributedText mutableCopy] ?: [[NSMutableAttributedString alloc] init];
   [self prepareAttributedString:mutableText];
   ASTextLayout *layout = [ASTextNode2 compatibleLayoutWithContainer:container text:mutableText];
   
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -18,6 +18,7 @@
 
 #import <AsyncDisplayKit/_ASDisplayLayer.h>
 #import <AsyncDisplayKit/ASDisplayNode+FrameworkSubclasses.h>
+#import <AsyncDisplayKit/ASHashing.h>
 #import <AsyncDisplayKit/ASHighlightOverlayLayer.h>
 #import <AsyncDisplayKit/ASDisplayNodeExtras.h>
 
@@ -31,13 +32,91 @@
 #import <AsyncDisplayKit/ASObjectDescriptionHelpers.h>
 #import <AsyncDisplayKit/ASTextLayout.h>
 
-@interface ASTextCacheValue : NSObject {
-  @package
-  ASDN::Mutex _m;
-  std::deque<std::tuple<CGSize, ASTextLayout *>> _layouts;
+@interface ASTextCacheKey : NSObject {
+  NSAttributedString *_attributedString;
+  ASTextContainer *_container;
+  
+  // nil if we don't have a layout yet.
+  // non-null for entries stored in the cache.
+  ASTextLayout *_layout;
 }
+
+- (instancetype)initWithContainer:(ASTextContainer *)container attributedString:(NSAttributedString *)attributedString;
 @end
-@implementation ASTextCacheValue
+
+@implementation ASTextCacheKey
+
+- (instancetype)initWithContainer:(ASTextContainer *)container attributedString:(NSAttributedString *)attributedString
+{
+  if (self = [super init]) {
+    _container = container;
+    _attributedString = [attributedString copy];
+  }
+  return self;
+}
+
+- (NSUInteger)hash
+{
+  // Don't include size in hash. Size -> layout mapping is many-to-one (fuzzy).
+#pragma clang diagnostic push
+#pragma clang diagnostic warning "-Wpadded"
+  struct {
+    size_t attributedStringHash;
+    size_t containerHash;
+#pragma clang diagnostic pop
+  } data = {
+    [_attributedString hash],
+    [_container hash]
+  };
+  return ASHashBytes(&data, sizeof(data));
+}
+
+- (BOOL)isEqual:(id)object
+{
+  ASTextCacheKey *otherKey = ASDynamicCast(object, ASTextCacheKey);
+  if (!otherKey) {
+    return NO;
+  }
+  
+  NSCAssert(!(_layout && otherKey->_layout), @"Should only compare incomplete layout with complete layout!");
+  ASTextLayout *layout = _layout ?: otherKey->_layout;
+  
+  CGRect containerBounds = (CGRect){ .size = _container.size };
+  CGSize layoutSize = layout.textBoundingSize;
+  // 1. CoreText can return frames that are narrower than the constrained width, for obvious reasons.
+  // 2. CoreText can return frames that are slightly wider than the constrained width, for some reason.
+  //    We have to trust that somehow it's OK to try and draw within our size constraint, despite the return value.
+  // 3. Thus, those two values (constrained width & returned width) form a range, where
+  //    intermediate values in that range will be snapped. Thus, we can use a given layout as long as our
+  //    width is in that range, between the min and max of those two values.
+  CGRect minRect = CGRectMake(0, 0, MIN(layoutSize.width, containerBounds.size.width), MIN(layoutSize.height, containerBounds.size.height));
+  if (!CGRectContainsRect(containerBounds, minRect)) {
+    return NO;
+  }
+  CGRect maxRect = CGRectMake(0, 0, MAX(layoutSize.width, containerBounds.size.width), MAX(layoutSize.height, containerBounds.size.height));
+  if (!CGRectContainsRect(maxRect, containerBounds)) {
+    return NO;
+  }
+  
+  // Now check container params.
+  ASTextContainer *otherContainer = layout.container;
+  if (!UIEdgeInsetsEqualToEdgeInsets(_container.insets, otherContainer.insets)) {
+    return NO;
+  }
+  if (!ASObjectIsEqual(_container.exclusionPaths, otherContainer.exclusionPaths)) {
+    return NO;
+  }
+  if (_container.maximumNumberOfRows != otherContainer.maximumNumberOfRows) {
+    return NO;
+  }
+  if (_container.truncationType != otherContainer.truncationType) {
+    return NO;
+  }
+  if (!ASObjectIsEqual(_container.truncationToken, otherContainer.truncationToken)) {
+    return NO;
+  }
+  return YES;
+}
 @end
 
 /**
@@ -380,81 +459,21 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 {
   // Allocate layoutCacheLock on the heap to prevent destruction at app exit (https://github.com/TextureGroup/Texture/issues/136)
   static ASDN::StaticMutex& layoutCacheLock = *new ASDN::StaticMutex;
-  static NSCache<NSAttributedString *, ASTextCacheValue *> *textLayoutCache;
+  static NSCache<ASTextCacheKey *, ASTextLayout *> *textLayoutCache;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
     textLayoutCache = [[NSCache alloc] init];
   });
+
   
-  ASTextCacheValue *cacheValue = ({
-    ASDN::StaticMutexLocker lock(layoutCacheLock);
-    cacheValue = [textLayoutCache objectForKey:text];
-    if (cacheValue == nil) {
-      cacheValue = [[ASTextCacheValue alloc] init];
-      [textLayoutCache setObject:cacheValue forKey:text];
-    }
-    cacheValue;
-  });
+  ASTextCacheKey *key = [[ASTextCacheKey alloc] initWithContainer:container attributedString:text];
   
-  CGRect containerBounds = (CGRect){ .size = container.size };
-  {
-    ASDN::MutexLocker lock(cacheValue->_m);
-    for (auto &t : cacheValue->_layouts) {
-      CGSize constrainedSize = std::get<0>(t);
-      ASTextLayout *layout = std::get<1>(t);
-      
-      CGSize layoutSize = layout.textBoundingSize;
-      // 1. CoreText can return frames that are narrower than the constrained width, for obvious reasons.
-      // 2. CoreText can return frames that are slightly wider than the constrained width, for some reason.
-      //    We have to trust that somehow it's OK to try and draw within our size constraint, despite the return value.
-      // 3. Thus, those two values (constrained width & returned width) form a range, where
-      //    intermediate values in that range will be snapped. Thus, we can use a given layout as long as our
-      //    width is in that range, between the min and max of those two values.
-      CGRect minRect = CGRectMake(0, 0, MIN(layoutSize.width, constrainedSize.width), MIN(layoutSize.height, constrainedSize.height));
-      if (!CGRectContainsRect(containerBounds, minRect)) {
-        continue;
-      }
-      CGRect maxRect = CGRectMake(0, 0, MAX(layoutSize.width, constrainedSize.width), MAX(layoutSize.height, constrainedSize.height));
-      if (!CGRectContainsRect(maxRect, containerBounds)) {
-        continue;
-      }
-      
-      // Now check container params.
-      ASTextContainer *otherContainer = layout.container;
-      if (!UIEdgeInsetsEqualToEdgeInsets(container.insets, otherContainer.insets)) {
-        continue;
-      }
-      if (!ASObjectIsEqual(container.exclusionPaths, otherContainer.exclusionPaths)) {
-        continue;
-      }
-      if (container.maximumNumberOfRows != otherContainer.maximumNumberOfRows) {
-        continue;
-      }
-      if (container.truncationType != otherContainer.truncationType) {
-        continue;
-      }
-      if (!ASObjectIsEqual(container.truncationToken, otherContainer.truncationToken)) {
-        continue;
-      }
-      // TODO: When we get a cache hit, move this entry to the front (LRU).
-      return layout;
-    }
+  ASDN::StaticMutexLocker lock(layoutCacheLock);
+  ASTextLayout *layout = [textLayoutCache objectForKey:key];
+  if (!layout) {
+    layout = [ASTextLayout layoutWithContainer:container text:text];
+    [textLayoutCache setObject:layout forKey:key];
   }
-  
-  // Cache Miss.
-  
-  // Compute the text layout.
-  ASTextLayout *layout = [ASTextLayout layoutWithContainer:container text:text];
-  
-  // Store the result in the cache.
-  {
-    ASDN::MutexLocker lock(cacheValue->_m);
-    cacheValue->_layouts.push_front(std::make_tuple(container.size, layout));
-    if (cacheValue->_layouts.size() > 3) {
-      cacheValue->_layouts.pop_back();
-    }
-  }
-  
   return layout;
 }
 
--- a/Source/Private/TextExperiment/Component/ASTextLayout.m
+++ b/Source/Private/TextExperiment/Component/ASTextLayout.m
@@ -13,7 +13,6 @@
 #import <AsyncDisplayKit/ASTextUtilities.h>
 #import <AsyncDisplayKit/ASTextAttribute.h>
 #import <AsyncDisplayKit/NSAttributedString+ASText.h>
-#import <AsyncDisplayKit/ASHashing.h>
 #import <AsyncDisplayKit/ASInternalHelpers.h>
 
 const CGSize ASTextContainerMaxSize = (CGSize){0x100000, 0x100000};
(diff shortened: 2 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextNode2.mm:</p>
<pre><code>//
//  ASTextNode2.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextNode2.mm:</p>
<pre><code>//
//  ASTextNode2.mm
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b8cb63a3db299bae591897a509724e75/raw/f79ba6fb70e4a4d19d1c78450518e5a681b77a2c/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/283d8dbaf782714a132f6956f8b13c52/raw/b272ab52abea281c996b40cfca78e55c7c4a8539/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.h:</p>
<pre><code>//
//  ASTextLayout.h
//  Texture
//
//
//
//
</code></</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASTextLayout.m:</p>
<pre><code>//
//  ASTextLayout.m
//  Texture
//
//
//
//
</code></</td>
    </tr>
  </tbody>
</table>

</p>

Adlai-Holler:
⚠️ This diff turned out pretty huge! I'll decompose it into a few smaller diffs.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/cfbdf3c9cb337d827972650f2ed0e056/raw/a4d82e9c54433a408b26df63e83a37a1ee29ce5f/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b5989a5798404e71f20c817ba31ccf70/raw/68d7aa84f96026abcc87a7f9f45687eed45ac323/log.txt)

Adlai-Holler:
Failure is spurious. Need to kick it.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASTextNode.mm
```diff
@@ -110,6 +110,7 @@ - (BOOL)isEqual:(ASTextNodeRendererKey *)object
  dispatch_once(&onceToken, ^{
    __rendererCache = [[NSCache alloc] init];
    __rendererCache.countLimit = 500; // 500 renders cache
```
appleguy: I know this is for ASTextNode1...however, this might actually be quite a bit of memory. Do we dump this cache on memory warning? Is there anything simple we can do to make it more responsive to some combination of memory consumption (by objects in the cache), device memory size, or frequency of memory pressure?

I think there's a need for something like ASMemoryManager (or MemoryOverlord lol) that provides some conveniences around memory. Total memory capacity could be bucketed into enums with perhaps a low/med/high — especially iPads would mostly be in low or med, for example. Most valuably would be registering some blocks or observers to call upon memory pressure, and providing an indication of how severe or sustained the memory pressure is.

For example, repeated memory pressure should certainly reduce the countLimit, display range size, or other such parameters rather than just one-time cache dumps. Even if there's no mechanism to increase these again, this would allow us to choose more aggressive defaults without as much concern for apps that may have e.g. many large text blocks.

appleguy: Filed a tracking task for this idea: https://github.com/TextureGroup/Texture/issues/927

Adlai-Holler: A memory management object is really interesting. Take a look at this, which provides 3 levels of severity. This is the hook that NSCache uses. Side note, I'm in favor of removing the explicit count limit here and letting NSCache decide.

https://developer.apple.com/documentation/dispatch/dispatch_source_memory_pressure_event_flags?language=objc

Source/ASTextNode2.mm
```diff
@@ -78,6 +70,12 @@ @implementation ASTextNode2 {
   NSAttributedString *_composedTruncationText;
   NSArray<NSNumber *> *_pointSizeScaleFactors;
   
+  // If nil, regenerate.
+  NSAttributedString *_preparedAttributedText;
+  
+  // Just a fast cache. May not be valid but good place to check first.
```
appleguy: Could you clarify this comment a bit, specifically what about it may not be valid (like if constrainedSize doesn't match?)

Same for "If nil, regenerate" -- could remove this comment, or make it a bit more clear how the variable is used and expectations for how it is maintained.

Source/ASTextNode2.mm
```diff
@@ -30,18 +29,11 @@
...
-  @package
-  ASDN::Mutex _m;
-  std::deque<std::tuple<CGSize, ASTextLayout *>> _layouts;
-}
-@end
-@implementation ASTextCacheValue
-@end
-
 /**
  * If set, we will record all values set to attributedText into an array
```
appleguy: It would be great to use this feature to collect a bunch of data and add it to our unit tests! This might collect some noise when iOS changes, but it would also be an excellent way to identify breakage / race conditions from core text classes changing (e.g. if we had an auto-scrolling collection showing ASTextCellNodes, measuring them in batches of 20 insertions to the data source).

Adlai-Holler: Yes we've actually got some – `AttributedStringsFixture0.plist` that I captured from Pinterest a while back. Just need to build a test that uses them.

Source/ASTextNode2.mm
```diff
@@ -228,18 +226,16 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
...
   container.size = constrainedSize;
-  [self _ensureTruncationText];
   
-  NSMutableAttributedString *mutableText = [attributedText mutableCopy];
-  [self prepareAttributedString:mutableText];
-  ASTextLayout *layout = [ASTextNode2 compatibleLayoutWithContainer:container text:mutableText];
+  ASTextLayout *layout = [self compatibleLayoutWithContainer:container text:[self l_preparedAttributedText]];
   
+  // Is this necessary?
   [self setNeedsDisplay];
```
appleguy: Hmm, I think you should remove this (as long as the tests pass, or fix them if the tests are not too hard to fix).

As long as we have the layer's .needsDisplayOnBoundsChange set to YES, then this should not be a factor.

One could argue that we should use default-synchronous display whenever the layout changes while interfaceState is Visible. That would fix some of the stretching texture issues that often happen when changing a string and it resizes the frame. However, this would require something like setNeedsSynchronousDisplay, and it would be called in -layout instead of here.

Source/ASTextNode2.mm
```diff
@@ -323,10 +320,17 @@ - (NSArray *)exclusionPaths
   return _textContainer.exclusionPaths;
 }
 
-- (void)prepareAttributedString:(NSMutableAttributedString *)attributedString
+- (NSAttributedString *)l_preparedAttributedText
 {
-  ASLockScopeSelf();
```
appleguy: Pretty sure the lock should stay here!

Source/ASTextNode2.mm
```diff
@@ -323,10 +320,17 @@ - (NSArray *)exclusionPaths
   return _textContainer.exclusionPaths;
```
appleguy: In the method above this (won't let me comment on it), setExclusionPaths:...is it possible for _textContainer to not be allocated yet? Should we call self.textContainer or something?

Source/ASTextNode2.mm
Adlai-Holler: Oops, meant to replace that with a lock assertion. I used the `l_` prefix to indicate the method is called while the lock is held. That may be too easy to miss, so I could change it to `locked_` although locked assertions will probably prevent people from missing it in reality (unit tests).

Source/ASTextNode2.mm
```diff
@@ -323,10 +320,17 @@ - (NSArray *)exclusionPaths
...
 
-- (void)prepareAttributedString:(NSMutableAttributedString *)attributedString
+- (NSAttributedString *)l_preparedAttributedText
 {
-  ASLockScopeSelf();
+  if (_preparedAttributedText) {
+    return _preparedAttributedText;
+  }
+  
+  if (!_attributedText) {
```
appleguy: Move this condition above the other one, and instead of early return, just do _preparedAttributedText = [[NSAttributedString alloc] init];

Then it will still early return, and _preparedAttributedText will be set.

Source/ASTextNode2.mm
```diff
@@ -350,6 +354,8 @@ - (void)prepareAttributedString:(NSMutableAttributedString *)attributedString
     shadow.shadowBlurRadius = _shadowRadius;
     [attributedString addAttribute:NSShadowAttributeName value:shadow range:NSMakeRange(0, attributedString.length)];
   }
+  _preparedAttributedText = [attributedString copy];
```
appleguy: This second copy is not necessary, right? Since this code is called so much, I'd suggest we just let the mutable object live on, as anyone up-casting it can crash as much as they want :).

Source/ASTextNode2.mm
```diff
@@ -323,10 +320,17 @@ - (NSArray *)exclusionPaths
...
-  ASLockScopeSelf();
+  if (_preparedAttributedText) {
+    return _preparedAttributedText;
+  }
+  
+  if (!_attributedText) {
+    return [[NSAttributedString alloc] init];
+  }
  
+  NSMutableAttributedString *attributedString = [_attributedText mutableCopy];
```
appleguy: Why was there not a copy here before? Is it definitely needed here, or should we be doing it at a higher level like setAttributedString?

Source/ASTextNode2.mm
```diff
@@ -360,115 +366,65 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
...
-    @"container": copiedContainer,
-    @"text": mutableText,
-    @"bgColor": self.backgroundColor ?: [NSNull null]
-  };
+           @"container": copiedContainer,
+           @"text": [self l_preparedAttributedText],
+           @"bgColor": self.backgroundColor ?: [NSNull null],
+           // Pass along self as a hack so that we can update _lastUsedLayout.
+           // We'll be careful
+           @"instance" : self
```
appleguy: Hmm, lol. Should we use ASWeakProxy, or is that guaranteed to not make a difference in when we're deallocated?

Since it is possible to have two displays going on at the same time (the sentinel in ASDisplayNode tracks which one should win), I'm not sure it is safe to set _lastUsedLayout unless we're checking which display is intended to / applied as the winner?

Source/ASTextNode2.mm
```diff
@@ -360,115 +366,65 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
...
+           };
 }
 
 /**
  * If it can't find a compatible layout, this method creates one.
  *
- * NOTE: Be careful to copy `text` if needed.
+ * NOTE: The `container` you pass into this should be a copy, as it may be
+ * retained by the cache in the event of a cache miss.
```
appleguy: Can't we just copy the container in the case of a cache miss? That would be a huge win in some workflows, if that's the only reason / case for the copy.

Source/ASTextNode2.mm
Adlai-Holler: No, the textContainer is created in -init and kept forever. 

A good way to think of it is the text container actually IS the text node in raw form. It stores many of the nodes properties and it's nicely copyable with atomic accessors. The text node itself is a bridge between the container and the rest of ASDK.

Source/ASTextNode2.mm
```diff
@@ -360,115 +366,65 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
...
-
 {
-  // Allocate layoutCacheLock on the heap to prevent destruction at app exit (https://github.com/TextureGroup/Texture/issues/136)
-  static ASDN::StaticMutex& layoutCacheLock = *new ASDN::StaticMutex;
-  static NSCache<NSAttributedString *, ASTextCacheValue *> *textLayoutCache;
+  // First quick check – if the last layout we computed is useable.
+  // On a cache miss, this method is going to run long and in some
+  // scenarios it may make sense to run concurrently with different
+  // sizes, so don't hold the lock the whole time.
+  auto lastLayout = ASLockedSelf(_lastUsedLayout);
```
appleguy: Doesn't the caller need to have self locked this whole time? Otherwise, it will need to lock after it receives the layout, and re-check that it is still valid.

Source/ASTextNode2.mm
Adlai-Holler: So my thing in cases like this, is if I think there's even one downstream copy that is likely, just go ahead and copy now to make that free, to avoid multiple copies of the same object, and to save my mental energy. This method will get called once per node per configuration (change of properties with a layout/render pass in between). So in one data controller batch, this should be called realistically <100 times (say, 25 items with 4 text nodes per).

In this case for example, the attributed string will get copied when we make an ASTextNodeCacheKey from it anyway. And in fact, it will get copied each layout/render pass, so we will do more copying if we don't change properties in between those passes. We could remove THAT copy too, but then we're putting a little crack in our abstractions. These copies will be fast, they will mostly happen off-main, and there will only be ~1 per text node in common cases so I think we should leave it.

Source/ASTextNode2.mm
Adlai-Holler: Yeah it's definitely possible for the loser to be set – but `lastUsedLayout` is not considered authoritative. As the comment says, it's just a fast-path for the common case – a first place to check before going to the global layout cache. If we miss, it's not a tragedy.

Source/ASTextNode2.mm
Adlai-Holler: No, because we're inside a method with immutable parameters. The validity of the layout can't change out from under us – whatever text and container they passed in, that's what we need to respond with. It's up to the consumer to worry about whether those parameters remain valid throughout the method.

I really hated to turn this from a class method to an instance method because it's so unclear that this method simply does not rely on any instance state. It asks the instance for a quick candidate layout to consult first, and it puts the result into that bucket for the next call.

For instance, the layout system calls this with the lock held for the reason you mentioned. The display system is operating on fixed draw params, so it doesn't need to lock the node to call this.

Source/ASTextNode2.mm
Adlai-Holler: The problem is that currently the callers of this method already have to create a copy, in order to change the container size. So the copy is free. If I copy inside this method, we create two copies. Does that make sense?

nguyenhuy:
@Adlai-Holler Can you follow up on this diff?

PR #842, opened 2018-03-18
Renew supplementary node on relayout

wsdwsd0829:
When a supplementary view experiences layout change from zero to non-zero, the map in dataController need to be updated.

Thanks @nguyenhuy to provide the hints here:)
 https://github.com/nguyenhuy/Texture/commit/478d68ecb091d1b5adb76bf5eb8915987486c292


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/7b33dcd3948ad80fd2e42800eba16a04/raw/e0c5aa0e7918874004609992c166b774e70c5abb/log.txt)

Source/Private/ASMutableElementMap.m
```diff
@@ -68,6 +68,17 @@ - (void)removeSectionsAtIndexes:(NSIndexSet *)indexes
...
 
+- (void)removeSupplementaryElementsAtIndexPaths:(NSArray<NSIndexPath *> *)indexPaths kind:(NSString *)kind
+{
+  NSMutableDictionary *supplementariesForKind = _supplementaryElements[kind];
+  if (supplementariesForKind == nil) {
+    return;
+  }
+  for (NSIndexPath *indexPath in indexPaths) {
+    supplementariesForKind[indexPath] = nil;
+  }
```
Adlai-Holler: Use `[_supplementaryElements[kind] removeObjectsForKeys:indexPaths];`

Source/Details/ASDataController.mm
```diff
@@ -281,6 +282,56 @@ - (void)_repopulateSupplementaryNodesIntoMap:(ASMutableElementMap *)map
...
+      NSMutableArray<NSIndexPath *> *indexPathsToInsertForKind = [[NSMutableArray alloc] init];
+      // If supplementary node does exist and size is now zero, remove it.
+      // If supplementary node doesn't exist and size is now non-zero, insert one.
+      for (NSIndexPath *indexPath in indexPaths) {
+        ASCollectionElement *previousElement = [previousMap supplementaryElementOfKind:kind atIndexPath:indexPath];
+        if (canDelegate) {
+          ASCollectionLayoutContext *context = [self.layoutDelegate layoutContextWithElements:previousMap];
+          Class<ASDataControllerLayoutDelegate> layoutDelegateClass = [self.layoutDelegate class];
+          ASCollectionLayoutState *layoutState = [layoutDelegateClass calculateLayoutWithContext:context];
+          UICollectionViewLayoutAttributes *attrs = [layoutState layoutAttributesForSupplementaryElementOfKind:kind atIndexPath:indexPath];
```
Adlai-Holler: Let's not call `calculateLayoutWithContext:` in a loop on main like this. 

These methods should only gather data and not perform expensive operations like layout. The problem I guess is that the ASCollectionLayout API is intentionally opaque about what size ranges it uses to measure its nodes.

@nguyenhuy Maybe we need new API to let ASCollectionLayoutDelegates specify what supplementary items exist during this phase, so that we can put them in the map.

nguyenhuy: I agree that we shouldn't perform a layout here just to figure out the size range. This is a limitation of ASCollectionLayout API.

Given the limited usage of ASCollectionLayout at the moment, and the possibility that it would take some time to come up with a sensible extension for its API, I'm wondering if we should aim to merge this PR without focusing much on ASCollectionLayout support, and then create a new ticket to follow up?

Source/Private/ASMutableElementMap.m
nguyenhuy: 👍

Source/Details/ASDataController.mm
wsdwsd0829: ok,  I will just leave this block as todo.

Source/Details/ASDataController.mm
```diff
@@ -281,6 +282,52 @@ - (void)_repopulateSupplementaryNodesIntoMap:(ASMutableElementMap *)map
...
+    ASSizeRange newSizeRange;
+    for (NSString *kind in [self supplementaryKindsInSections:sectionIndexes]) {
+      NSArray<NSIndexPath *> *indexPaths = [self _allIndexPathsForItemsOfKind:kind inSections:sectionIndexes];
+      NSMutableArray<NSIndexPath *> *indexPathsToDeleteForKind = [[NSMutableArray alloc] init];
+      NSMutableArray<NSIndexPath *> *indexPathsToInsertForKind = [[NSMutableArray alloc] init];
+      // If supplementary node does exist and size is now zero, remove it.
+      // If supplementary node doesn't exist and size is now non-zero, insert one.
+      for (NSIndexPath *indexPath in indexPaths) {
+        ASCollectionElement *previousElement = [previousMap supplementaryElementOfKind:kind atIndexPath:indexPath];
+        if (canDelegate) {
```
wsdwsd0829: do we need a size range here? If so Zero? does not sound quite right though

nguyenhuy: Yeah, doesn't sound good to me neither. I think we can just completely ignore this case and bail early.

wsdwsd0829: oh right.

Adlai-Holler:
Thanks for the diff Max!

PR #851, opened 2018-03-24
Make objects conform to NSLocking

Adlai-Holler:
- Display node, layout spec, element style now conform to `NSLocking`. They use the same `__instanceLock__` as before, but they can be publicly locked and unlocked.
- Subclasses (both inside and outside the framework) can use the existing lock, rather than adding their own locks.
- Users can lock/unlock nodes to group changes.
- `ASThread.h` (public) has useful C++-esque locking macros.
- Message-sends are not the end of the world. The readability and safety benefits here are worth the trivial amount of overhead.
- NSLocking is nice because we can retain the lock object, avoiding crashes we had before where an object died while someone else held its lock.
- New macros:
  - `ASLockScope(id<NSLocking> l)` retains and adds the lock to the current scope.
  - `ASLockScopeSelf` locks self for the current scope. Skips retain (self is safe.)
  - `ASLocked(id<NSLocking> l, expr)` Evaluates to the `expr` while holding the lock.
  - `ASLockedSelf(expr)` Evaluates to the `expr` while using self as the lock (skip retain).
  - `ASCompareAssign(lvalue, newValue)` compares the two values, updating the lvalue if different, returns BOOL.
  - `ASLockedSelfCompareAssign(lvalue)` Same as above but holds self as lock.
  - Other variants of the same things.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e25fab7d258cdea4f94e4191dfceca8f/raw/ac694572e7ce7924ae59b868a90d75cea327297c/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/3b04b9a3cfffdc94d11bc5e0a7ecd11c/raw/39d5f239a4f24aaf2915dd57e74b7cff9c9b515d/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/56418dfba2fad426f3e3ddca5bb7c4b0/raw/553e606f7528a77fa07d7bc2f3041915cc45ad79/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f008970fa05013dbe27b9057bcaec0d1/raw/8dc8616ca2c9deac11571de829b44683c5c38c02/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/13e78d55c9bed1e28eaf9ecd91da9cb0/raw/2c494433840c0340b4e727bc7582a4ef14554b64/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/578382323b8d628cecb305f9a9720314/raw/bdc8ce91781df8c65523f73721416b0625958718/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASNetworkImageNode.mm
```diff
@@ -29,19 +30,21 @@
...
 {
-  // Only access any of these with __instanceLock__.
+  // Only access any of these while locked.
   __weak id<ASNetworkImageNodeDelegate> _delegate;
 
   NSURL *_URL;
   UIImage *_defaultImage;
 
-  NSUUID *_cacheUUID;
+  std::atomic_uint _cacheToken;
```
maicki: We may should have this change in a follow up diff. Not really 100% related to adopting NSLocking.

Source/Private/ASDisplayNode+FrameworkSubclasses.h
```diff
@@ -1,40 +0,0 @@
-//
-//  ASDisplayNode+FrameworkSubclasses.h
```
maicki: Nice that we can get rid of that

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASImageNode.mm
```diff
@@ -305,11 +305,11 @@ - (void)setPlaceholderColor:(UIColor *)placeholderColor
 
 - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
 {
-  ASDN::MutexLocker l(__instanceLock__);
+  ASLockScopeSelf();
   
   ASImageNodeDrawParameters *drawParameters = [[ASImageNodeDrawParameters alloc] init];
   drawParameters->_image = [self _locked_Image];
-  drawParameters->_bounds = [self threadSafeBounds];
+  drawParameters->_bounds = self.bounds;
```
appleguy: Is this particular change necessary, and also safe? IIRC it was important to use threadSafeBounds here.

Source/ASNetworkImageNode.mm
```diff
@@ -606,13 +593,13 @@ - (void)_downloadImageWithCompletion:(void (^)(id <ASImageContainerProtocol> ima
...
 {
-  __instanceLock__.lock();
+  [self lock];
     __weak id<ASNetworkImageNodeDelegate> delegate = _delegate;
     BOOL delegateDidStartFetchingData = _delegateFlags.delegateDidStartFetchingData;
     BOOL isImageLoaded = _imageLoaded;
     NSURL *URL = _URL;
     id currentDownloadIdentifier = _downloadIdentifier;
-  __instanceLock__.unlock();
+  [self unlock];
```
appleguy: The difference in style between these statements and the macros feels a bit less prominent / conspicuous than before, where all locks involved __instanceLock__.

I wonder if we should use a macro for even the simple lock / unlock message send, just to make locking operations consistent? Even if we did that, lock / unlock would still be available, especially for third party code.

Source/ASVideoNode.mm
```diff
@@ -650,33 +649,29 @@ - (void)setMuted:(BOOL)muted
...
 
   if (_playerNode == nil) {
     _playerNode = [self constructPlayerNode];
 
-    __instanceLock__.unlock();
-      [self addSubnode:_playerNode];
-    __instanceLock__.lock();
-      
+    ASUnlockScope(self);
+    [self addSubnode:_playerNode];
```
appleguy: I'm not sure if it is important to re-lock before the setNeedsLayout, but the locking scope did change here (may want to add a { } scope around the addSubnode)

Source/Base/ASBaseDefines.h
```diff
@@ -232,6 +232,26 @@
   ((c *) ([__val class] == [c class] ? __val : nil));\
 })
 
+// Compare two primitives, assign if different. Returns whether the assignment happened.
+#define ASCompareAssign(lvalue, newValue) ({  \
```
appleguy: Shouldn't there be a variant that does use ASObjectIsEqual, but doesn't do a copy (just a regular assignment which may involve a retain?)

Is there any way to enforce that ASCompareAssign is *not* used for objects accidentally?

Source/Details/ASThread.h
```diff
@@ -31,6 +29,60 @@ static inline BOOL ASDisplayNodeThreadIsMain()
   return 0 != pthread_main_np();
 }
 
+/**
+ * Adds the lock to the current scope.
+ *
+ * A C version of the C++ lockers. Pass in any id<NSLocking>.
+ * One benefit this has over C++ lockers is that the lock is retained. We
+ * had bugs in the past where an object would be deallocated while someone
```
appleguy: I recall there was a solution to this issue that prevented deallocation when the lock was owned. Is this saying the problem was not fully solved, even in ASDisplayNode?

My main concern with this patch is of course not the message sends themselves, but the retain / release traffic if we were to use these facilities in core codepaths would be quite significant - and almost always unnecessary. That is to say, only one reference to an object is needed for it to be safe, and in most codepaths there does exist such a reference.

Since we're mainly changing the subclasses instead of ASDisplayNode, that should be OK. That said, the optimizations in ASNetworkImageNode were developed through profile-driven improvements. It may be appropriate to use the Unowned variant in core code such as ASNetworkImageNode where there is a known reference, e.g. if we have a keepalive from our view object which is still in a hierarchy then there is really no reason to retain here (and we could instead lock the object before clearing that keepalive).

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Base/ASBaseDefines.h
Adlai-Holler: Good point, I added that variant as `ASCompareAssignObject`.

I think there's value in using plain ASCompareAssign for objects, in the case where we want to skip the `isEqual:` (say, in `setDelegate:` or similar), so I'd rather keep that option open.

Source/ASVideoNode.mm
Adlai-Holler: Good catch, either way it's outside the scope of this diff 👍

Source/ASImageNode.mm
Adlai-Holler: Oh I didn't know. When I removed the FrameworkSubclasses import, the import for this field was lost and I figured it was OK to use bounds since this method is (allegedly) called on main. I put it back since the change is outside the purview of this diff 👍

Source/ASNetworkImageNode.mm
Adlai-Holler: It's true – I'm iffy on this. I totally see your point that there's a mismatch in look and feel, but I don't like when there's two ways to do the same thing, and I like the referential transparency.

I also kind of like the mismatch, since it "discourages" use of these methods by making them stick out.

Let's sleep on this. There will be more locking diffs in the future and we can add the macros easily at any time.

Issue #853, opened 2018-03-25
Interface state tests fail when transaction queue is disabled


Adlai-Holler:
This is a test run with the interface state coalescing disabled:

https://gist.githubusercontent.com/pinterest-ios/f9d888dad32cb905d560a67b03f45987/raw/6de6a9b4ede5484a7d703d698678360abb6350a3/log.txt

We need to make the coalescing opt-in and also address these issues. cc @appleguy

Issue #845, opened 2018-03-21
[ASPagerNode] Layout bug introduced by 2618c5 #788


flovouin:
Hi guys,

While upgrading Texture in my App, I encountered a layout bug about a month ago. I promised a sample project but then forgot about it. At last, here it is: https://github.com/flovouin/Texture-PagerNodeLayoutBug

It occurs when using the `ASPagerNode` full screen and... well that's it. There's not much more to it: the cells do not take the full pager node frame, and leave an inset at the top.

The bug has been around since `2618c5` and still happens on `master`. The same behaviour (which might be intentional in this case?) can also be observed in previous versions by setting `allowsAutomaticInsetsAdjustment` to `true` in the `ASPagerNode`.

As always, I can provide more details on request. :-)

Cheers,

Flo

nguyenhuy:
Thanks for helping to test `master` and reporting issues. The new run loop queue introduced in `2618c5` has been known to be a bit unstable at the moment. We had to disable it in the Pinterest app. Now that we have a new framework-level config/experiment functionality, we'll probably make an opt-in option for using the new queue.

appleguy:
@flovouin for the short term, try syncing past https://github.com/TextureGroup/Texture/pull/862 and make sure all is well!

Thanks a lot for reporting, and especially the repro case. @wsdwsd0829 and I will look into it.

flovouin:
Thanks @nguyenhuy and @appleguy for the replies!
Looks like #862 solves the problem for both the sample project and my app.
I'll reopen this if I notice something strange again.

Cheers,

Flo

PR #861, opened 2018-03-28
Revert #788 – Interface State Runloop Queue

Adlai-Holler:
Unfortunately, diff #788 has a few issues that need to be resolved before we can re-land it.

- When the run loop queue is disabled, the unit tests need to pass (#853). It appears that the diff affected the normal behavior of the framework, and Pinterest and other users (#845) have seen issues when the feature turned on.
- The experimental feature should be opt-in, not opt-out. We now have ASConfiguration to make this easy.

This reverts the diff until we can get these issues resolved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 - [ASDisplayNode] Add unit tests for layout z-order changes (with an open issue to fix).
 - [ASWrapperCellNode] Introduce a new class allowing more control of UIKit passthrough cells.
 - [ASDisplayNode] Consolidate main thread initialization and allow apps to invoke it manually instead of +load.
-- [ASRunloopQueue] Introduce new runloop queue(ASCATransactionQueue) to coalesce Interface state update calls for view controller transitions.
 - [ASRangeController] Fix stability of "minimum" rangeMode if the app has more than one layout before scrolling.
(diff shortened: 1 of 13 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

appleguy:
@Adlai-Holler thanks for putting this up — it sounds like there is a clear need to disable this by default. 

To that end, I've asked my team to make this a P0 and we'll either submit a PR today that accomplishes this (with all the coalescing code bypassed by default, and original timing unchanged) — or alternatively we can land this revert if the path to disable has any complications.

I'll let you guys know within a few hours what we plan to do. Sorry to hear this has created some issues!

Adlai-Holler:
OK @appleguy that's super reasonable IMO. To disable the coalescing, modify the `+ (ASConfiguration *)defaultConfiguration` method in `ASConfigurationInternal.m` and remove the Coalescing flag there.

Adlai-Holler:
Superseded by #862

appleguy:
Thanks for being on top of this - glad we could resolve quickly with a minimum of risk on the other PR.

Issue #838, opened 2018-03-16
width calculation error after setting flexBasis and flexWrap attribute


aimobier:
## AsyncDisplayKit some Waring

Width calculation error after setting flexBasis attribute

I don’t know if I’m dealing with this issue

When I set flexWrap using ASStackLayoutSpec I found this problem. In the end I found out that because of the calculation, the width was out of range. This main problem appears in the following code
[ASInternalHelpers  ASCeilPixelValue](https://github.com/TextureGroup/Texture/blob/master/Source/Private/ASInternalHelpers.m#L196)
````objective-c
CGFloat ASCeilPixelValue(CGFloat f)
{
  CGFloat scale = ASScreenScale();
  return ceil(f * scale) / scale;
}
````
For example
in iPhoneX, the width is 375 and flexBasis is set to 0.5. The scale is 3.
The result is 186.666666666666....
Caused the width to exceed 375

## Reproduce this problem

[AsyncCeilWidthError](https://github.com/aimobierInformal/AsyncCeilWidthError)

**Comment by rcancro** [comment]
I noticed something similar when creating an `ASStackLayoutSpec`. The stack's computed height comes back as `147.66666666666669`. This value is then place in an `ASLayout` which rounds the size vial `ASCeilPixelValue`. On a 3x device does the the following:

`147.666666669 * 3 = 443.0000000000000697`
`ceil(443.0000000000000697) = 444`
`444/3 = 148`

If there wasn't any precision error on the original stack layout's height, this wouldn't happen:

`147.66666666 * 3 = 442.999999997999` *this is the value the debugger gave me
`ceil(442.999999997999) = 443`
`443/3 = 147.6666666666`

Would this be a reasonable change to ASCeilPixelValue?

```
CGFloat ASCeilPixelValue(CGFloat f)
{
  CGFloat scale = ASScreenScale();
  ceil(f * scale - FLT_EPSILON) / scale;
}
```

Or is that too hacky and heavy handed?

aimobier:
@rcancro Maybe I know how to solve this problem. Although he still sometimes has the problem mentioned above, it is much better. Maybe we shouldn't use percentages to set the width. It's probably a good idea to use `constrainedSize: ASSizeRange` to calculate the actual width.

nguyenhuy:
I think we're definitely dealing with an issue caused by floating points precision on 3x devices. I think a solution along the line of @rcancro's is the best we can do.

[...]

PR #864, opened 2018-03-29
[Issue 838] Update ASCeilPixelValue and ASRoundPixelValue

rcancro:
Layouts can come back for 3x devices as a repeating decimal. Floats/doubles have a hard time representing these values and often the last digit or two will be  garbage. This garbage can result in unexpected values from `ceil` or `round`. I try to fix this by subtracting `FLT_EPSILON` from the value before calling `ceil` or `round`

https://github.com/TextureGroup/Texture/issues/838

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## master
 * Add your own contributions to the next release on the line below this with your name.
+- User FLT_EPSILON in ASCeilPixelValue and ASFloorPixelValue to help with floating point precision errors when computing layouts for 3x devices. [Ricky Cancro](https://github.com/rcancro) [#838](https://github.com/TextureGroup/Texture/pull/864)
 - Disable interface colescing and match to pre-colescing interface update behavior [Max Wang](https://github.com/wsdwsd0829) [#862](https://github.com/TextureGroup/Texture/pull/862) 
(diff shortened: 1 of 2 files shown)


wiseoldduck:
Yes I think `ASFloorPixelValue` would want `+FLT_EPSILON`, and `ASRoundPixelValue` would want to be left alone.

Adlai-Holler:
I am not a smart man… is it worth using ASCeilPixelValue? It seems like it's asking for this kind of trouble, and subpixel clipping of content isn't a serious risk.

Anyway, I support landing this change. I can't think of how it could produce bad results in practice.

wiseoldduck:
Yes fair warning I am quite naive as to how this is used in practice!

But am pretty sure if you apply this to `ASRoundPixelValue` and can come up with examples where it mattered it would be incorrect about half the time.

wiseoldduck:
OK I see the example in the comments cites `100.666666666669`, not what I was expecting to see  like `100.666666666667`, where we were addressing just the inability to capture a nonterminating rational in a `CGFloat`. There is apparently some other source of inaccuracy than that! I will take this conversation somewhere else as it is now more about my enlightenment than constructive review.  😄

rcancro:
@wiseoldduck I think you are right -- We probably shouldn't do anything to round. I'll remove that.

@Adlai-Holler When creating an ASLayout, it uses `ASCeilPixelValue` on the given size. I don't know the performance implications, but in the past making sure you staying on pixel bounds was important for keeping text and images crisp.

PR #866, opened 2018-04-04
Update Podspec

Adlai-Holler:
See inline comments.

The motivating factor for this diff is that the explicit specification of C++ options forced Pinterest's build system (based on Bazel) to build all Texture files using a C++ compiler. The recent file `ASRecursiveUnfairLock.m` cannot be built with a C++ compiler and so that broke the build. Other clients are also using this "all-C++" workaround as well.

The default C++ options (currently GNU++14) will probably be fine indefinitely. We can always reintroduce these flags if we decide to lean on, say, c++17 options but for now let's toss them overboard.

### DIFF
--- a/Texture.podspec
+++ b/Texture.podspec
@@ -61,4 +61,5 @@ Pod::Spec.new do |spec|
   # Include optional PINRemoteImage module
   spec.default_subspec = 'PINRemoteImage'
 
+  spec.social_media_url = 'https://twitter.com/TextureiOS'
 end
--- a/Texture.podspec
+++ b/Texture.podspec
@@ -14,8 +14,6 @@ Pod::Spec.new do |spec|
   spec.ios.weak_frameworks = 'AssetsLibrary'
   spec.weak_frameworks = 'Photos','MapKit'
 
-  spec.requires_arc = true
-
   spec.ios.deployment_target = '9.0'
   spec.tvos.deployment_target = '9.0'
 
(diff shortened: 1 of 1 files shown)


Texture.podspec
```diff
@@ -14,8 +14,6 @@ Pod::Spec.new do |spec|
   spec.ios.weak_frameworks = 'AssetsLibrary'
   spec.weak_frameworks = 'Photos','MapKit'
 
-  spec.requires_arc = true
-
```
Adlai-Holler: This has been the default since 2014.

Texture.podspec
```diff
@@ -41,7 +39,6 @@ Pod::Spec.new do |spec|
         # See https://github.com/facebook/AsyncDisplayKit/issues/1153
         'Source/TextKit/*.h',
     ]
-    core.xcconfig = { 'GCC_PRECOMPILE_PREFIX_HEADER' => 'YES' }
```
Adlai-Holler: We no longer use a prefix header.

Texture.podspec
```diff
@@ -65,10 +62,4 @@ Pod::Spec.new do |spec|
   spec.default_subspec = 'PINRemoteImage'
 
   spec.social_media_url = 'https://twitter.com/TextureiOS'
-  spec.library = 'c++'
-  spec.pod_target_xcconfig = {
-       'CLANG_CXX_LANGUAGE_STANDARD' => 'c++11',
-       'CLANG_CXX_LIBRARY' => 'libc++'
-  }
-
```
Adlai-Holler: (See PR description). These options have been here since the initial OSS release and aren't adding value anymore. The default clang C++ options under Xcode 9 (and probably Xcode 8) are sufficient.

Adlai-Holler:
Hmmm the linker is failing but it won't say what symbols it's not finding. Maybe it's missing the c++ standard library. I'll put that back.

PR #876, opened 2018-04-10
Interface coalescing sample

wsdwsd0829:
Test Diff


Source/Details/ASRangeController.mm
```diff
@@ -376,7 +386,7 @@ - (void)_updateVisibleNodeIndexPaths
         [newVisibleNodes addObject:node];
       }
       // Skip the many method calls of the recursive operation if the top level cell node already has the right interfaceState.
-      if (node.interfaceState != interfaceState) {
+      if (node.pendingInterfaceState != interfaceState) {
```
wsdwsd0829: cell when be pushed will have thrashed state: exit - enter - exit (first 2 in one run loop and last in another)
before this fix: the call will be cellExit - NewNode enter - collection exit
after: new node enter - cell exit - collection exit

PR #881, opened 2018-04-12
Add experiment to skip creating UIViews altogether for constants

Adlai-Holler:
This also fixes a missing `exp_unfair_lock` constant 😅 .

This can avoid some pretty serious pre-main time.

The diff is hard to read, so here's what it does:
- Create static `NSNumber *`'s for the two BOOLs. If you're not in the experiment, these will be set during `ASInitializeFrameworkMainThread`. If you are in the experiment, the initializer function does nothing.
- In the two accessor functions e.g. `BOOL ASDefaultAllowsGroupOpacity()` we have a dispatch_once that checks the static to see if it was set in the initialize function, and if not it selects the default based on UIKit documentation (by reading Info.plist).
- There was some thread-unsafety before which is maintained. Namely, the statics were not thread-affined and they still aren't but the dispatch_once has a memory barrier which should make us safer even than before.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4100d1d6b4c51f642f69a643c4096f44/raw/e6f8229161fb3fa742be1f8e7317b311df5739e2/log.txt)

PR #889, opened 2018-04-19
Fix ASBasicImageDownloader Fatal Exception: Key (*URL) cannot be nil

GeekTree0101:


hashemp206:
the top crash reason in my app is because of this!
is this pull request a solution? and if so when it will be merged?

hashemp206:
@Adlai-Holler in my NetworkImageNode, I check against null url, but still I have encounter this crash!
I think the url is getting nil under the hood! still confused, if it is Texture bug? or should I do some precautions?

GeekTree0101:
In Texture 2.7, ASBasicImageDownloaderContext crash increased 6 times rather then Texture2.6
```Fatal Exception: NSInvalidArgumentException
*** -[__NSDictionaryM setObject:forKeyedSubscript:]: key cannot be nil, ASThread.h line 416
+[ASBasicImageDownloaderContext contextForURL:]```

GeekTree0101:
https://github.com/TextureGroup/Texture/pull/1019 i just reopen this issue

PR #903, opened 2018-05-03
[ASTextNode] Fix a deadlock that could occur when enabling experimental ASTextNode2 via ASConfiguration

appleguy:
Because multiple threads can enter this allocWithZone: method around the same time, it is possible for one of them to setSuperclass first, and then the second thread would get stuck in an infinite loop.

When climbing the inheritance hierarchy, ASTextNode2 would be encountered by this second thread, and it would continue until reaching c == Nil. Since there was no case to catch this, an infinite loop would result.

Then the main thread can be deadlocked if a method like waitUntilAllUpdatesAreProcessed is called on ASCollectionView.

PR #905, opened 2018-05-03
Test text node allocation in race condition

wsdwsd0829:
This is to test https://github.com/TextureGroup/Texture/pull/903, 
it should fail CI tests before #903 merged in, should not afterwards.


Tests/ASTextNodeTests.m
```diff
@@ -67,7 +68,7 @@ - (void)setUp
 {
   [super setUp];
   _textNode = [[ASTextNode alloc] init];
-  
+  _textNodeBucket = [[NSMutableArray alloc] init];
```
Adlai-Holler: Let's make this variable local to the new test method, rather than an instance var.

Tests/ASTextNodeTests.m
```diff
@@ -255,6 +256,32 @@ - (void)testThatTheExperimentWorksCorrectly
...
+  [ASConfigurationManager test_resetWithConfiguration:config];
+  XCTestExpectation *exp = [self expectationWithDescription:@"wait for full bucket"];
+
+  dispatch_queue_t queue = dispatch_queue_create("com.texture.AsyncDisplayKit.ASTextNodeTestsQueue", DISPATCH_QUEUE_CONCURRENT);
+  dispatch_group_t g = dispatch_group_create();
+  for (int i = 0; i < 20; i++) {
+    dispatch_group_async(g, queue, ^{
+      ASTextNode *textNode = [[ASTextNodeSecondSubclass alloc] init];
+      XCTAssert([textNode isKindOfClass:[ASTextNode2 class]]);
+      @synchronized(self.textNodeBucket) {
```
Adlai-Holler: Seems like the dispatch group ended up not getting used, so let's remove it.

wsdwsd0829: Actually the test fails very rarely may be 1/30 when I tested locally for pre-fix code.
Also 
https://developer.apple.com/documentation/objectivec/1441536-class_setsuperclass?language=objc suggest we not use method. 
Any way to avoid this at all and achieve same idea without this runtime change of super class?

wsdwsd0829:
Adding a NSLog at end of for loop in allocWithZone of ASTextNode would cause failure consistently though, I am not sure how can we make the test fail more consistent for pre-fix case.

Tests/ASTextNodeTests.m
Adlai-Holler: Since we change the superclass before allocating any instances, I expect it's relatively safe for this limited usage. If the test is flaky then I think we should skip this diff – I doubt this issue will return now that the fix has landed.

wsdwsd0829:
The test is not flaky itself, it's just cannot test the problem well in sense that without the nil guard the test will pass. because it's rare to reach the race condition point.

Adlai-Holler:
OK since the test is not very powerful it's probably not worth having. I doubt this particular issue will re-emerge and if it did we wouldn't catch it. Let's skip this one and keep moving.

PR #911, opened 2018-05-07
#trivial Fix the C++ assertion

garrettmoon:


Source/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -26,6 +26,10 @@
 #import <map>
 #import <mutex>
 
+# ifndef __STRICT_ANSI__
+  #warning "Texture must be compiled with std=c++11 to prevent layout issues. gnu++ is not supported. This is hopefully temporary."
+# endif
```
Adlai-Holler: Remove the space between `# ifndef`

PR #916, opened 2018-05-10
[ASDisplayNode] Improve thread-safety of didExitHierarchy #trivial

nguyenhuy:
Since `didExitHierarchy` is called without the node's instance lock being held, the lock must be acquired before accessing `pendingInterfaceState`.


nguyenhuy:
Thanks, @maicki.

appleguy:
Thanks for the fix!

PR #917, opened 2018-05-11
[ASCollectionNode/ASTableNode] Fix a crash occurs while remeasuring cell nodes

nguyenhuy:
According to the stack trace below, this crash happened when `_cellsForLayoutUpdates` was mutated while being enumerated in `-[ASCollectionView layoutSubviews]`. Fix by making a copy of `_cellsForLayoutUpdates` (and immediately clear it) before passing it to `-[ASDataController relayoutNodes:nodesSizeChanged:`.

Apply the same fix to ASTableView.

screen shot 2018-05-11 at 4 40 21 pm

[Internal JIRA ticket](https://jira.pinadmin.com/browse/BUG-77357)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

nguyenhuy:
@Adlai-Holler Indeed, the only re-entrant path I can think of is `-setNeedsLayout` *and* `-layoutIfNeeded` being called during the re-measurement.

I agree that your solution sounds like the ideal behavior. However, since it'll be a bigger and riskier change, I'd propose to merge this diff as is and follow up on another diff. What do you think?

Adlai-Holler:
Yep thats the way

nguyenhuy:
Thanks!

PR #918, opened 2018-05-14
[ASTextNode2] Upgrade lock safety by protecting all ivars (including rarely-changed ones).

appleguy:
Although I don't know of any specific crashes caused by this, I think we should
lock all properties by default. There are also some indications of premature
optimization in keeping lock scope small, where it is actually important to
have transactional integrity, and also where the ASDisplayNode base class is
otherwise going to repeatedly re-lock the object anyway.

I think this will remain pretty efficient, especially with os_unfair_lock enabled.

### DIFF
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -175,8 +175,9 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   [super didLoad];
   
   // If we are view-backed and the delegate cares, support the long-press callback.
+  // Locking is not needed, as all instance variables used are main-thread-only.
   SEL longPressCallback = @selector(textNode:longPressedLinkAttribute:value:atPoint:textRange:);
-  if (!self.isLayerBacked && [_delegate respondsToSelector:longPressCallback]) {
+  if (!self.isLayerBacked && [self.delegate respondsToSelector:longPressCallback]) {
     _longPressGestureRecognizer = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(_handleLongPress:)];
     _longPressGestureRecognizer.cancelsTouchesInView = self.longPressCancelsTouches;
     _longPressGestureRecognizer.delegate = self;
@@ -189,9 +190,10 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   if (!super.supportsLayerBacking) {
     return NO;
   }
-  
+
+  ASLockScopeSelf();
   // If the text contains any links, return NO.
-  NSAttributedString *attributedText = self.attributedText;
+  NSAttributedString *attributedText = _attributedText;
   NSRange range = NSMakeRange(0, attributedText.length);
   for (NSString *linkAttributeName in _linkAttributeNames) {
     __block BOOL hasLink = NO;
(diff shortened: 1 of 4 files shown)


Source/ASTextNode2.mm
```diff
@@ -910,16 +927,14 @@ - (CGFloat)shadowOpacity
...
+  ASLockScopeSelf();
+
+  if (_shadowOpacity == shadowOpacity) {
+    return;
   }
-  
+
+  _shadowOpacity = shadowOpacity;
+
   [self setNeedsDisplay];
```
Adlai-Holler: Great place to use `ASLockedSelfCompareAssign`!

Source/ASTextNode2.mm
```diff
@@ -986,29 +998,36 @@ - (NSArray *)pointSizeScaleFactors
...
   }
 }
 
 - (void)setTruncationAttributedText:(NSAttributedString *)truncationAttributedText
 {
-  if (ASLockedSelfCompareAssignCopy(_truncationAttributedText, truncationAttributedText)) {
+  ASLockScopeSelf();
+  if (!ASObjectIsEqual(_truncationAttributedText, truncationAttributedText)) {
+    _truncationAttributedText = [truncationAttributedText copy];
     [self _invalidateTruncationText];
```
Adlai-Holler: Why this change? Is it so that we now call `_invalidateTruncationText` with the lock held?

Source/ASTextNode2.mm
```diff
@@ -210,6 +212,7 @@ - (BOOL)supportsLayerBacking
 
 - (void)setTextContainerInset:(UIEdgeInsets)textContainerInset
 {
+  ASLockScopeSelf();
   BOOL needsUpdate = !UIEdgeInsetsEqualToEdgeInsets(_textContainer.insets, textContainerInset);
   _textContainer.insets = textContainerInset;
```
Adlai-Holler: Good place for `ASLockedSelfCompareAssignCustom`

Source/ASTextNode2.mm
```diff
@@ -189,9 +190,10 @@ - (BOOL)supportsLayerBacking
   if (!super.supportsLayerBacking) {
     return NO;
   }
-  
+
+  ASLockScopeSelf();
   // If the text contains any links, return NO.
-  NSAttributedString *attributedText = self.attributedText;
+  NSAttributedString *attributedText = _attributedText;
```
Adlai-Holler: Since we just took the lock, can we go back to using the ivar here?

Source/ASTextNode2.mm
appleguy: Yes, otherwise there is no transactional integrity.

Adlai-Holler:
Just to note: `_textContainer` is a true invariant of the node and never changes. Its properties are atomic. We should hold the lock when making batch accesses to properties but otherwise it can be accessed without the lock. Its `-copy` method is also atomic.

Source/ASTextNode2.mm
Adlai-Holler: OK in cases such as this (side effect keeps lock), we can use the non-locked compare-assign macros, like so:

```
ASLockScopeSelf();
if (ASCompareAssignCopy(_truncationAttributedText, truncationATtributedText)) {
  [self _invalidateTruncationText];
}
```

I set it up to work either way!

appleguy: Great! This is definitely an improvement.

I'm having trouble getting the minutes to circle back to this; will try, maybe this weekend.

PR #930, opened 2018-05-21
Update text layout cache checking #trivial

Adlai-Holler:
Fixes the bug in #553 .

The solution in that diff was actually landed incidentally in #918 which fixes the correctness of the check, but degrades layout performance and memory by recomputing layouts more often.

This diff changes the check to have both the correctness fix and improved performance. I tested the diff against the sample code that @wsdwsd0829 gave and it fixes that case.

It also modifies ASTextLayout to use pixel-ceiling rather than point-ceiling.

### DIFF
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -401,28 +401,34 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   layoutCacheLock.unlock();
 
   CGRect containerBounds = (CGRect){ .size = container.size };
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

jparise:
@Adlai-Holler, looks like this could use a rebase.

PR #934, opened 2018-05-22
[ASTextNode2] Simplify compare-assign check & lock _pointScaleFactors accessor #trivial

appleguy:


appleguy:
These new ASCompare and AS*Lock methods are highly enjoyable. Thanks a lot for devising and materializing them, @Adlai-Holler !

PR #956, opened 2018-06-02
[WIP] Implement ASDisplayTree to manage locking during tree-level operations

Adlai-Holler:
Right now when we walk up or down a tree, there's no way to guarantee that we get a consistent picture without deadlocking.

This diff creates a new class `ASDisplayTree` which owns a lock and which performs all tree-level queries and mutations. A reference to the tree is pushed down to nodes when they enter the tree.

Tree modifications currently require us to take _three_ locks simultaneously – the source tree, the destination tree, and the node's property lock.

Still need to come up with a testing scaffold and think through if there's a way to modify the tree without taking all three locks (although I think taking all three simultaneously is actually safe, since we use a deadlock-avoidant approach. 

At this stage it would be great to take input! But it's also far from landing so feel free to ignore for now.

PR #958, opened 2018-06-05
Add support for acquiring multiple locks at once

Adlai-Holler:
- New protocol `ASLocking` which extends `NSLocking` and adds the `-tryLock` method.
- New macros `ASSynthesizeLockingMethodsWithMutex` and `ASSynthesizeLockingMethodsWithObject` to remove locking method boilerplate.
- Make all lockables conform to `ASLocking`.
- Add `ASDN::Mutex::tryLock()`.
- Add `ASLockSequence` which repeatedly tries to lock different locks in sequence.

No functional changes.

This is a precursor to #956 .

### DIFF
--- a/Source/ASLocking.h
+++ b/Source/ASLocking.h
@@ -1,0 +1,95 @@
+//
+//  ASLocking.h
+//  Texture
+//
+//  Copyright (c) 2018-present, Pinterest, Inc.  All rights reserved.
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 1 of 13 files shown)


Source/ASLocking.h
```diff
@@ -0,0 +1,149 @@
...
+- (BOOL)tryLock;
+
+@end
+
+/**
+ * A set of locks acquired during ASLockSequence.
+ */
+typedef struct {
+  unsigned count;
+  CFTypeRef _Nullable locks[32];
```
maicki: Do we do any checks against having more locks as 32? it seems if we call `ASUnlockSet` we could go over 32 to unlock if added more.

Adlai-Holler: You're right, done

PR #959, opened 2018-06-05
Generalize the main thread ivar deallocation system

Adlai-Holler:
This makes it possible to:
- Trampoline ivars from non-display-node classes.
- Require other classes of ivar to be trampolined to the main thread.


Adlai-Holler:
@nguyenhuy Thanks for the merge 💞

PR #967, opened 2018-06-13
Update showcase to include Apollo for Reddit

christianselig:
@nguyenhuy asked (admittedly awhile back) if it's okay to include my app ([Apollo for Reddit](https://apolloapp.io)) in the [Showcase](http://texturegroup.org/showcase.html), and I'd be honored if you all would be still interested in that. Really appreciate the library, it's at the very core of my app.


christianselig:
Likewise!

PR #968, opened 2018-06-14
Reduce usage of autorelease pools

Adlai-Holler:
Unfortunately, Foundation and UIKit are not going to be compiled with ARC any time soon.

So until then, calling methods like `+[NSArray array]` under ARC will not use the fast-return path and will result in `alloc -> autorelease -> retain` to get a +1 instead of just `alloc`. Plus draining the pool takes time! Why waste it?

### DIFF
--- a/Source/ASRunLoopQueue.mm
+++ b/Source/ASRunLoopQueue.mm
@@ -238,14 +238,12 @@ static void runLoopSourceCallback(void *info) {
 
 - (void)drain
 {
-  @autoreleasepool {
-    _lock.lock();
-    auto q = std::move(_queue);
-    _lock.unlock();
-    for (auto ref : q) {
-      // NOTE: Could check that retain count is 1 and retry later if not.
-      CFRelease(ref);
-    }
+  _lock.lock();
+  auto q = std::move(_queue);
+  _lock.unlock();
+  for (auto ref : q) {
+    // NOTE: Could check that retain count is 1 and retry later if not.
+    CFRelease(ref);
   }
 }
 
--- a/Source/Details/ASCollectionLayoutState.mm
+++ b/Source/Details/ASCollectionLayoutState.mm
@@ -159,7 +159,7 @@ elementToLayoutAttributesTable:[NSMapTable elementToLayoutAttributesTable]];
   }
 
   // Use a set here because some items may span multiple pages
-  NSMutableSet<UICollectionViewLayoutAttributes *> *result = [NSMutableSet set];
+  auto result = [NSMutableSet<UICollectionViewLayoutAttributes *> set];
   for (id pagePtr in pages) {
     ASPageCoordinate page = (ASPageCoordinate)pagePtr;
     NSArray<UICollectionViewLayoutAttributes *> *allAttrs = [_pageToLayoutAttributesTable objectForPage:page];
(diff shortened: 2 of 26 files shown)


Source/ASRunLoopQueue.mm
```diff
@@ -238,14 +238,12 @@ - (void)releaseObjectInBackground:(id  _Nullable __strong *)objectPtr
...
-    for (auto ref : q) {
-      // NOTE: Could check that retain count is 1 and retry later if not.
-      CFRelease(ref);
-    }
+  _lock.lock();
+  auto q = std::move(_queue);
+  _lock.unlock();
+  for (auto ref : q) {
+    // NOTE: Could check that retain count is 1 and retry later if not.
+    CFRelease(ref);
```
Adlai-Holler: I was silly for putting this in in the first place. Creating autorelease pools takes a lot of time and there's absolutely no chance of anything getting added to it in here 🤦‍♂️

Source/Details/NSIndexSet+ASHelpers.m
```diff
@@ -96,15 +96,4 @@ + (NSIndexSet *)as_sectionsFromIndexPaths:(NSArray<NSIndexPath *> *)indexPaths
...
-- (NSArray<NSIndexPath *> *)as_filterIndexPathsBySection:(id<NSFastEnumeration>)indexPaths
-{
-  NSMutableArray *result = [NSMutableArray array];
-  for (NSIndexPath *indexPath in indexPaths) {
-    if ([self containsIndex:indexPath.section]) {
-      [result addObject:indexPath];
-    }
-  }
-  return result;
-}
```
Adlai-Holler: Unused

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Details/ASCollectionLayoutState.mm
```diff
@@ -159,7 +159,7 @@ - (UICollectionViewLayoutAttributes *)layoutAttributesForElement:(ASCollectionEl
   }
 
   // Use a set here because some items may span multiple pages
-  NSMutableSet<UICollectionViewLayoutAttributes *> *result = [NSMutableSet set];
+  auto result = [NSMutableSet<UICollectionViewLayoutAttributes *> set];
```
maicki: Use `alloc init` in here too?

Source/Private/TextExperiment/Component/ASTextLayout.m
```diff
@@ -1959,7 +1959,7 @@ - (NSArray *)selectionRectsForRange:(ASTextRange *)range {
   range = [self _correctedRangeWithEdge:range];
   
   BOOL isVertical = _container.verticalForm;
-  NSMutableArray *rects = [NSMutableArray array];
+  NSMutableArray *rects = [[NSMutableArray<NSValue *> alloc] init];
```
maicki: We may should decide on a standard way to initialize these. E.g. we are now using different ways across the framework:
```
NSMutableArray *rects = [[NSMutableArray<NSValue *> alloc] init];
auto rects = [[NSMutableArray<NSValue *> alloc] init];
NSMutableArray<NSValue *> rects = [[NSMutableArray alloc] init];
```

Adlai-Holler: I completely agree. `auto` isn't available in C, so I've been aiming for:

```
C++: auto rects = [[NSMutableArray<NSValue *> alloc] init];
C: NSMutableArray *rects = [[NSMutableArray<NSValue *> alloc] init];
```

We could introduce a macro that evaluates to `__auto_type` like `as_auto` or `autotype`. The only drawback is we can't name the macro `auto` because the old C keyword takes precedence over macros.

I think I'll add `as_auto` in a follow up diff and see what people think.

PR #973, opened 2018-06-16
Remove unused/unneeded header macros

Adlai-Holler:
- `AS_ARRAY_COUNT` is replaced by C++ array iteration.
- `hidden/pure/const` aren't used.
- `ASDISPLAYNODE_WARN_UNUSED` is unused in favor of `AS_WARN_UNUSED_RESULT`
- `ASDISPLAYNODE_NOT_DESIGNATED_INITIALIZER` is unnecessary since we can mark the initializer unavailable.
- `ASDISPLAYNODE_NOTHROW` is superseded by the `noexcept` decorator from C++14

### DIFF
--- a/Source/Details/ASThread.h
+++ b/Source/Details/ASThread.h
@@ -160,7 +160,7 @@ namespace ASDN {
   public:
 #if !TIME_LOCKER
 
-    Locker (T &l) ASDISPLAYNODE_NOTHROW : _l (l) {
+    Locker (T &l) noexcept : _l (l) {
       _l.lock ();
     }
 
@@ -174,7 +174,7 @@ namespace ASDN {
 
 #else
 
-    Locker (T &l, const char *name = NULL) ASDISPLAYNODE_NOTHROW : _l (l), _name(name) {
+    Locker (T &l, const char *name = NULL) noexcept : _l (l), _name(name) {
       _ti = CACurrentMediaTime();
       _l.lock ();
     }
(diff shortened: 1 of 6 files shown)


Source/Details/ASThread.h
```diff
@@ -160,7 +160,7 @@ namespace ASDN {
   public:
 #if !TIME_LOCKER
 
-    Locker (T &l) ASDISPLAYNODE_NOTHROW : _l (l) {
+    Locker (T &l) noexcept : _l (l) {
```
maicki: I'm not pretty sure if we compile Texture with C++14 at the moment. Can you double check that please. Thanks!

Source/Private/_ASCoreAnimationExtras.mm
```diff
@@ -98,29 +98,29 @@ extern void ASDisplayNodeSetResizableContents(id<ASResizableContents> obj, UIIma
 
 NSString *ASDisplayNodeNSStringFromUIContentMode(UIViewContentMode contentMode)
 {
-  for (int i=0; i< ARRAY_COUNT(UIContentModeDescriptionLUT); i++) {
-    if (UIContentModeDescriptionLUT[i].contentMode == contentMode) {
-      return UIContentModeDescriptionLUT[i].string;
+  for (auto &e : UIContentModeDescriptionLUT) {
```
maicki: Maybe a better variable name as `e` in this cases ;)

Source/Details/ASThread.h
Adlai-Holler: Yep this is from `c++11` but glad I double-checked.

Source/Private/_ASCoreAnimationExtras.mm
Adlai-Holler: I think `e` is short for `entry` and maybe there's a better name out there but I don't really want to wait for CI to rebuild.

PR #979, opened 2018-06-21
Adds support for having multiple interface state delegates.

garrettmoon:
Hopefully in a performant way.


Source/ASDisplayNode+InterfaceState.h
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Interface state is available on ASDisplayNode and ASViewController, and
+ * allows checking whether a node is in an interface situation where it is prudent to trigger certain
+ * actions: measurement, data loading, display, and visibility (the latter for animations or other onscreen-only effects).
+ *
+ * The defualt state, ASInterfaceStateNone, means that the element is not predicted to be onscreen soon and
+ * preloading should not be performed. Swift: use [] for the default behavior.
+ */
+typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
```
garrettmoon: This is just moved, no changes.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASDisplayNode+InterfaceState.h
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * The node is not contained in a cell but it is in a window.
+     *
+     * Currently we only set `interfaceState` to other values for
+     * nodes contained in table views or collection views.
+     */
+    ASInterfaceStateInHierarchy   = ASInterfaceStateMeasureLayout | ASInterfaceStatePreload | ASInterfaceStateDisplay | ASInterfaceStateVisible,
+};
+
+@protocol ASInterfaceStateDelegate <NSObject>
```
garrettmoon: Only change here is that these are now all optional.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0bb363f27e6539e3bd15834724feeded/raw/5e02db10206145e1eb4ff52ea3c1919d605c6492/log.txt)

Source/ASDisplayNode.mm
```diff
@@ -543,8 +543,8 @@ - (void)_didLoad
   for (ASDisplayNodeDidLoadBlock block in onDidLoadBlocks) {
     block(self);
   }
-
-  [_interfaceStateDelegate nodeDidLoad];
+
```
nguyenhuy: nit: No trailing white spaces.

Source/ASNodeController+Beta.m
```diff
@@ -47,6 +47,11 @@ - (instancetype)init
   return self;
 }
 
+- (void)dealloc
+{
+    [_weakNode removeInterfaceStateDelegate:self];
```
nguyenhuy: `-removeInterfaceStateDelegate:` needs to be called on main, and objects are de-allocated off main by default (code [here](https://github.com/TextureGroup/Texture/blob/master/Source/ASMainThreadDeallocation.mm#L147)). Since `NSHashTable` is not thread-safe, we probably have to dispatch to main.

Source/ASDisplayNode+InterfaceState.m
```diff
@@ -0,0 +1,154 @@
...
+    if (self = [super init]) {
+        _interfaceDidChangeDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceDidEnterVisibleDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceDidExitVisibleDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceDidEnterDisplayDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceDidExitDisplayDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceDidEnterPreloadDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceDidExitPreloadDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceNodeDidLayoutDelegates = [NSHashTable weakObjectsHashTable];
+        _interfaceNodeDidLoadDelegates = [NSHashTable weakObjectsHashTable];
```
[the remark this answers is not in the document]
Adlai-Holler: On second thought, using `-respondsToSelector:` is only a tiny bit slower and allows for proxying, so that's probably the way to go. Still fast.

Source/ASDisplayNode.mm
```diff
@@ -3152,6 +3152,24 @@ - (BOOL)shouldScheduleDisplayWithNewInterfaceState:(ASInterfaceState)newInterfac
...
 }
 
+- (void)addInterfaceStateDelegate:(id <ASInterfaceStateDelegate>)interfaceStateDelegate
+{
+  ASDisplayNodeAssertMainThread();
+
+  // Not a fan of lazy loading, but this method won't get called very often and avoiding
+  // the overhead of creating this is probably worth it.
+  if (_interfaceStateDelegateManager == nil) {
+    [[ASDisplayNodeInterfaceDelegateManager alloc] init];
```
maicki: Missing an assign to `_interfaceStateDelegateManager`?

Source/ASDisplayNode.h
```diff
@@ -292,6 +262,24 @@ AS_EXTERN NSInteger const ASDefaultDrawingPriority;
  */
 @property (readonly) ASInterfaceState interfaceState;
 
+/**
+ * @abstract Adds a delegate to recieve notifications on interfaceState changes.
```
maicki: nit: Typo: "recieve"

Source/ASDisplayNode.h
```diff
@@ -292,6 +262,24 @@ AS_EXTERN NSInteger const ASDefaultDrawingPriority;
...
+ *
+ * @warning This must be called from the main thread.
+ *
+ * @see ASInterfaceState
+ */
+- (void)addInterfaceStateDelegate:(id <ASInterfaceStateDelegate>)interfaceStateDelegate;
+
+/**
+ * @abstract Removes a delegate from recieving notifications on interfaceState changes.
```
maicki: nit: "Typo"

Source/ASDisplayNode+InterfaceState.m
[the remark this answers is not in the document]
appleguy: I don't think we should have to create all these up front...since we would only have to use the respondsToSelector: once per class (not even once per delegate), we should get a significant win by relying on just one object tracking the delegates, and we could use a simple bitfield / NS_OPTIONS to cache the ASInterfaceStateDelegateMethodsImplemented (similar to the flags in ASCollection that cache what the delegate / datasource implement, but just an NS_OPTIONS).

appleguy:
This could be pretty useful! Could you share some background on the use case? I've found ASNodeController quite valuable in the cases I've come across, but maybe it doesn't work for the case you have.

I think we can design an ideal solution if we have the same / general use case in mind to know where we need to scale. It seems maybe more generalized than needed to track an arbitrary number of delegates.

Source/ASDisplayNode+InterfaceState.m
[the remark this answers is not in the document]
Adlai-Holler: In recent versions of Objective-C, `respondsToSelector:` is truly very fast. It amortizes to 1 additional objc_msgSend, 3 static function invocations, and one inline-assembly routine. It also has the added benefit of checking-from and storing-to the ObjC impcache, so the subsequent message-send is optimized.

Writing our own bitfield-by-class would amount to us reimplementing that portion of the impcache, without the benefit of direct access to the class method table. And even accessing our table – say it were in a singleton registry object – would require retaining-releasing it unless we coded around that, which would easily subsume the entire rest of the lookup. The added boilerplate is also nontrivial.

So there is a win available here, but I believe it's small. Let's avoid these calls only if we can do it in a simple and tremendously fast & scalable way. If we can't, then we shouldn't worry about it – I believe we can absolutely achieve far superior optimizations by using resources for other parts of the framework (such as transfer-collections or more judicious applications of `__unsafe_unretained`). 

Implementations;
- [`-respondsToSelector:`](https://opensource.apple.com/source/objc4/objc4-723/runtime/NSObject.mm)
- [`class_respondsToSelector_inst`](https://opensource.apple.com/source/objc4/objc4-723/runtime/objc-class.mm.auto.html)
- [`lookupImpOrForward` and `lookupImpOrNil`](https://opensource.apple.com/source/objc4/objc4-723/runtime/objc-runtime-new.mm)
- [ARM64: `cache_getImp`, used by both `objc_msgSend` and `lookupImpOrNil`](https://opensource.apple.com/source/objc4/objc4-723/runtime/Messengers.subproj/objc-msg-arm64.s)

garrettmoon:
@appleguy We have two use cases, both related to tracking performance. It'd be ideal if there were separate managers for each of these which could monitor the states.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/8f04c80f5888515622c5385c425fb02a/raw/992697e408f87a9de95b9d56349d915a9eb43dab/log.txt)

Source/ASDisplayNode.mm
```diff
@@ -433,6 +433,9 @@ - (void)dealloc
 
   // TODO: Remove this? If supernode isn't already nil, this method isn't dealloc-safe anyway.
   [self _setSupernode:nil];
+  
+  ASPerformMainThreadDeallocation(&_interfaceStateDelegates);
```
nguyenhuy: I don't think this is necessary. If the node is gone, `_interfaceStateDelegates` can be safely deallocated as well, especially because it only stores weak references.

Source/ASNodeController+Beta.m
nguyenhuy: @garrettmoon I still think that we should dispatch to main here, no?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/db3fb77dd173d652f1a6995c834f4992/raw/ad5847a63e4dfe69687631927c2371797910b642/log.txt)

PR #981, opened 2018-06-21
Changed lost images to existing one. #trivial

tataevr:
New images taken from http://texturegroup.org/docs/automatic-layout-containers.html.


tataevr:
@maicki yeah, i can see them. But when you look to github markdown - they are lost. 
Why i'm reading tutorial at github - because of markdown, you can perfectly see all markdown element (which you can't see at http://texturegroup.org/docs/automatic-layout-containers.html).

So, that's why i suggest to update images at github markdown document.

nguyenhuy:
@tataevr I see. Would be great if we can fix other links throughout as well. Let's land this diff as is for now.

PR #982, opened 2018-06-22
Optimize layout flattening

Adlai-Holler:
Layout flattening is one of the most common operations in our framework and it currently involves a ton of needless retain/release traffic (4 rr pairs per layout). This diff reduces that to 0.

Since I hope to make more use of the `__unsafe_unretained` qualifier in the future, I added an `unowned` alias for it, to make it prettier.

Obviously please review this diff carefully, since no optimization is worth crashing.

### DIFF
--- a/Source/Base/ASBaseDefines.h
+++ b/Source/Base/ASBaseDefines.h
@@ -18,6 +18,7 @@
 #import <Foundation/NSObjCRuntime.h>
 
 #define AS_EXTERN FOUNDATION_EXTERN
+#define unowned __unsafe_unretained
 
 #ifdef __GNUC__
 # define ASDISPLAYNODE_GNUC(major, minor) \
--- a/Source/Layout/ASLayout.mm
+++ b/Source/Layout/ASLayout.mm
@@ -243,6 +243,7 @@ static std::atomic_bool static_retainsSublayoutLayoutElements = ATOMIC_VAR_INIT(
     queue.pop_front();
     
     unowned ASLayout *layout = context.layout;
+    // Direct ivar access to avoid retain/release, use existing +1.
     const NSUInteger sublayoutsCount = layout->_sublayouts.count;
     const CGPoint absolutePosition = context.absolutePosition;
     
--- a/Source/Layout/ASLayout.mm
+++ b/Source/Layout/ASLayout.mm
@@ -55,22 +55,6 @@ ASDISPLAYNODE_INLINE AS_WARN_UNUSED_RESULT BOOL ASLayoutIsDisplayNodeType(ASLayo
   return layout.type == ASLayoutElementTypeDisplayNode;
 }
 
-ASDISPLAYNODE_INLINE AS_WARN_UNUSED_RESULT BOOL ASLayoutIsFlattened(ASLayout *layout)
-{
-  // A layout is flattened if its position is null, and all of its sublayouts are of type displaynode with no sublayouts.
-  if (! ASPointIsNull(layout.position)) {
-    return NO;
-  }
-  
-  for (ASLayout *sublayout in layout.sublayouts) {
-    if (ASLayoutIsDisplayNodeType(sublayout) == NO || sublayout.sublayouts.count > 0) {
-      return NO;
-    }
-  }
-  
-  return YES;
-}
-
 @interface ASLayout () <ASDescriptionProvider>
 {
   ASLayoutElementType _layoutElementType;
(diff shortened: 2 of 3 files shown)


Source/Layout/ASLayout.mm
```diff
@@ -55,22 +55,6 @@ ASDISPLAYNODE_INLINE AS_WARN_UNUSED_RESULT BOOL ASLayoutIsDisplayNodeType(ASLayo
...
-  
-  for (ASLayout *sublayout in layout.sublayouts) {
-    if (ASLayoutIsDisplayNodeType(sublayout) == NO || sublayout.sublayouts.count > 0) {
-      return NO;
-    }
-  }
-  
-  return YES;
-}
-
```
Adlai-Holler: Had to make this a method so that it can see the `_sublayouts` ivar for direct access (remove retain/release pair per sublayout).

Source/Layout/ASLayout.mm
```diff
@@ -216,52 +200,70 @@ - (void)setRetainSublayoutLayoutElements:(BOOL)retainSublayoutLayoutElements
...
-      for (ASLayout *sublayout in sublayouts) {
-        sublayoutContexts.push_back({sublayout, absolutePosition + sublayout.position});
+    } else if (sublayoutsCount > 0) {
+      // Fast-reverse-enumerate the sublayouts array by copying it into a C-array and push_front'ing each into the queue.
+      unowned ASLayout *rawSublayouts[sublayoutsCount];
+      [layout->_sublayouts getObjects:rawSublayouts range:NSMakeRange(0, sublayoutsCount)];
+      for (NSInteger i = sublayoutsCount - 1; i >= 0; i--) {
+        queue.push_front({rawSublayouts[i], absolutePosition + rawSublayouts[i].position});
       }
-      queue.insert(queue.cbegin(), sublayoutContexts.begin(), sublayoutContexts.end());
```
Adlai-Holler: The bottom here (the search expansion) is far and away the biggest win. We no longer create another `vector` (heap allocation) and we remove 2 retain/release pairs per sublayout. The first is inserting into the vector then destroying the vector, the second is inserting into the queue and then removing from the queue.

Source/Layout/ASLayout.mm
```diff
@@ -216,52 +200,70 @@ - (void)setRetainSublayoutLayoutElements:(BOOL)retainSublayoutLayoutElements
...
   
   while (!queue.empty()) {
     const Context context = std::move(queue.front());
     queue.pop_front();
     
-    ASLayout *layout = context.layout;
-    const NSArray<ASLayout *> *sublayouts = layout.sublayouts;
-    const NSUInteger sublayoutsCount = sublayouts.count;
+    unowned ASLayout *layout = context.layout;
+    const NSUInteger sublayoutsCount = layout->_sublayouts.count;
```
Adlai-Holler: Direct ivar access isn't so much about avoiding the message-send, it's about avoiding the retain/release pair. `_sublayouts` is already a `+1` so let's use it.

Source/Base/ASBaseDefines.h
```diff
@@ -18,6 +18,7 @@
 #import <Foundation/NSObjCRuntime.h>
 
 #define AS_EXTERN FOUNDATION_EXTERN
+#define unowned __unsafe_unretained
```
maicki: Not sure if we should just be consistent and use `__unsafe_unretained` in a framework.

Source/Layout/ASLayout.mm
```diff
@@ -216,52 +200,70 @@ - (void)setRetainSublayoutLayoutElements:(BOOL)retainSublayoutLayoutElements
 
 #pragma mark - Layout Flattening
 
+- (BOOL)isFlattened
+{
+  // A layout is flattened if its position is null, and all of its sublayouts are of type displaynode with no sublayouts.
+  if (! ASPointIsNull(_position)) {
```
maicki: Remove whitespace?

Source/Layout/ASLayout.mm
```diff
@@ -216,52 +200,70 @@ - (void)setRetainSublayoutLayoutElements:(BOOL)retainSublayoutLayoutElements
...
+    return NO;
+  }
+  
+  for (ASLayout *sublayout in _sublayouts) {
```
maicki: Would `NSFastEnumeration` allow us to cast that as `unowned void *` and cast down with a bridge to the type we need? In fact does it do a retain for the sub layouts in here?

Adlai-Holler: Fast enumeration doesn't retain the objects.

Source/Base/ASBaseDefines.h
Adlai-Holler: Pros:
- Less underscores
- Shorter, cleaner

Cons:
- Not standard
- Could one day be a collision and we'd have to revert

Source/Layout/ASLayout.mm
Adlai-Holler: Done

Issue #834, opened 2018-03-13
ListDisplayDelegate method not be called in the latest version


bawn:
Version: 2.6

I added this code in the demo `Texture-master/examples/ASDKgram/Sample/PhotoFeedSectionController.m `

Add  IGListDisplayDelegate protocol
```
@interface PhotoFeedSectionController () <ASSupplementaryNodeSource, IGListSupplementaryViewSource, IGListDisplayDelegate>
```

Set displayDelegate
```
- (instancetype)init
{
  if (self = [super init]) {
    _paginatingSpinner = @"Paginating Spinner";
    self.supplementaryViewSource = self;
    self.displayDelegate = self;
  }
  return self;
}
```

Implementation protocol method

```
- (void)listAdapter:(nonnull IGListAdapter *)listAdapter didEndDisplayingSectionController:(nonnull IGListSectionController *)sectionController {
  
}

- (void)listAdapter:(nonnull IGListAdapter *)listAdapter didEndDisplayingSectionController:(nonnull IGListSectionController *)sectionController cell:(nonnull UICollectionViewCell *)cell atIndex:(NSInteger)index {
  
}

- (void)listAdapter:(nonnull IGListAdapter *)listAdapter willDisplaySectionController:(nonnull IGListSectionController *)sectionController {
  
}

- (void)listAdapter:(nonnull IGListAdapter *)listAdapter willDisplaySectionController:(nonnull IGListSectionController *)sectionController cell:(nonnull UICollectionViewCell *)cell atIndex:(NSInteger)index {
  NSLog(@"%@", cell);
}
```

But no method is called, and I sure these operations are working in 2.5.1

Finally I find the solution way, but I not sure this is completely correct.

In ASIGListAdapterBasedDataSource.m add following code
```
- (void)collectionNode:(ASCollectionNode *)collectionNode willDisplayItemWithNode:(ASCellNode *)node {
    [self.delegate collectionView:collectionNode.view willDisplayCell:node.view forItemAtIndexPath:node.indexPath];
}

- (void)collectionNode:(ASCollectionNode *)collectionNode didEndDisplayingItemWithNode:(ASCellNode *)node {
    [self.delegate collectionView:collectionNode.view didEndDisplayingCell:node.view forItemAtIndexPath:node.indexPath];
}

```

yoali11:
I can confirm that they are not called...

PR #1011, opened 2018-07-05
[IGListKit] Extended IGListKit support for displaying delegates

hebertialmeida:
This pull request is related to #834. I'm opening this with a possible solution but also to discuss if that is the best approach for fixing this problem.

It basically introduces the following improvements when using Texture + IGListKit:

`IGListDisplayDelegate`:

```swift
func listAdapter(_ listAdapter: ListAdapter, willDisplay sectionController: ListSectionController)
func listAdapter(_ listAdapter: ListAdapter, didEndDisplaying sectionController: ListSectionController)
func listAdapter(_ listAdapter: ListAdapter, willDisplay sectionController: ListSectionController, cell: UICollectionViewCell, at index: Int)
func listAdapter(_ listAdapter: ListAdapter, didEndDisplaying sectionController: ListSectionController, cell: UICollectionViewCell, at index: Int)
```

`IGListSectionController`:

```swift
func didDeselectItem(at index: Int)
func didHighlightItem(at index: Int)
func didUnhighlightItem(at index: Int)
```

Also, I am proposing a rename of two methods because:
1. `collectionNode:willDisplayItemWithNode:` don't have the indexPath like UICollectionView's method `collectionView:willDisplayCell:forItemAtIndexPath:`, although we can get it from the node it might be `nil`, in that case, it will crash because IGListKit needs a non optional IndexPath and it is asserting for that [here](https://github.com/Instagram/IGListKit/blob/master/Source/Internal/IGListDisplayHandler.m#L46).
2. Most of the `ASCollectionNode` names as related to `UICollectionView` and these are not

```objective-c
// From:
- (void)collectionNode:(ASCollectionNode *)collectionNode willDisplayItemWithNode:(ASCellNode *)node
- (void)collectionNode:(ASCollectionNode *)collectionNode didEndDisplayingItemWithNode:(ASCellNode *)node

// To:
- (void)collectionNode:(ASCollectionNode *)collectionNode willDisplayNode:(ASCellNode *)node forItemAtIndexPath:(NSIndexPath *)indexPath
- (void)collectionNode:(ASCollectionNode *)collectionNode didEndDisplayingNode:(ASCellNode *)node forItemAtIndexPath:(NSIndexPath *)indexPath
```

This will not only conform to the UIKit naming but also have a strong indexPath.

Please let me know what you think and if I am going in the right direction.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Adlai-Holler:
Hi @hebertialmeida because of the asynchronous nature of ASCollectionNode, there can be multiple index paths associated with a node at any given time. There's the index path of the node in data source terms, and the index path in UIKit terms.

In the appearance callback methods, we know that there is a UIKit index path, but there might not be a data source index path (if the node has been deleted but the delete has not been committed yet.)

To get a hold of this view-index-path and pass it to IGListKit, call `indexPathForNode:` on the `ASCollectionView` object (as opposed to ASCollectionNode). You can then pass that on and it will be nonnull.

In the future, it might make sense to augment these callbacks by providing _both_ index paths, rather than providing neither. In the past, the one index path argument was dropped because users were attempting to use it to access their data sources, leading to hard-to-debug crashes.

Adlai-Holler:
@hebertialmeida Awesome! Do we need to get the actual `UICollectionViewCell` to give to IGListKit, rather that the node's view (which goes in the content view?)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1fadc20b30b4a680d12aaefa34727cf4/raw/9798ab714cbcf5d55a89b27b25e692478c1c4b18/log.txt)

hebertialmeida:
We are sending the `UICollectionViewCell` here `willDisplayCell:node.view` right? In my tests, it works perfectly.

I don't know why CI is failing now, there are fewer changes...

Adlai-Holler:
The CI failed thanks to a transient bug in Xcode simulator. I restarted it, should work.

`node.view` is actually a plain UIView that we put inside the UICollectionViewCell's content view. We need to walk up the view hierarchy to find the actual UICollectionViewCell.

hebertialmeida:
Alright, I've just pushed the changes, as the `superview` is optional I am checking if it is `nil` just in case.

hebertialmeida:
While testing I did notice that in some cases the indexPath from `indexPathForNode:` can be `nil`, so I also added a safety check for that.

Source/Private/ASIGListAdapterBasedDataSource.m
```diff
@@ -139,6 +154,32 @@ - (void)collectionNode:(ASCollectionNode *)collectionNode willBeginBatchFetchWit
...
+{
+  NSIndexPath *indexPath = [collectionNode.view indexPathForNode:node];
+  UIView *contentView = node.view.superview;
+  UICollectionViewCell *cell = contentView.superview;
+
+  if (cell == nil || indexPath == nil) {
+    return;
+  }
+
+  [self.delegate collectionView:collectionNode.view didEndDisplayingCell:cell forItemAtIndexPath:indexPath];
```
Adlai-Holler: Actually, since we conform to `ASCollectionDelegateInterop` we should be already bridging these. See lines ~190 below, where we pass these calls directly through to the delegate.

Can you put breakpoints in there or investigate why we're not seeing them?

hebertialmeida: So, I've checked that and it never fires the delegate because don't pass this checking:
```obj-c
    if (node.shouldUseUIKitCell) {
      [(id <ASCollectionDelegateInterop>)_asyncDelegate collectionView:collectionView willDisplayCell:rawCell forItemAtIndexPath:indexPath];
    }
```
Here: https://github.com/TextureGroup/Texture/blob/master/Source/ASCollectionView.mm#L1183

I am trying to understand the `shouldUseUIKitCell` and why do we need the check here to fire that.

hebertialmeida: If I comment out this check it works, Also I noticed that this will be always `NO` and will never fire the delegate because it is a read-only property, it is only "YES" if it is an `ASWrapperCellNode`.

Here it was checking:
https://github.com/TextureGroup/Texture/blob/550da242b7852563cbcbf409cb9e3373e626304a/Source/ASCollectionView.mm#L1100

It was first introduced here:
https://github.com/TextureGroup/Texture/commit/c6e3dd7a5d8bb0dc4dd26b96f96bf96946ade806#diff-8b82b1556d1293f6f9789beb236c9271R1167

Simplified here:
https://github.com/TextureGroup/Texture/commit/63efdbde8f958fcf5cf09a8df19e96cb90843a54#diff-8b82b1556d1293f6f9789beb236c9271R1169

For `ASWrapperCellNode` I have also found this:
https://github.com/TextureGroup/Texture/blob/master/Source/ASCellNode.mm#L387-L393

and there is a comment 
>// TODO: Consider if other calls, such as willDisplayCell, should be bridged to this class.

MiguelChiang: Looks like @hebertialmeida is pointing out the problem, do this issue has any update?

Manual modifying ASIGListAdapterBasedDataSource.m isn't a nice solution for now as it is a private class. So it will be great if this issue get sort out and merge to master soon.

Thanks guys for the help.

hebertialmeida: @MiguelChiang Still waiting someone get back to me.

hebertialmeida:
@Adlai-Holler any updates on this?

3a4oT:
ping pong

Source/Private/ASIGListAdapterBasedDataSource.m
3a4oT: @Adlai-Holler  can you please take a look when you get a chance?

WeiweiYuan810:
do this issue has any update？

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR modifies CHANGELOG.md, which is a generated file. Add #changelog to the title to suppress this warning.</td>
    </tr>
  </tbody>
</table>

</p>

hebertialmeida:
@Adlai-Holler I have being using this in production for at least 6 months already and not problems so far, do you think this is something that could be merged in at some point?

Few other users have commented on this, might be useful for someone else.

chml:
Any Update?

bolsinga:
Please rebase if possible to see if we can get this to build green based upon master. Thank you for your help!

hebertialmeida:
My branch is 255 commits behind I think it is going to be painful... Is there any problem if I merge instead of rebasing?

bolsinga:
> My branch is 255 commits behind I think it is going to be painful... Is there any problem if I merge instead of rebasing?

I don't mind; the CI has changed over the last several years(!) and I want to ensure your PR passes CI.

hebertialmeida:
I have merged but it looks like the master is also not passing the CI

hebertialmeida:
@bolsinga It's all green now!

PR #1019, opened 2018-07-11
[ASNetworkImageNode] Fix ASNetworkImageNode downloadImageWithCompletion URL nil crash

GeekTree0101:
When i forcefully injected nil to _URL property before calling downloadImageWithURL method then i faced same crash like an upper screenshot
I guess that _URL became to nil before reobtain the lock on downloadImageWithCompletion method.

### DIFF
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -561,7 +561,12 @@
     // it and try again.
     {
       ASLockScopeSelf();
-      url = _URL;
+      if (_URL) {
+        url = _URL;
+      } else {
+        [self _locked_cancelDownloadAndClearImageWithResumePossibility:NO];
+        return;
+      }
     }
 
 


GeekTree0101:
Recently, My Product Feed UIs replaced ASNetworkImageNode to ASVideoNode
After that increased upper crash 5~6 times
My ASVideoNode doesn't use URL property (URL is nil). In this case, I show that ASVideoNode generate placeholderImage.

##### ASVideoNode.mm line 183
 ```objc 
  if (self.image == nil && self.URL == nil) {
    [self generatePlaceholderImage];
  }
```

after that ASVideoNode call setImage method with _URL setup nil
(because ASNetworkImageNode should cancel & clear at _locked_setImage method )

#### ASNetworkImageNode line 144
```objc 
/// Setter for public image property. It has the side effect of setting an internal _imageWasSetExternally that prevents setting an image internally. Setting an image internally should happen with the _setImage: method
- (void)setImage:(UIImage *)image
{
  ASLockScopeSelf();
  [self _locked_setImage:image];
}
```

#### ASNetworkImageNode line 150
```objc

- (void)_locked_setImage:(UIImage *)image
{
  BOOL imageWasSetExternally = (image != nil);
  BOOL shouldCancelAndClear = imageWasSetExternally && (imageWasSetExternally != _imageWasSetExternally);
  _imageWasSetExternally = imageWasSetExternally;
  if (shouldCancelAndClear) {
    ASDisplayNodeAssertNil(_URL, @"Directly setting an image on an ASNetworkImageNode causes it to behave like an ASImageNode instead of an ASNetworkImageNode. If this is what you want, set the URL to nil first.");
    _URL = nil;
    [self _locked_cancelDownloadAndClearImageWithResumePossibility:NO];
  }

```
In this case, does ASVideoNode can occur upper crash?

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

GeekTree0101:
@nguyenhuy  Okey got it, thank your feedback

Issue #932, opened 2018-05-22
Assertion failure in void ASDN::Mutex::unlock()()


choiks14:
this is my code 

```

    DispatchQueue.main.async {
      self.present(UINavigationController(rootViewController: SigninViewController()), animated: false)
    }
```

below is exceptions.

what happend?
please help me.

```
2018-05-22 14:40:22.864056+0900 jjaljub-iphone[67826:4720838] [Node] Propagating new traits for <jjaljub_iphone.AuthViewController: 0x7fb76d903e80>: { userInterfaceIdiom = Phone; containerSize = {375, 667}; horizontalSizeClass = Compact; verticalSizeClass = Regular; forceTouchCapability = Unavailable }
2018-05-22 14:40:22.868868+0900 jjaljub-iphone[67826:4720838] [Node] View controller <jjaljub_iphone.AuthViewController: 0x7fb76d903e80> will appear
2018-05-22 14:40:26.475113+0900 jjaljub-iphone[67826:4720838] [Node] Propagating new traits for <jjaljub_iphone.SigninViewController: 0x7fb76b426b30>: { userInterfaceIdiom = Phone; containerSize = {375, 667}; horizontalSizeClass = Compact; verticalSizeClass = Regular; forceTouchCapability = Unavailable }
2018-05-22 14:40:26.480710+0900 jjaljub-iphone[67826:4720838] [Node] View controller <jjaljub_iphone.SigninViewController: 0x7fb76b426b30> will appear
2018-05-22 14:40:26.481517+0900 jjaljub-iphone[67826:4720838] *** Assertion failure in std::vector<NSUInteger> findNodesInLayoutAtIndexesWithFilteredNodes(ASLayout *__strong, NSIndexSet *__strong, NSArray<ASDisplayNode *> *__strong, NSArray<ASDisplayNode *> *__strong *)(), /Users/kangsikchoi/Workspace/jjal/jjaljub-iphone/Pods/Texture/Source/Private/ASLayoutTransition.mm:256
2018-05-22 14:40:26.482139+0900 jjaljub-iphone[67826:4720838] *** Assertion failure in void ASDN::Mutex::unlock()(), /Users/kangsikchoi/Workspace/jjal/jjaljub-iphone/Pods/Texture/Source/Details/ASThread.h:238
[...]

PR #1022, opened 2018-07-12
Fix misleading/scary stack trace shown when an assertion occurs during node measurement

nguyenhuy:
- Currently, there is a pair of mutex unlock and unlock that wraps around `-_u_measureNodeWithBoundsIfNecessary:` in `__layout`. That is because this method must be called without the lock.
- When an assertion occurs within that method, the runtime bails early without reacquire the lock (so the lock is free now). However, the runtime then hits the end of the outmost mutex locker scope and tries to release the lock that it no longer holds, causing another assertion in ASThread to be shown to user (#932) (see [this video](https://cl.ly/stlh) here for more details). This makes it extremely hard to idenfity the root assertion.
- Fix by replacing the unlock/lock pair with a mutex unlocker.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -1040,10 +1040,11 @@ ASSynthesizeLockingMethodsWithMutex(__instanceLock__);
 
     // This method will confirm that the layout is up to date (and update if needed).
     // Importantly, it will also APPLY the layout to all of our subnodes if (unless parent is transitioning).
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #1023, opened 2018-07-12
Make sure -_completePendingLayoutTransition is called without the node's instance lock #trivial

nguyenhuy:
This is because committing the layout transition (aka `-_completeLayoutTransition:`) results in subnode insertions and removals which must be called lock-free.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/07f0a2c6f6f2e93c8b4d8378c8ec57ff/raw/e836a7ccafa2fdef008c6cf7fbaa95e83e88e8a2/log.txt)

Issue #969, opened 2018-06-15
Thread-local storage is not supported for the current target + AsyncDisplayKit Framework name


Kaspik:
Hey Texture team! I haven't found any issue here (except https://github.com/TextureGroup/Texture/issues/873).

I have problem with tests target. My main target runs okay on both XCode 9.4 and XCode 10.
But tests target on main target is not running because of:
1) Framework not found
```
Ld /Users/kaspik/Library/Developer/Xcode/DerivedData/ClassDojo-atbalbdthvzlqngvkcwmnjpzninu/Build/Products/Testing-iphonesimulator/ClassDojo.app/ClassDojo normal x86_64 (in target: ClassDojo)
    cd /Users/kaspik/Documents/Vyvoj/STRV/dojo-behavior
    export IPHONEOS_DEPLOYMENT_TARGET=9.0
[...]

Issue #929, opened 2018-05-21
ASTableNode deleteRowsAtIndexPath Crash


fuhailong:
[weakSelf.tableNode deleteRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationNone];
crash。
Have anyone ever met this problem？

nguyenhuy:
Hi, please share more details the crash, like stack traces, version of the framework that you're using, and maybe a sample project that can reliably reproduce this crash. Thanks!

fuhailong:
I found this problem on the IOS version of the 11.3.1 device.
I tried again on the IOS version of 10.3.3 for normal use.

I don't know where to view the frame version. I download it from the master of AsyncDisplayKit project. Due to support ios8.0, Texture project will be wrong, so the use of AsyncDisplayKit project.

> ASAssert.m
> Thread-local storage is not supported for the current target. 

If the minimum support for ios9.0, compile it normally.

At this point, All exception breakpoint is positioned in ASTableView.mm, line:1638.

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x0000000184dbd2ec libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x0000000184f5e288 libsystem_pthread.dylib`pthread_kill$VARIANT$mp + 376
    frame #2: 0x0000000184d2bd0c libsystem_c.dylib`abort + 140
    frame #3: 0x00000001844c72c8 libc++abi.dylib`abort_message + 132
    frame #4: 0x00000001844e1398 libc++abi.dylib`std::__terminate(void (*)()) + 44
    frame #5: 0x00000001844e0ccc libc++abi.dylib`__cxa_throw + 132
    frame #6: 0x00000001844f0720 libobjc.A.dylib`objc_exception_throw + 364
    frame #7: 0x0000000185336bf8 CoreFoundation`+[NSException raise:format:arguments:] + 104
    frame #8: 0x0000000185d26fa0 Foundation`-[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 112
    frame #9: 0x000000018f381194 UIKit`-[UITableView _heightForShadowRowAtIndexPath:] + 476
    frame #10: 0x000000018f4c804c UIKit`-[_UITableViewShadowUpdatesController(UITableViewDelegate) tableView:heightForRowAtIndexPath:] + 172
[...]

Issue #873, opened 2018-04-06
Texture + Xcode 9.3/Swift 4.1 via Carthage is broken


rainypixels:
I downloaded Xcode 9.3 last night, ran `carthage update` (pointed at Texture's `release` branch), and rebuilt my project. It builds successfully, but `SIGABRT`s on run. I've done all the usual stuff: deleted derived data, removed texture and dependencies entirely from checkout/build folders (and done a fresh install), etc. but nothing has worked. I've also tried the `master` branch which bails during `carthage update` as follows:

```
🚀  carthage update --platform iOS

*** Fetching texture
*** Fetching PINOperation
*** Checking out PINRemoteImage at "3.0.0-beta.13"
*** Checking out PINOperation at "1.1.1"
*** Checking out libwebp at "v0.6.0"
*** Checking out PINCache at "3.0.1-beta.6"
*** Checking out texture at "2a0c6f8a420da91617202aa7bca67314ef8c97e4"
*** xcodebuild output can be found in /var/folders/b1/5l9q543x2mx8bb3zmfhl34m80000gn/T/carthage-xcodebuild.dtfS3T.log
*** Downloading PINCache.framework binary at "3.0.1-beta.6"
*** Downloading PINRemoteImage.framework binary at "3.0 beta 13: Animated WebP support / better GIF handling"
*** Downloading PINOperation.framework binary at "Just fixing some warnings"
*** Skipped building libwebp due to the error:
Dependency "libwebp" has no shared framework schemes for any of the platforms: iOS
*** Building scheme "AsyncDisplayKit" in AsyncDisplayKit.xcworkspace
Build Failed
	Task failed with exit code 65:
[...]

PR #1025, opened 2018-07-12
Fix 32-bit simulator build on Xcode >= 9.3

Adlai-Holler:
Fixes part of the issue discussed in #873 #929 and #969.

We now provide fallbacks for i386.


3a4oT:
thanks, @Adlai-Holler !! would be great to have ETA for next release=)

PR #1027, opened 2018-07-13
Remove misleading comment and add assertion #trivial

wiseoldduck:
The comment at line 305 strongly implies this is expected to be a dynamic cast.


nguyenhuy:
On a second thought, I think @Adlai-Holler is right. These find nodes functions are called with flattened layouts only.

wiseoldduck:
Nooooooooooo!

Yeah no, actually I agree - I forgot I had made this PR. It all makes sense I will adjust.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/73edad0963812507503e2102dcbfdf8f/raw/4af22add0508a1cbd9236167bc0bd5f31ab9b8e9/log.txt)

wiseoldduck:
@maicki Done-and-done

Issue #910, opened 2018-05-06
I need help about ASCornerRoundingTypeClipping


LisonFan:
I use `ASCornerRoundingTypeClipping` mode, but app crash
My code
```objective-c
self.newsImageNode = [[ASNetworkImageNode alloc] init];
self.newsImageNode.URL = [NSURL URLWithString:self.model.pic];
self.newsImageNode.placeholderFadeDuration = 0.2;
self.newsImageNode.placeholderEnabled = YES;
self.newsImageNode.cornerRoundingType = ASCornerRoundingTypeClipping;
self.newsImageNode.backgroundColor = [UIColor whiteColor];
self.newsImageNode.cornerRadius = 10.f;
[self addSubnode:self.newsImageNode];
```

Crash log
```
2018-05-06 22:05:35.306897+0800 Texture_Learning_1[73306:8769316] [Node] View controller <NewsViewController: 0x7fd15c10a520> will appear
2018-05-06 22:05:35.307878+0800 Texture_Learning_1[73306:8769316] [Collection] performBatchUpdates <ASTableNode: 0x7fd15c3060f0> {  }
2018-05-06 22:05:35.308007+0800 Texture_Learning_1[73306:8769316] [Collection] <ASTableView: 0x7fd15d819a00> Skipped update because load hasn't happened.
2018-05-06 22:05:35.308797+0800 Texture_Learning_1[73306:8769316] [Collection] Initial reloadData <ASTableNode: 0x7fd15c3060f0>
2018-05-06 22:05:35.309158+0800 Texture_Learning_1[73306:8769316] [Collection] New content: { itemCounts = [ <S0: 11> ] }
2018-05-06 22:05:35.325826+0800 Texture_Learning_1[73306:8769316] [Node] Propagating new traits for <NewsViewController: 0x7fd15c10a520>: { userInterfaceIdiom = Phone; containerSize = {414, 736}; horizontalSizeClass = Compact; verticalSizeClass = Regular; forceTouchCapability = Available }
2018-05-06 22:05:35.325995+0800 Texture_Learning_1[73306:8769316] [Node] Propagating new traits for <NewsViewController: 0x7fd15c10a520>: { userInterfaceIdiom = Phone; containerSize = {414, 736}; horizontalSizeClass = Compact; verticalSizeClass = Regular; forceTouchCapability = Available }
[...]

PR #1029, opened 2018-07-15
Create and set delegate for clip corner layers within ASDisplayNode

### LINKED ISSUE #1020: Crash on node deallocation when using cornerRoundingType == .clipping
Sample code is just a few lines, creating a plain ASDisplayNode, setting cornerRadius/ cornerRoundingType, and adding it to a view:
https://github.com/yxztj/cornerRoundingCrash

If you run the code directly, it would hit the assert in ASDisplayNode(CALayerDelegate) method "actionForLayer:forKey", as mentioned in #910. This looks strange because the 4 _clipCornerLayers are setting "self" as layerDelegate, but the delegate method needs self to be isLayerBacked == true(sometimes we need subnodes to receive touch events so we can't set isLayerBacked). Seems this delegate method isn't for the clipping layers?

And if I try to comment out the assert code(same effect as release build), or just return nil in the delegate method(shouldn't affect because the node is not layerBacked), the app crashes when we remove the node from parent view.

Call stack:
screen shot 2018-07-11 at 5 58 06 pm

Turning on Zombie Objects in build scheme and we get:
```
2018-07-11 17:49:13.871070+0800 cornerClippingCrash[19362:291616] *** -[ASDisplayNode isKindOfClass:]: message sent to deallocated instance 0x7fe735c05c40
```
Looks like UIKit is calling the clippingLayer's delegate(aka the node) for some work, but the node has already been deallocated in [ASDisplayView didMoveToSuperView:]
If we don't set the node as clippingLayer's delegate, the issue is gone. Is it possible not to set the delegate?
@nguyenhuy @appleguy

maicki:
Currently the delegate for all rounded corner layers are set to the owning display node. This has the effect that `actionForLayer:forKey:` is called on the node for every corner layer although it should have their own delegate. In `actionForLayer:forKey:` we assert that the node is layer backed. This is not necessary needed if using corner rounding though.
This diff will create an `ASDisplayNodeCornerLayerDelegate` and set it as delegate for the corner layers.

Fixes #1020, #910


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/befa1ca2320ad7752a9a09766f513821/raw/8fc02df5e5e9d3580fdc42735e6e3062d3130c25/log.txt)

nguyenhuy:
@maicki Did you run the test suite locally?

maicki:
@nguyenhuy Yeah but I would rather wait for the CI to pass tbh

nguyenhuy:
👍

Source/Private/ASDisplayNodeCornerLayerDelegate.m
```diff
@@ -0,0 +1,27 @@
...
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+
+#import "ASDisplayNodeCornerLayerDelegate.h"
+
+@implementation ASDisplayNodeCornerLayerDelegate
+
+- (id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event
+{
+  return nil;
```
Adlai-Holler: - Should we return `(id)kCFNull` to prevent the default action, like we do for other layers? That suppresses the default action (e.g. fade in).
- Is this different than setting no delegate on the layers?
- Let's share one single instance of this rather than creating one for each layer.

maicki: - Yes this should return `(id)kCFNull` I will change that.
- Yes setting no layer should animate the layer
- We are already sharing one instance if you look where we create the shared delegate for all of the corner layers within the code

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0324edaf4d9a769ae808e4048693315f/raw/46d6f96467b7d349a54ee02e399fe6cf856ed36d/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ec08d38b055fa6fb8d7ed7fd432e383a/raw/6844cdf78759ceccfe453b19a05ed8f07328bbf8/log.txt)

PR #1024, opened 2018-07-12
Stricter locking assertions

nguyenhuy:
- Rename `ASDisplayNodeAssertLockUnownedByCurrentThread` to `ASAssertUnlocked`, and `ASDisplayNodeAssertLockOwnedByCurrentThread` to `ASAssertLocked` -> shorter and hopefully easier to distinguish between the two.
- Add assertions to `_locked_` and `_u_` (i.e "unlocked") methods.
- Turn `CHECK_LOCKING_SAFETY` on by default. After #1022 and #1023, we're in a good shape to actually enforce locked/unlocked requirements of internal methods. Our test suite passed, and we'll test more at Pinterest after the sync this week.
- Fix ASVideoNode to avoid calling `play` while holding the lock. That method inserts a subnode and must be called lock free.
- Other minor changes.

PR #1032, opened 2018-07-16
Turn lock-checking on only when assertions are enabled #trivial

Adlai-Holler:
We accidentally landed this being set to 1 for all builds a few days ago (#1024).


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e784281636b292b6ed4155dc1eb4d138/raw/1d736e2dbdc0db323a9c86f878ed4a5f489e6283/log.txt)

PR #1033, opened 2018-07-16
Turn off exceptions to reduce binary size (-600KB for arm64)

Adlai-Holler:
We don't currently use exceptions so we don't need the extra binary to support them.

`-fno-exceptions` removes unwind tables and makes C++ STL put "abort" wherever there would be an exception.
`-fno-objc-arc-exceptions` removes code that ARC adds to prevent leaks in the case of exceptions https://clang.llvm.org/docs/AutomaticReferenceCounting.html#exceptions

**How big is the win?**
- In terms of total o-file size, the framework goes from **33.3MB to 31.5MB** for ARM64 release builds.
- In terms of executable size, the CatDealsCollectionView ARM64 release binary goes from **12.3MB to 11.7MB**.

So modulo any Apple fiddling with your binary, the 64-bit one should shrink by about 600KB.

### DIFF
--- a/Source/ASDisplayNodeExtras.h
+++ b/Source/ASDisplayNodeExtras.h
@@ -30,7 +30,7 @@
   #define ASSetDebugName(node, format, ...) node.debugName = [NSString stringWithFormat:format, __VA_ARGS__]
   #define ASSetDebugNames(...) _ASSetDebugNames(self.class, @"" # __VA_ARGS__, __VA_ARGS__, nil)
 #else
-  #define ASSetDebugName(node, name)
+  #define ASSetDebugName(node, format, ...)
   #define ASSetDebugNames(...)
 #endif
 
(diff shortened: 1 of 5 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/14f4e95e612625f37216b15407407d87/raw/3609a7141046c13bc75c5c96611e8378adb338ef/log.txt)

Adlai-Holler:
The CI is just plain broken =/

Source/ASDisplayNodeExtras.h
```diff
@@ -30,7 +30,7 @@
   #define ASSetDebugName(node, format, ...) node.debugName = [NSString stringWithFormat:format, __VA_ARGS__]
   #define ASSetDebugNames(...) _ASSetDebugNames(self.class, @"" # __VA_ARGS__, __VA_ARGS__, nil)
 #else
-  #define ASSetDebugName(node, name)
+  #define ASSetDebugName(node, format, ...)
```
Adlai-Holler: This was breaking release builds. I guess nobody uses this macro!!

PR #1036, opened 2018-07-17
Fix warning for ASLayout method override for the designated initializer of the superclass '-init' not found #trivial

maicki:


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/53c984939bb8e74543b5d90d89e143bb/raw/26c2dcb2b5e74ba82b371539cb2f2f58378f0eb4/log.txt)

Source/Layout/ASLayout.mm
```diff
@@ -137,6 +137,12 @@ - (instancetype)initWithLayoutElement:(id<ASLayoutElement>)layoutElement
   return self;
 }
 
+- (instancetype)init
+{
+  ASDisplayNodeAssert(NO, @"Use the designated initializer");
+  return [self init];
+}
```
Adlai-Holler: Rather than doing this, let's move the declaration for this out of the "Unavailable" category in the header file, into the primary class interface. That'll get rid of the warning.

maicki: Yeah decided between that and doing this. We do the same I put up in `ASLayoutTransition`I will change it in there too.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/292c72774264e7cb1f58c10d2fd318cb/raw/77c4783fb1755e01253db4fa49b2039857eec60c/log.txt)

PR #1042, opened 2018-07-19
Improve locking situation in ASVideoPlayerNode

maicki:
In ASVideoPlayerNode we are currently adding and removing subnodes with lock held. That's not allowed.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f7f7c7a41aa44c3f7475e41fec2ee2d9/raw/e442d3131c6207082cddce42bab594bc63d5605d/log.txt)

Source/ASVideoPlayerNode.mm
```diff
@@ -644,9 +671,11 @@ - (void)showSpinner
     }];
     
     _spinnerNode.style.preferredSize = CGSizeMake(44.0, 44.0);
-
-    [self addSubnode:_spinnerNode];
```
maicki: The spinner can be remove and added on different threads we may should grab the `_spinnerNode` before unlocking and adding it.

Source/ASVideoPlayerNode.mm
```diff
@@ -658,7 +687,10 @@ - (void)removeSpinner
   if (!_spinnerNode) {
     return;
   }
-  [_spinnerNode removeFromSupernode];
+  {
```
maicki: We may should grab the `_spinnerNode` here before unlocking

Source/ASVideoPlayerNode.mm
```diff
@@ -284,15 +282,23 @@ - (void)createControls
 
     if (_delegateFlags.delegateCustomControls && _delegateFlags.delegateLayoutSpecForControls) {
       NSDictionary *customControls = [_delegate videoPlayerNodeCustomControls:self];
```
nguyenhuy: In general we should avoid calling delegates with the lock held. May not worth address in this PR though.

Source/ASVideoPlayerNode.mm
```diff
@@ -284,15 +282,23 @@ - (void)createControls
...
+        subnodes.push_back(node);
         [_cachedControls setObject:node forKey:key];
       }
+      
+      {
+        ASUnlockScope(self);
+        for (var subnode : subnodes) {
+          [self addSubnode:subnode];
+        }
+      }
```
nguyenhuy: We can avoid a pair of lock and unlock calls if we add subnodes outside of the outmost lock scope.

Source/ASVideoPlayerNode.mm
```diff
@@ -315,14 +321,22 @@ - (NSArray *)createDefaultControlElementArray
...
+
+  {
+    ASUnlockScope(self);
+    for (ASDisplayNode *node in [cachedControls objectEnumerator]) {
+      [node removeFromSupernode];
+    }
   }
 
-  [self cleanCachedControls];
+  [self __locked_cleanCachedControls];
```
nguyenhuy: I think it's safe to do this before calling `removeFromSupernode` on cached controls? If yes we can avoid the unlock scope.

Source/ASVideoPlayerNode.mm
```diff
@@ -315,14 +321,22 @@ - (NSArray *)createDefaultControlElementArray
...
 }
 
-- (void)cleanCachedControls
+- (void)__locked_cleanCachedControls
```
nguyenhuy: ha!

Source/ASVideoPlayerNode.mm
nguyenhuy: Let's do this at the end so we don't need to re-acquire the lock.

Source/ASVideoPlayerNode.mm
nguyenhuy: Also, should we use 1 underscore at the beginning of this method's name (i.e `-_locked_cleanCachedControls`)?

Source/ASVideoPlayerNode.mm
nguyenhuy: Talked to @maicki offline. Let's leave this as is since my suggested approach is a bit more complicated and in case of early return(s), the vector will not be used.

PR #1044, opened 2018-07-23
Add documentation for rounding corners within Texture #trivial

maicki:


maicki:
@Adlai-Holler That should be just a Github issue. I ran the docs locally via Jekyll and there it looked fine.

PR #1054, opened 2018-07-29
Add NS_DESIGNATED_INITIALIZER to ASViewController initWithNode:

maicki:
This may introduce an API change in the `ASViewController` and changes in client code is needed but I will put that up for now.

### DIFF
--- a/Source/ASViewController.h
+++ b/Source/ASViewController.h
@@ -44,7 +44,7 @@ typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitWindowSizeBlock)(C
  *
  * @see ASVisibilityDepth
  */
-- (instancetype)initWithNode:(DisplayNodeType)node;
+- (instancetype)initWithNode:(DisplayNodeType)node NS_DESIGNATED_INITIALIZER;
 
 NS_ASSUME_NONNULL_END
 


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hashemp206:
@maicki can you please explain why you have changed the designated initializer? now in the whole project I should replace  `MyViewController()` with `MyViewController(node: nil)` and I should remember not to call empty initializer else my custom init won't work!

maicki:
@hashemp206 Hey - what is your use case using an `ASViewController` subclass without initializing it via a node? ... Instead you can just use a plain `UIViewController` in this case. It's kind of a side effect that `ASViewController` works as an `UIViewController` if you initializer it via a nil node. In fact the `initWithNode:` initializer is declared as assume nonnull, so it should not be even possible.

hashemp206:
@maicki I'm encapsulating node creation in `init` method 
like this:
``` swift
// MARK: Object lifecycle
    override init(node: ASDisplayNode? = nil)
    {
        let displayNode = ASDisplayNode()
        displayNode.backgroundColor = .backgroundColor
        
        super.init(node: displayNode)
        displayNode.addSubnode(collectionNode)
        hidesBottomBarWhenPushed = true
        setup()
    }
```

I prefer initiation with empty parameter than passing nil node, because I may forgot to pass nil node, and the init method won't called! I don't create node outside of ViewController and pass it to the ViewController. display node creation is part of ViewController responsibility.

PR #1059, opened 2018-08-01
[ASTextNode2] Simplify allocWithZone: + initialize implementation #trivial

Adlai-Holler:
### Motivation

Some code paths in our app use `class_createInstance` directly on subclasses of ASTextNode and so bypass `allocWithZone:`. In order to get the correct class hierarchy, the change must be made somewhere else.

### Changes

- Move the class change for subclasses from `allocWithZone:` to `initialize`.

We get the added benefit of improving performance.

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -1346,36 +1346,29 @@ static NSAttributedString *DefaultTruncationAttributedString()
 }
 #endif
 
-+ (id)allocWithZone:(struct _NSZone *)zone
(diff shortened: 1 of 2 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/54c0180901190be799a0afd65fd36172/raw/64d2a653fb6f316935f56998439162dd15c59bd3/log.txt)

PR #1065, opened 2018-08-03
Reduce copying in ASTextNode2 stack

Adlai-Holler:
Make text rendering faster.
- Remove pointless attributed string mutable copy in ASTextLayout. No clue why this was here but it is not useful in the least.
- Provide an option to make ASTextContainers immutable and get free copying. ASTextLayout copies its container, which is often already a copy so we save that.

### DIFF
--- a/Source/Private/TextExperiment/Component/ASTextLayout.m
+++ b/Source/Private/TextExperiment/Component/ASTextLayout.m
@@ -424,7 +424,7 @@ dispatch_semaphore_signal(_lock);
   container = [container copy];
   if (!text || !container) return nil;
   if (range.location + range.length > text.length) return nil;
-  [container setReadonly];
+  [container makeImmutable];
   maximumNumberOfRows = container.maximumNumberOfRows;
   
   // It may use larger constraint size when create CTFrame with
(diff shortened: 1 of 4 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/91734a863de7f63ab058846a7af5ecf2/raw/d9fb5cbcdd8eb523288bcaebdcd2299ef193839a/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/098181f09b6a6bbddf4dc3948218629c/raw/432abdc150e359d44d8a846a4ad666c469571290/log.txt)

Source/Private/TextExperiment/Component/ASTextLayout.m
```diff
@@ -134,26 +135,36 @@ - (instancetype)init {
   return self;
 }
 
-- (id)copyWithZone:(NSZone *)zone {
-  ASTextContainer *one = [self.class new];
+- (id)copyForced:(BOOL)forceCopy
+{
   dispatch_semaphore_wait(_lock, DISPATCH_TIME_FOREVER);
```
maicki: Curious why we use a semaphore as lock in this class?

Adlai-Holler: It was inherited from YYText where he did that. I actually really like it for small lock scopes. It's a lot faster than pthread mutex and there is a risk of priority inversion but it's much smaller since the waiting thread sleeps instead of spinning.

PR #1069, opened 2018-08-08
Refactored `accessibleElements` to `accessibilityElements`

jiawernlim:
Refactored `accessibleElements` to `accessibilityElements` and removed the re-definition of the property in ASDisplayView.

With this refactor, the field can now be used as a single access point into the accessibility elements of a view. Also, removing the re-definition of the property in _ASDisplayViewAccessibility.h enables clients to make use of the field and its associated helper methods directly from the `UIAccessibilityContainer` API rather than having Texture roll its own implementation.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -252,43 +252,26 @@ @implementation _ASDisplayView (UIAccessibilityContainer)
...
   
-  NSMutableArray *accessibleElements = [[NSMutableArray alloc] init];
-  CollectAccessibilityElementsForView(self, accessibleElements);
-  SortAccessibilityElements(accessibleElements);
-  _accessibleElements = accessibleElements;
-  
-  return _accessibleElements;
-}
-
-- (NSInteger)accessibilityElementCount
```
maicki: Hey! Are you sure we can remove this accessors? I created a sample project and tried to verify that, but all of these accessors do not call through to the `accessibilityElements` getter. I also looked into Hopper and the default implementation just returns 0 for e.g. `accessibilityElementCount`.

screen shot 2018-08-09 at 7 28 56 am

jiawernlim: Hi @maicki , thanks for checking! I wrote a small unit test to test the accessors, and they appear to work fine. I've added those tests here. Could you cross-reference with your sample project and see if there's anything I'm missing in the tests?

maicki: @jiawernlim I just ran your test and it seems to be failing:
screen shot 2018-08-09 at 7 08 24 pm

Here is also the example project that shows the behavior I said before:
[iOSPlayground.zip](https://github.com/TextureGroup/Texture/files/2276354/iOSPlayground.zip)

Also the part of the Apple documentation that describes the `UIAccessibilityContainer` API:
https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/iPhoneAccessibility/Making_Application_Accessible/Making_Application_Accessible.html#//apple_ref/doc/uid/TP40008785-CH102-SW10

Adlai-Holler:
@maicki Would you mind unblocking the CI build for Lim?

Source/Details/_ASDisplayViewAccessiblity.mm
Adlai-Holler: @maicki Lim and I have found some pretty crazy behavior here.

If accessibility is enabled, say by turning on the accessibility inspector, the system links the private `UIAccessibility` framework, which overrides these methods to be based on `accessibilityElements`:

screen shot 2018-08-10 at 10 27 22 am

So basically, you get 0 if no accessibility and otherwise you get the array.

And, these implementations are faster than the naïve implementations of `accessibilityElementAtIndex:` and friends, which we can't get at if we override them. So there's that.

Also, the documentation on `-accessibilityElements` says explicitly `This can be used as an alternative to implementing the dynamic methods.` implying that it's OK to do either.

So if we remove them, we get better performance but people calling them directly when accessibility is disabled will get 0.

One crazy option is to use `#if DEBUG` and if someone calls them when accessibility is disabled (could detect by using the public functions or by searching the runtime for the private `NSObjectAccessibility` accessibility class) then we make an assertion. That gets us the production performance and some safety. Thoughts?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e6847b667d27b9872f84bb8091d1d49f/raw/b00c5515310780d17abb91c2537c3e17255f3093/log.txt)

Source/Details/_ASDisplayViewAccessiblity.mm
maicki: Crazy town! Based on that let's move forward with the diff. Less code in this area is better in the first place.

We should put some safety in though. We could use `UIAccessibilityIsVoiceOverRunning` for detecting (not a fan of the private class) and throw an assertion in this place.

maicki:
@Adlai-Holler Based on the investigation within 
https://github.com/TextureGroup/Texture/pull/1069#discussion_r209332719 the test will fail on CI though as there is not accessibility enabled.

Source/Details/_ASDisplayViewAccessiblity.mm
jiawernlim: Thanks for the suggestion @maicki ! However, it appears `UIAccessibilityIsVoiceOverRunning` returns `NO` even though the accessibility inspector is on. When running your little example project with the accessibility inspector on, `displayView.accessibilityElementCount` still returns 2 while `UIAccessibilityIsVoiceOverRunning` returns NO. 

Do you have any other suggestions on how to better detect if it's running with accessibility enabled?

jiawernlim: Hello @maicki, just checking in to see if you have any suggestions on better ways to determine if accessibility is on. We could definitely just proceed with using `UIAccessibilityIsVoiceOverRunning`, but as mentioned earlier it will still fail when developers have the accessibility inspector on but voiceover off. We could also go with the option that Adlai had suggested, which was to search for the private class, but I agree that's not an elegant solution either. 

What are your thoughts?

jiawernlim: Hi @appleguy, the problem is that those unit tests will fail if the accessibility inspector is *not* on, and as such it'd fail on most systems by default. There does not seem to be an elegant way to test if the inspector is on, nor is there any way I have been able to find to programmatically turn on the inspector. 

That unit test could be removed, though then we would have no guarantees that they'd continue to work in the future. If we take this route, then perhaps we could add warnings in the header file "_ASDisplayViewAccessibility.h" to recommend use of `accessibilityElements` exclusively, and also mention that no guarantees are offered on the older accessors e.g. `accessibilityElementCount`. 

That being said, the [UIKit docs](https://developer.apple.com/documentation/objectivec/nsobject/1615147-accessibilityelements) quite specifically state that "Containers can implement this (`accessibilityElements`) property instead of the dynamic methods to support the retrieval of the contained elements." So realistically, the correctness of the older accessors should not be an issue as long as accessibility is enabled. Hence, I personally don't think this should be an issue even without having the unit tests, especially if we added some kind of warning in the header file.

AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1255,6 +1257,8 @@
 		058D09C5195D04C000B7D73C /* Tests */ = {
 			isa = PBXGroup;
 			children = (
+				F3F698D1211CAD4600800CB1 /* ASDisplayViewAccessibilityTests.mm */,
+				CC35CEC520DD87280006448D /* ASCollectionsTests.m */,
```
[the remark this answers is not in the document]
jiawernlim: I assume you're referring to line 1261? I checked and it looks like "ASCollectionsTests.m" was not in the `children()` list before, so yup it's probably Xcode that fixed some kind of state issue.

Source/Details/_ASDisplayViewAccessiblity.mm
maicki: @jiawernlim Sorry didn't see your messages a couple of days ago, but I would say let's target get this PR landed :)

I would suggest let's comment out the unit tests and add a comment why they are commented out and add some kind of warning in the header file (as you mentioned the docs for `accessibilityElements` we should be relatively save there) and we should be good to go.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0432a4858d03c5e25c2e259d84d661cd/raw/6e798e56f7c3ff75046af9523eb1772587ff9d76/log.txt)

maicki:
@jiawernlim LGTM let's get it in.

PR #1079, opened 2018-08-21
Fix compiler error in ASLocking #trivial

nguyenhuy:
Hitting this error because we're compiling with a stricter set of build rules internally.

nslockset

### DIFF
--- a/Source/ASLocking.h
+++ b/Source/ASLocking.h
@@ -107,7 +107,7 @@ NS_INLINE void ASUnlockSet(ASLockSet *lockSet) {
  */
 NS_INLINE ASLockSet ASLockSequence(NS_NOESCAPE ASLockSequenceBlock body)
 {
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/a9c4a5c48a1ee5b1f96f8045f28697e9/raw/cead724cab7c540929a47070b462334fca5130e1/log.txt)

nguyenhuy:
Good point, updated the diff. The compiler complaints about missing initialization for the lock set’s locks array (i.e locks.locks).

nguyenhuy:
Looks like we do, but the engineer who's working on this project hasn't hit it yet (and I heard it's non-trivial for me to apply all of her changes). Will follow-up if need to.

nguyenhuy:
Thanks btw!

PR #1081, opened 2018-08-22
Fix collection editing

wsdwsd0829:
It's used to fix cell editing bug for iOS 9 & 10 where UICV moveItemToIndexPath will be called every time cell is dragged through other cells (before dataSource's collectionNode:moveItemAtIndexPath... is called). For these cases we should call super code to handle the cell swapping internally in UICV. (moveItemToIndexPath for iOS 11+ will not be called though, so the reasoning is that the code after iOS11 have abstracted the method call to internal without triggering through public APIs).
ASCollectionView example app's reordering cell is fixed by this PR.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASCollectionView.mm
```diff
@@ -1021,9 +1027,28 @@ - (void)reloadItemsAtIndexPaths:(NSArray *)indexPaths
 - (void)moveItemAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)newIndexPath
 {
   ASDisplayNodeAssertMainThread();
-  [self performBatchUpdates:^{
-    [_changeSet moveItemAtIndexPath:indexPath toIndexPath:newIndexPath animationOptions:kASCollectionViewAnimationNone];
-  } completion:nil];
+  if (!_editing) {
+    [self performBatchUpdates:^{
```
appleguy: Should performBatchUpdates itself change behavior when interactiveMovement is in progress?

Imagine a reorder starts at the same time that a new page of content comes in, creating an insertion. I'm not sure what's actually supposed to happen there, but maybe we would want to just have performBatchUpdates internally run the batch operations without actually calling performBatchUpdates on the parent view?

That said, doing this only for moves right now seems reasonable if there is not a clear answer to this other question. Ideally in the case of the insertion, we at least know the UI will end up in a good state and not crash, even if it forces the user out of interactive mode.

Source/ASCollectionView.mm
```diff
@@ -147,6 +147,12 @@ @interface ASCollectionView () <ASRangeControllerDataSource, ASRangeControllerDe
    * (0 sections) we always check at least once after each update (initial reload is the first update.)
    */
   BOOL _hasEverCheckedForBatchFetchingDueToUpdate;
+
+  /**
+   * Set during beginInteractiveMovementForItemAtIndexPath and UIGestureRecognizerStateEnded
+   * (or UIGestureRecognizerStateFailed, UIGestureRecognizerStateCancelled.
+   */
+  BOOL _editing;
```
appleguy: Let's call this reordering, or maybe _interactiveMovementInProgress ?

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASCollectionView.mm
wsdwsd0829: I think this method is called on main thread and should not be happening at same time? One case should be blocked waiting to the other to finish?

Source/ASCollectionView.mm
```diff
@@ -1021,9 +1027,28 @@ - (void)reloadItemsAtIndexPaths:(NSArray *)indexPaths
...
+      [_changeSet moveItemAtIndexPath:indexPath toIndexPath:newIndexPath animationOptions:kASCollectionViewAnimationNone];
+    } completion:nil];
+  } else {
+    [super moveItemAtIndexPath:indexPath toIndexPath:newIndexPath];
+  }
+}
+
+- (BOOL)beginInteractiveMovementForItemAtIndexPath:(NSIndexPath *)indexPath {
+  _reordering = YES;
+  return [super beginInteractiveMovementForItemAtIndexPath:indexPath];
```
Adlai-Holler: If super returns `NO` here, I think it's appropriate for us not to set our flag. If the user notices that interactive movement was refused, he may not ever end the movement.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

maicki:
@wsdwsd0829 Would you please add an entry to the `CHANGELOG.md` and than let's land it!

wsdwsd0829:
done, Thanks for reminding.

PR #1086, opened 2018-08-28
Layer backed nodes thrash issue

wsdwsd0829:
Layer backed nodes will get following calls very often (thrashing) during UIKit operations like reloadItemAtIndexPath: or pop a view top view controller will be thrashed!!!):

- (id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event
{
  if (event == kCAOnOrderIn) {
    NSLog(@"&&&& entering: %@", self);
    [self __enterHierarchy];
  } else if (event == kCAOnOrderOut) {
    NSLog(@"&&&& exiting: %@", self);
    [self __exitHierarchy];
  }
  ASDisplayNodeAssert(_flags.layerBacked, @"We shouldn't get called back here unless we are layer-backed.");
  return (id)kCFNull;
}

Prob1: Even though interface coalescing will solve didEnterVisible Calls correctly for views, it will not for isLayerBacked nodes. 
Prob2: __exitHierarchy will  [self._locked_asyncLayer c ancelAsyncDisplay]; causing some layerBacked view’s not updated properly during thrashing.

Sol for 1; during “exiting”, layer will exit visible interface state no matter what, we should update their interface state when entering hierarchy. Since rangedMangedNodes will enter top down,  a non-layerbacked ancestor’s pendingInterfaceState seems to be reasonable candidate. 
Sol for 2: Do not cancel display for layerBackedNodes because they must be descendant of some view backed nodes which will be cancelled correctly as before.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

wsdwsd0829:
The fix is not ideal, may better just remove [self._locked_asyncLayer cancelAsyncDisplay];? (it seems to used before interface state is introduced)

wsdwsd0829:
Will fix them separately with better approaches.

PR #1087, opened 2018-08-29
Unlock before cleanup and calling out to subclass hooks for animated images.

maicki:
- Unlock before cleanup
- Unlock before calling out to `animatedImageSet:previousAnimatedImage:` to prevent a deadlock.

### DIFF
--- a/Source/ASImageNode+AnimatedImage.mm
+++ b/Source/ASImageNode+AnimatedImage.mm
@@ -44,12 +44,12 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
 - (void)_locked_setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
 {
   ASAssertLocked(__instanceLock__);
-  
+
   if (ASObjectIsEqual(_animatedImage, animatedImage) && (animatedImage == nil || animatedImage.playbackReady)) {
     return;
   }
   
-  id <ASAnimatedImageProtocol> previousAnimatedImage = _animatedImage;
+  __block id <ASAnimatedImageProtocol> previousAnimatedImage = _animatedImage;
   
   _animatedImage = animatedImage;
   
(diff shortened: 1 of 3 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASImageNode+AnimatedImage.mm
```diff
@@ -70,13 +71,20 @@ - (void)_locked_setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
...
-      self.contents = nil;
-      [self setCoverImage:nil];
+    // Clean up after ourselves.
+    shouldCleanup = YES;
   }
   
-  [self animatedImageSet:_animatedImage previousAnimatedImage:previousAnimatedImage];
+  {
+    // Unlock for cleanup and calling subclass hook
+    ASUnlockScope(self);
```
garrettmoon: Is this actually safe? We don't necessarily know what the calling scope is doing and therefore unlocking here could be dangerous? I know it's very unlikely (I can only think of convoluted examples)…

I'm not sure of a better solution though… If we do end up deciding this, we should at least add a comment that unlocking a lock someone else locked might cause us a bad time in the future.

Source/ASImageNode+AnimatedImage.mm
```diff
@@ -70,13 +71,20 @@ - (void)_locked_setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
...
     
+    self.contents = nil;
+    [self setCoverImage:nil];
```
nguyenhuy: Shouldn't we wrap the above lines in a `shouldCleanup` block?

Source/ASImageNode.h
```diff
@@ -173,7 +173,6 @@ typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
  *
  * @discussion This method is for subclasses to override so they can know if an animated image
  * has been set on the node.
- * @warning this method is called with the node's lock held.
```
nguyenhuy: Overall, I think it's great to remove this caveat.

Source/ASImageNode+AnimatedImage.mm
nguyenhuy: Agree. Another thing is that with a recursive lock, the lock may be acquired multiple times and so unlocking once is not enought.

Ideally, we have to either make this method `_u_` but it can be painful and may not be possible.

Another idea is dispatching to call `-animatedImageSet:previousAnimatedImage:` and the deallocation of previous image.

Source/ASImageNode+AnimatedImage.mm
nguyenhuy: Hm, can you explain why `-setCoverImage:` needs to be called unlocked? 

Somewhat unrelated: that method acquires the lock right away, so shouldn't we just call `-_locked_setCoverImage:` directly here?

maicki:
Unfortunately this is all not so trivial to solve. If we would like to solve this without dispatching to the next runloop it will take a lot of effort to refactor every (AS) image node and the fact we have to call a delegate subclass hook needs to be considered every time we change the code.

After looking into it more deeply I think there are a couple of ways we could solve that:
1. Refactor ASImageNode / ASNetworkImageNode to add some way to call the subclass hook without the lock held. As mentioned above this would need some massive refactor and also always some knowledge needed so calling the subclass hook will be called eventually.
2. Create a internal serial queue that would call to the subclass method, like a delegate (or move the method out as an internal delegate method in the first place and dispatch the call to the queue)
3. Dispatch the subclass call to the main queue / thread (this would go against the current trend though as it seems we are trying to get rid of calling delegate / subclass hooks on main)
4. Dispatch the call to the current queue (dispatch_get_current_queue). dispatch_get_current_queue is deprecated at the moment and there are a lot of online material that definetly speak against it.
5. Use performSelector:afterDelay:... This would be one way, but as we cannot pass two objects in this methods, we have to create a helper method. This could be factored out into an NSObject category though to reusable schedule a block via `performSelector:`
6. Use CFRunLoopPerformBlock (There would also be -[NSRunLoop performBlock:] but that's iOS 10+). This would work too to schedule a block on the next run loop cycle and this without having to introduce any more infrastructure around it as we can scheduel a block.

I eventually went with version 6 for now you can see the history too for also implementing 5. Happy for any thoughts or better ideas cc @nguyenhuy @garrettmoon @appleguy @adlai-holler

Source/ASImageNode+AnimatedImage.mm
```diff
@@ -70,13 +70,23 @@ - (void)_locked_setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
...
+  // We have to schedule the block on the common modes otherwise the tracking mode will not be included and it will
+  // not fire e.g. while scrolling down
+  CFRunLoopPerformBlock(CFRunLoopGetCurrent(), kCFRunLoopCommonModes, ^(void) {
+    [self animatedImageSet:animatedImage previousAnimatedImage:previousAnimatedImage];
+  });
+  // Don't need to wakeup the runloop as the current is already running
+// CFRunLoopWakeUp(runLoop); // Should not be necessary
+
   // Animated image can take while to dealloc, do it off the main queue
   if (previousAnimatedImage != nil) {
```
nguyenhuy: I think we need to deallocate the previous image after calling out to subclass, otherwise it will not be deallocated in background if the deallocation occurs before the next run loop.

maicki:
@nguyenhuy You wanna give that another look please

Issue #1094, opened 2018-09-01
Crashes in ASAssertUnlocked(__instanceLock__)


hashemp206:
I'm using the latest commit from branch master, but `ASDisplayNode` crashes in `_didLoad` method

any idea?

> 'NSInternalInconsistencyException', reason: 'Expected __instanceLock__.locked() to be false.'

<img width="342" alt="screen shot 2018-09-01 at 8 37 05 am" src="https://user-images.githubusercontent.com/3206748/44942215-3b99dc80-adc2-11e8-9612-aba3c75e7650.png">

<img width="804" alt="screen shot 2018-09-01 at 8 35 55 am" src="https://user-images.githubusercontent.com/3206748/44942212-2755df80-adc2-11e8-8a06-6ac05f268206.png">

maicki:
@hashemp206 I'm looking into something similar at the moment and at least know why the stack trace looks like. Can you please share the the output if you do `bt all` in the console if the assert is hit. Thanks!

hashemp206:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x0000000209fb5a08 libobjc.A.dylib`objc_exception_throw
    frame #1: 0x000000020acfc130 CoreFoundation`+[NSException raise:format:arguments:] + 104
    frame #2: 0x000000020b7e3560 Foundation`-[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 112
  * frame #3: 0x000000010420506c AsyncDisplayKit`::-[ASDisplayNode _didLoad](self=0x00000001442a4f40, _cmd="_didLoad") at ASDisplayNode.mm:552
    frame #4: 0x0000000104205c24 AsyncDisplayKit`::-[ASDisplayNode view](self=0x00000001442a4f40, _cmd="view") at ASDisplayNode.mm:644
    frame #5: 0x000000010419ddd4 AsyncDisplayKit`::-[ASCollectionNode view](self=0x00000001442a4f40, _cmd="view") at ASCollectionNode.mm:233
    frame #6: 0x000000010419e164 AsyncDisplayKit`::-[ASCollectionNode rangeController](self=0x00000001442a4f40, _cmd="rangeController") at ASCollectionNode.mm:291
[...]

PR #1103, opened 2018-09-08
Remove necessity to use view to access rangeController in ASTableNode, ASCollectionNode

maicki:
While setting the interface state of a node in certain circumstances we hit the path that first grabs the instance lock and calls `-[ASCollectionNode clearContents]` while the view is not loaded. Within `clearContents` we need access to the rangeController from the view by calling self.view.rangeController what will trigger creating the view and the assert linked in the issue will be thrown as `didLoad` should not be called with the instance locked held.

For now I removed the need to use the view to access the `rangeController` what should fix the situation described above and linked here: #1094

Further todos should be that we improve the locking impelementation around this area. Currently `-[ASDisplayNode clearContents]` is accessing instance variables and therefore should be prefixed with `locked_`, but `-[ASCollectionNode clearContents]` is accessing the rangeController by calling self instead of using the instance variable. The follow up issue is here: #1104

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -30,9 +30,12 @@
 
 #pragma mark - _ASCollectionPendingState
 
-@interface _ASCollectionPendingState : NSObject
-@property (weak, nonatomic) id <ASCollectionDelegate>   delegate;
-@property (weak, nonatomic) id <ASCollectionDataSource> dataSource;
+@interface _ASCollectionPendingState : NSObject {
+@public
+  std::vector<std::vector<ASRangeTuningParameters>> _tuningParameters;
+}
+@property (nonatomic, weak) id <ASCollectionDelegate>   delegate;
+@property (nonatomic, weak) id <ASCollectionDataSource> dataSource;
 @property (nonatomic) UICollectionViewLayout *collectionViewLayout;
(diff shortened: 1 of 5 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

maicki:
@nguyenhuy @Adlai-Holler Would appreciate a review. Thanks!

Source/ASCollectionNode.mm
```diff
@@ -30,17 +30,20 @@
 
 #pragma mark - _ASCollectionPendingState
 
-@interface _ASCollectionPendingState : NSObject
-@property (weak, nonatomic) id <ASCollectionDelegate>   delegate;
-@property (weak, nonatomic) id <ASCollectionDataSource> dataSource;
+@interface _ASCollectionPendingState : NSObject {
+@public
+  std::vector<std::vector<ASRangeTuningParameters>> _tuningParameters;
```
Adlai-Holler: Not specific to this diff, but we should probably use `ASRangeTuningParameters[ASLayoutRangeModeCount][ASLayoutRangeTypeCount] _tuningParameters` here so we get static allocation.

PR #1122, opened 2018-09-17
Remove use of NSHashTable for interface state delegates #trivial

Adlai-Holler:
- An empty hash table takes 2 heap allocations at 288 bytes.
- 2 hash tables per node (original and copy for safe enumeration) is 576 bytes per node.
- ASDisplayNode is only 928 bytes, so these hash tables add 60% more weight.
- Most of these blocks are global blocks (no value capture, free retain/release) and some are stack blocks. No malloc blocks.

### DIFF
--- a/Source/Private/ASDisplayNodeInternal.h
+++ b/Source/Private/ASDisplayNodeInternal.h
@@ -75,7 +75,6 @@ AS_EXTERN NSString * const ASRenderingEngineDidDisplayNodesScheduledBeforeTimest
 #define TIME_DISPLAYNODE_OPS 0 // If you're using this information frequently, try: (DEBUG || PROFILE)
 
 #define NUM_CLIP_CORNER_LAYERS 4
-#define MAX_INTERFACE_STATE_DELEGATES 4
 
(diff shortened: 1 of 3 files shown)


Source/Private/ASDisplayNodeInternal.h
```diff
@@ -75,6 +75,7 @@ AS_EXTERN NSString * const ASRenderingEngineDidDisplayNodesScheduledBeforeTimest
 #define TIME_DISPLAYNODE_OPS 0 // If you're using this information frequently, try: (DEBUG || PROFILE)
 
 #define NUM_CLIP_CORNER_LAYERS 4
+#define MAX_INTERFACE_STATE_DELEGATES 4
```
nguyenhuy: Most nodes don't need more than 4 delegates, but there should be an API to increase it. That way users can do so without changing the framework.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/c703474171f317edeed9272cfd5341f0/raw/69a469f0061a788451deba4023eba481a59f2608/log.txt)

PR #1124, opened 2018-09-17
Add a method for setting preconfigured PINRemoteImageManager

ernestmama:
Add a method for setting preconfigured PINRemoteImageManager instead of using the self-created ASPINRemoteImageManager.

In this way, we can add custom configuration to the PINRemoteImageManager such as request configuration/url configuration.

### DIFF
--- a/Source/Details/ASPINRemoteImageDownloader.h
+++ b/Source/Details/ASPINRemoteImageDownloader.h
@@ -41,6 +41,15 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfiguration *)configuration;
 
 /**
+ * Sets a custom perconconfigured PINRemoteImageManager that will be used by @c ASNetworkImageNodes and @c ASMultiplexImageNodes
+ * while loading images off the network. This can be called at anytime. Once this is set, it will always be used with priority.
+ * If this is not set, it will call setSharedImageManagerWithConfiguration with nil.
+ *
+ * @param PINRemoteImageManager the preconfigured remote image manager that will be used by `sharedDownloader`
+ */
+- (void)setPreconfiguredPINRemoteImageManager:(PINRemoteImageManager *)preconfiguredPINRemoteImageManager;
+
+/**
  * The shared instance of a @c PINRemoteImageManager used by all @c ASPINRemoteImageDownloaders
  *
  * @discussion you can use this method to access the shared manager. This is useful to share a cache
--- a/Source/Details/ASPINRemoteImageDownloader.h
+++ b/Source/Details/ASPINRemoteImageDownloader.h
@@ -41,12 +41,12 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfiguration *)configuration;
 
 /**
- * Sets a custom perconconfigured PINRemoteImageManager that will be used by @c ASNetworkImageNodes and @c ASMultiplexImageNodes
+ * Sets a custom preconfigured PINRemoteImageManager that will be used by @c ASNetworkImageNodes and @c ASMultiplexImageNodes
  * while loading images off the network. This must be specified early in the application lifecycle before
- * `sharedDownloader` is accessed. If nil is passed in as the PINRemoteImageManager, it will call
- * setSharedImageManagerWithConfiguration with nil configuration.
+ * `sharedDownloader` is accessed. If nil is passed in as the PINRemoteImageManager, it will create
+ * a default image manager with a nil session configuration.
  *
- * @param PINRemoteImageManager the preconfigured remote image manager that will be used by `sharedDownloader`
+ * @param PINRemoteImageManager The preconfigured remote image manager that will be used by `sharedDownloader`
  */
 + (void)setSharedPreconfiguredRemoteImageManager:(nullable PINRemoteImageManager *)preconfiguredPINRemoteImageManager;
 
--- a/Source/Details/ASPINRemoteImageDownloader.h
+++ b/Source/Details/ASPINRemoteImageDownloader.h
@@ -42,12 +42,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  * Sets a custom perconconfigured PINRemoteImageManager that will be used by @c ASNetworkImageNodes and @c ASMultiplexImageNodes
- * while loading images off the network. This can be called at anytime. Once this is set, it will always be used with priority.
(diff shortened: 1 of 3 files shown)


Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -155,9 +159,18 @@ + (PINRemoteImageManager *)sharedPINRemoteImageManagerWithConfiguration:(NSURLSe
 
 - (PINRemoteImageManager *)sharedPINRemoteImageManager
 {
+  if (_preconfiguredPINRemoteImageManager) {
```
maicki: Is this thread safe? Do we have to lock this?

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -155,9 +159,18 @@ + (PINRemoteImageManager *)sharedPINRemoteImageManagerWithConfiguration:(NSURLSe
...
+    return _preconfiguredPINRemoteImageManager;
+  }
   return [ASPINRemoteImageDownloader sharedPINRemoteImageManagerWithConfiguration:nil];
 }
 
+- (PINRemoteImageManager *)setPreconfiguredPINRemoteImageManager:(PINRemoteImageManager *)preconfiguredPINRemoteImageManager
+{
+  _preconfiguredPINRemoteImageManager = preconfiguredPINRemoteImageManager;
```
maicki: I think we should add a lock here. Furthermore why do we return the same object here and od not make it a regular setter?

ernestmama: Actually, I was going to make it return a void since it is setting.

ernestmama: Let me make it thread safe.

nguyenhuy:
It's a valid concern that the new addition should be thread-safe. However, I'm not sure about adding a lock around `-sharedPINRemoteImageManager`. Almost every method of the downloader hits it eventually, and since the downloader is shared between all nodes, it'll be a very hot code path and there'll be lock contention.

How about we use the same pattern as `-setSharedImageManagerWithConfiguration:` and allow allow users to set the class of `sharedPINRemoteImageManager`? The class is required to be set very early on, even before `+setSharedImageManagerWithConfiguration:` which then will be picked up by `+sharedPINRemoteImageManagerWithConfiguration:`.

maicki:
The suggestion from @nguyenhuy sounds good to me.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/69ad8fab94a81ca6fb187ee6fb5cc50c/raw/d33cfcbdb33b8b740922ddf64c25aa7690eb4062/log.txt)

ernestmama:
@nguyenhuy Shouldn't Texture knows nothing about the custom image manager except it is a subclass of PINRemoteImageMager to keep the abstraction? The subclass is in a higher level app/library.

nguyenhuy:
@ernestmama Yes, and with the new API, it still wouldn't know anything, does it?

ernestmama:
How about this update? This will essentially change that the RemoteImageManager can only be set once (same as previous with configuration) instead of able to swapping it in runtime. It will need to be set before the first time ImageManager is being used.

Source/Details/ASPINRemoteImageDownloader.h
```diff
@@ -40,6 +40,16 @@ NS_ASSUME_NONNULL_BEGIN
  */
 + (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfiguration *)configuration;
 
+/**
+ * Sets a custom perconconfigured PINRemoteImageManager that will be used by @c ASNetworkImageNodes and @c ASMultiplexImageNodes
+ * while loading images off the network. This must be specified early in the application lifecycle before
+ * `sharedDownloader` is accessed. If nil is passed in as the PINRemoteImageManager, it will call
+ * setSharedImageManagerWithConfiguration with nil configuration.
```
nguyenhuy: > it will call setSharedImageManagerWithConfiguration with nil configuration.

Does it? How about:

> it will create a default image manager with a `nil` session configuration.

ernestmama: You are rrright.

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -119,15 +121,25 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
-  static ASPINRemoteImageManager *sharedPINRemoteImageManager;
-  static dispatch_once_t onceToken;
+  NSAssert(sharedDownloader == nil, @"Singleton has been created and session can no longer be configured.");
+  NSAssert1(sharedPINRemoteImageManager == nil, @"An instance of %@ has been set. Either configuration or preconfigured image manager can be set at a time and only once.", [[sharedPINRemoteImageManager class] description]);
+  __unused PINRemoteImageManager *sharedManager = [self sharedPINRemoteImageManagerWithConfiguration:nil preconfiguredPINRemoteImageManager:preconfiguredPINRemoteImageManager];
+}
+
++ (PINRemoteImageManager *)sharedPINRemoteImageManagerWithConfiguration:(NSURLSessionConfiguration *)configuration preconfiguredPINRemoteImageManager:(PINRemoteImageManager *)preconfiguredPINRemoteImageManager
+{
+  NSAssert(configuration != nil && preconfiguredPINRemoteImageManager != nil, @"Either configuration or preconfigured image manager can be set at a time.");
```
nguyenhuy: This assertion means both configuration and preconfigured manager must not be nil. I don't think it's right.

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -119,15 +121,25 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
   dispatch_once(&onceToken, ^{
 
+    if (preconfiguredPINRemoteImageManager) {
+      sharedPINRemoteImageManager = preconfiguredPINRemoteImageManager;
+    } else {
 #if PIN_ANIMATED_AVAILABLE
     // Check that Carthage users have linked both PINRemoteImage & PINCache by testing for one file each
     if (!(NSClassFromString(@"PINRemoteImageManager"))) {
```
nguyenhuy: The indentation seems to be off. The whole else body here should be indented.

PR #1131, opened 2018-09-19
Remove reliance on shared_ptr for ASDisplayNodeLayouts

Adlai-Holler:


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/75b71b528f74a0d2d6141ae6aa0722c7/raw/cbc7144a3112eb1485d787c75902aa7dbbaa4011/log.txt)

Source/ASDisplayNode+Layout.mm
```diff
@@ -654,10 +653,10 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
         
         // Update calculated layout
         let previousLayout = _calculatedDisplayNodeLayout;
-        let pendingLayout = std::make_shared<ASDisplayNodeLayout>(newLayout,
-                                                                  constrainedSize,
-                                                                  constrainedSize.max,
-                                                                  newLayoutVersion);
+        var pendingLayout = ASDisplayNodeLayout(newLayout,
```
maicki: Do you change the `pendingLayout` somewhere to change it to `var`?

Source/Private/ASDisplayNodeLayout.h
```diff
@@ -43,10 +43,16 @@ struct ASDisplayNodeLayout {
   /**
    * Returns whether this is valid for a given version
    */
-  BOOL isValid(NSUInteger version);
+  BOOL isValid(NSUInteger versionArg) {
```
maicki: You think it's worth it to inline that?

Source/Private/ASLayoutTransition.h
```diff
@@ -52,12 +50,12 @@ AS_SUBCLASSING_RESTRICTED
 /**
  * Previous layout to transition from
  */
-@property (nonatomic, readonly) std::shared_ptr<ASDisplayNodeLayout> previousLayout;
+@property (nonatomic, readonly) const ASDisplayNodeLayout &previousLayout;
```
maicki: Are we sure we are holding on that reference as the transition lifetime?

Source/ASDisplayNode+Layout.mm
Adlai-Holler: You're right – in a previous version, `setCalculatedLayout` mutated the input (to apply the unflattenedLayout thing) but I changed it to copy the input and never committed reverting this to `let`.

Source/Private/ASDisplayNodeLayout.h
Adlai-Holler: I figure there's no real downside. What do you think?

Source/Private/ASLayoutTransition.h
Adlai-Holler: It's an ivar on the transition object (it's copied in `-init`) so it'll live as long as the transition lives.

Adlai-Holler: There _is_ a small risk here, which is you read this ref and store it/don't copy it, and then you release the transition while you're holding on to the ref. Our current usage is OK. We could actually change this to be a pointer, and use NS_RETURNS_INNER_POINTER to prevent that … I might try it.

Source/Private/ASDisplayNodeLayout.h
maicki: Yep agree

Source/Private/ASLayoutTransition.h
maicki: Nice

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/11f2ee3995fb5f63a3f6b1eae8eba91b/raw/2c8bc871fd9e84436d4123f7764bdb28f53ee002/log.txt)

PR #1141, opened 2018-09-25
Fix crash setting attributed text on multiple threads

maicki:
Currently some race condition can happen within `-[ASTextNode setAttributedText:]`:
- Thread A calls `setAttributedText:`
- We change the text with the lock held and we grab the length of the _attributedString
- After getting the length a thread switch happens
- A thread B calls setAttributedText:. The attributed text instance var get's updated and it finishes the method
- Thread switch happens back to A and it's going again. It goes into the if block and try to access the last character based on the old length that is not up to date anymore and it crashes

You can repro it if you simulate a thread switch and the crash on the descender line via a semaphore like: 

```
- (void)setAttributedText:(NSAttributedString *)attributedText semaphore:(dispatch_semaphore_t)semaphore
{
  if (attributedText == nil) {
    attributedText = [[NSAttributedString alloc] initWithString:@"" attributes:nil];
  }
  
  // Don't hold textLock for too long.
  {
    ASLockScopeSelf();
    if (ASObjectIsEqual(attributedText, _attributedText)) {
      return;
    }

    _attributedText = ASCleanseAttributedStringOfCoreTextAttributes(attributedText);
#if AS_TEXTNODE_RECORD_ATTRIBUTED_STRINGS
	  [ASTextNode _registerAttributedText:_attributedText];
#endif
  }
    
  // Since truncation text matches style of attributedText, invalidate it now.
  [self _invalidateTruncationText];
  
  NSUInteger length = _attributedText.length;
  if (length > 0) {
    if (semaphore) {
      // Wait until the sempahore is released
      dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
    }
    
    self.style.ascender = [[self class] ascenderWithAttributedString:_attributedText];
    self.style.descender = [[_attributedText attribute:NSFontAttributeName atIndex:length - 1 effectiveRange:NULL] descender];
  }

  // Tell the display node superclasses that the cached layout is incorrect now
  [self setNeedsLayout];

  // Force display to create renderer with new size and redisplay with new string
  [self setNeedsDisplay];
  
  
  // Accessiblity
  self.accessibilityLabel = _attributedText.string;
  self.isAccessibilityElement = (length != 0); // We're an accessibility element by default if there is a string.
}
```

And in the frontend with something like:

```
_textNode = [[ASTextNode alloc] init];

dispatch_semaphore_t sem = dispatch_semaphore_create(0);
dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1.0 * NSEC_PER_SEC)), dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
    // Set some text
        [self.textNode setAttributedText:[[NSAttributedString alloc] initWithString:@"Short" attributes:@{}] semaphore:nil];
    dispatch_semaphore_signal(sem);
});

// Set some textand let it wait on the semaphore
[_textNode setAttributedText:[[NSAttributedString alloc] initWithString:@"Longer longer text" attributes:@{}] semaphore:sem];
```

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -440,39 +440,46 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   if (attributedText == nil) {
     attributedText = [[NSAttributedString alloc] initWithString:@"" attributes:nil];
   }
-  
-  // Don't hold textLock for too long.
+
   {
     ASLockScopeSelf();
     if (ASObjectIsEqual(attributedText, _attributedText)) {
       return;
     }
 
-    _attributedText = ASCleanseAttributedStringOfCoreTextAttributes(attributedText);
-#if AS_TEXTNODE_RECORD_ATTRIBUTED_STRINGS
-	  [ASTextNode _registerAttributedText:_attributedText];
-#endif
-  }
-    
-  // Since truncation text matches style of attributedText, invalidate it now.
-  [self _invalidateTruncationText];
-  
-  NSUInteger length = _attributedText.length;
-  if (length > 0) {
-    self.style.ascender = [[self class] ascenderWithAttributedString:_attributedText];
-    self.style.descender = [[_attributedText attribute:NSFontAttributeName atIndex:length - 1 effectiveRange:NULL] descender];
-  }
+    NSAttributedString *cleanedAttributedString = ASCleanseAttributedStringOfCoreTextAttributes(attributedText);
 
+    // Invalidating the truncation text must be done while we still hold the lock. Because after we release it,
+    // another thread may set a new truncation text that will then be cleared by this thread, other may draw
+    // this soon-to-be-invalidated text.
+    [self _locked_invalidateTruncationText];
+
+    NSUInteger length = cleanedAttributedString.length;
+    if (length > 0) {
+      // Updating ascender and descender in one transaction while holding the lock.
+      ASLayoutElementStyle *style = [self _locked_style];
+      style.ascender = [[self class] ascenderWithAttributedString:cleanedAttributedString];
+      style.descender = [[attributedText attribute:NSFontAttributeName atIndex:cleanedAttributedString.length - 1 effectiveRange:NULL] descender];
+    }
+   
+    // Update attributed text with cleaned attributed string
+    _attributedText = cleanedAttributedString;
+  }
+  
   // Tell the display node superclasses that the cached layout is incorrect now
   [self setNeedsLayout];
 
   // Force display to create renderer with new size and redisplay with new string
   [self setNeedsDisplay];
   
-  
   // Accessiblity
-  self.accessibilityLabel = _attributedText.string;
-  self.isAccessibilityElement = (length != 0); // We're an accessibility element by default if there is a string.
+  let currentAttributedText = self.attributedText; // Grab attributed string again in case it changed in the meantime
+  self.accessibilityLabel = currentAttributedText.string;
+  self.isAccessibilityElement = (currentAttributedText.length != 0); // We're an accessibility element by default if there is a string.
+
+#if AS_TEXTNODE_RECORD_ATTRIBUTED_STRINGS
+  [ASTextNode _registerAttributedText:_attributedText];
+#endif
 }
 
 #pragma mark - Text Layout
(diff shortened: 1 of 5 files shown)


maicki:
@Adlai-Holler Any reason we are not using `ASCleanseAttributedStringOfCoreTextAttributes` in `ASTextNode2`?

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASTextNode.mm
```diff
@@ -436,43 +436,44 @@ - (NSAttributedString *)attributedText
...
       return;
     }
 
-    _attributedText = ASCleanseAttributedStringOfCoreTextAttributes(attributedText);
-#if AS_TEXTNODE_RECORD_ATTRIBUTED_STRINGS
-	  [ASTextNode _registerAttributedText:_attributedText];
-#endif
+    NSAttributedString *cleanedAttributedString = ASCleanseAttributedStringOfCoreTextAttributes(attributedText);
+
+    [self _locked_invalidateTruncationText];
```
nguyenhuy: ok, so invalidating the truncation text must be done while we still hold the lock. Because after we release it, another thread may set a new truncation text that will then be cleared by this thread, other may draw this soon-to-be-invalidated text.

Let's leave a comment/explaination here.

Source/ASTextNode.mm
```diff
@@ -436,43 +436,44 @@ - (NSAttributedString *)attributedText
...
+
+    NSUInteger length = cleanedAttributedString.length;
+    if (length > 0) {
+      ASLayoutElementStyle *style = [self _locked_style];
+      style.ascender = [[self class] ascenderWithAttributedString:cleanedAttributedString];
+      style.descender = [[attributedText attribute:NSFontAttributeName atIndex:cleanedAttributedString.length - 1 effectiveRange:NULL] descender];
```
nguyenhuy: Same deal with `ascender` and `descender` here. We basically use the lock as a transactional one.

Source/ASTextNode.mm
```diff
@@ -436,43 +436,44 @@ - (NSAttributedString *)attributedText
...
   [self setNeedsLayout];
 
   // Force display to create renderer with new size and redisplay with new string
   [self setNeedsDisplay];
   
-  
   // Accessiblity
   self.accessibilityLabel = _attributedText.string;
-  self.isAccessibilityElement = (length != 0); // We're an accessibility element by default if there is a string.
+  self.isAccessibilityElement = (_attributedText.length != 0); // We're an accessibility element by default if there is a string.
```
nguyenhuy: `_attributedText` is accessed without the lock here. Can we do these operations while having the locks as well?

At least, we need to access the text via its setter (`self.attributedText`) so that if it's changed by another thread right before this, we'll end up grabbing the updated text (and its length) here. Also, it's no longer safe to use the `attributedText` param of this method.

Source/ASTextNode.mm
```diff
@@ -1233,12 +1234,17 @@ - (void)_invalidateTruncationText
 {
   {
     ASLockScopeSelf();
-    _composedTruncationText = nil;
+    [self _locked_invalidateTruncationText];
   }
 
   [self setNeedsDisplay];
```
nguyenhuy: I feel like now is the time to let callers of this (unlocked) method to call `-setNeedsDisplay` themselves. It's a bit confusing to only automatically do it with the unlock version.

Source/ASTextNode2.mm
```diff
@@ -1061,10 +1062,15 @@ - (NSUInteger)lineCount
 - (void)_invalidateTruncationText
 {
   ASLockScopeSelf();
-  _textContainer.truncationToken = nil;
+  [self _locked_invalidateTruncationText];
   [self setNeedsDisplay];
```
nguyenhuy: Is it actually ok to call `-setNeedsDisplay` here?

maicki: I asked that myself too, but I didn't want to change it within this diff. In theory the code in `ASTextNode2` should be adjusted to the same as in `ASTextNode` after this lands.

maicki:
@nguyenhuy Addressed your comments

Source/ASTextNode2.mm
nguyenhuy: 👍

Issue #1151, opened 2018-09-29
[ASImageNode] Incorrect image scaling after precompositing rounded corners, leading to huge memory waste


yxztj:
I just created a repo for demo: https://github.com/yxztj/ASDKImageScaleIssue

If I have an image of 100px * 100px, the backing store(layer's contents) of ASImageNode should be 100*100 size. There's no problem with that.

However when setting ASImageNode's cornerRoundingType to precomposited, the layer's contents size will become 300*300(on a 3x device). This is due to the compositing process used layer's scale(3x) instead of image's scale(ASImageNode will always use 1x when generating image), therefore the resulting image is 100px * 3 * 100 px * 3, leading to 9x larger memory when rendering, and that's  how I found this issue :)

PR #1152, opened 2018-09-29
fix: incorrect backing image scale after precompositing rounded corners

yxztj:
Try to fix https://github.com/TextureGroup/Texture/issues/1151

In __didDisplayNodeContentWithRenderingContext with cornerRoundingType set to precomposited, the image after and before processing should share the same scale. So just use the input image's scale when creating context.

### DIFF
--- a/Source/Private/ASDisplayNode+AsyncDisplay.mm
+++ b/Source/Private/ASDisplayNode+AsyncDisplay.mm
@@ -323,7 +323,7 @@
       bounds.size.height *= contentsScale;
       CGFloat white = 0.0f, alpha = 0.0f;
       [backgroundColor getWhite:&white alpha:&alpha];
-      ASGraphicsBeginImageContextWithOptions(bounds.size, (alpha == 1.0f), contentsScale);
+      ASGraphicsBeginImageContextWithOptions(bounds.size, (alpha == 1.0f), (*image).scale);
       [*image drawInRect:bounds];
     } else {
       bounds = CGContextGetClipBoundingBox(context);


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

yxztj:
@maicki @appleguy This issue would lead to a 9x memory cost(on 3x devices) which quickly drain the whole available memory. Please help review if this fix works :)

Source/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -323,7 +323,7 @@ - (void)__didDisplayNodeContentWithRenderingContext:(CGContextRef)context image:
       bounds.size.height *= contentsScale;
       CGFloat white = 0.0f, alpha = 0.0f;
       [backgroundColor getWhite:&white alpha:&alpha];
-      ASGraphicsBeginImageContextWithOptions(bounds.size, (alpha == 1.0f), contentsScale);
+      ASGraphicsBeginImageContextWithOptions(bounds.size, (alpha == 1.0f), (*image).scale);
```
appleguy: Particularly for corner rounding, I don't think this is correct. For the corners to have crisp edges, the backing store has to be as high-res as the screen scale to capture the fine detail of the corner's curve. Otherwise the corners will be up-scaled when presented.

This does mean increased memory usage, in which case the Clip Corner mode may be a better fit (if this significantly affects your app).

yxztj: @appleguy Thanks for the information. The problem is that in ASImageNode, image(for example 100*100 with 3x) is drawn with contentsScale of 1(now we get 300px*300px), which comes as input of this function, and we draw the image again on a 3x context, results in a 900px*900px image, which consumes 9 times of memory as much as it should use. This is quite unbearable for large images in our app. Is it the only way to use clip corners?

yxztj: @appleguy Thanks for the information. I understand the backing store has to be as much pixels as there are on the screen, but the problem is that(please refer to #1151 and the demo project in it) we are creating a backing store with 9 times(on 3x device) pixels as there are actually on the screen, which is a huge waste of memory.

Per my understanding, the function is trying to precomposite rounded corners without modifying the amount of pixels which will be used as layer backing store. For example we would set a 300px*300px image to a 100*100 imageNode on a 3x device, for the backing image we either use an image with 3x scale/100point or a 1x scale/300point combination. There won't be any up-scaling issue since we've provided the exact amount of pixels needed on the screen.

But the code here is both multiplying size with scale and setting context scale to 3x, creating a 3x scale/300point context, which seems unreasonable.

PS: We've tried 2 possible solutions to resolve the issue:
1. changing context scale to image scale, resulting in a 1x/300point backing store
2. remove all contentsScale multiplication in this func, resulting in a 3x/100point backing store
Both work great, and we don't see sacrificing any image/corner rounding details even zooming in the screenshot of it.

yxztj: @appleguy Hope you have time review this, since it really saves a lot of memory. We've been applying the fix in our own fork since last year, everything looks great :)

yxztj:
@Adlai-Holler @nguyenhuy Please help if possible..

nguyenhuy:
@appleguy Can you take another look on this PR? Thanks!

PR #1136, opened 2018-09-19
Add ASExperimentalSkipClearData

maicki:
We are seeing crashes to an internal inconsistency deep within `UICollectionView`. We were able to track down the range of commits where we start seeing this crash. To verify the hunch of the root cause of this crash we will introduce a new experiment that will skip calling clear data if `dataSource` / `delegate` is cleared within `ASCollectionView`.

PR #1154, opened 2018-10-02
Only clear ASCollectionView's data during deallocation

nguyenhuy:
This is a follow up on #1136. Our experiment results show that clearing data frequently is the cause of our #1 crash. @maicki and I believe that this is because silently clearing its data while the collection view is still being used, without notifying the backing UICollectionView, puts it out-of-sync and causes mayhem next time the collection view processes a batch update. If you look at the stack trace closely, you'll notice that the crash doesn't occur on the same run loop that `clearData` was called, making it extremely tricky to investigate and identify the root cause.

To fix this problem, I'm proposing to only call `clearData` if the collection view is actually being torn down. I'm also going to set up another experiment to slowly roll this out to make sure we don't reintroduce the crash, as well as to get insights on whether it's actually worth it. My gut feeling is that it's not, since the data will be autimatically cleared out soon anyway. Please let me know your thoughts.

screen shot 2018-10-01 at 5 08 47 pm


maicki:
Will us further looking into fixing: #1155

PR #1156, opened 2018-10-02
Copy pending layout during transition lifecycle

mikezucc:
Before display, a node's layout tree is flattened in order to calculate display information such as position. In almost all cases, this happens early in the layout transition lifecycle (during `calculateLayoutThatFits:` ). However, there are special cases where this layout tree is not flattened until the very last opportunity so that debugging tools (Weaver) can inspect the layout tree.

With the move to use const references instead of shared_ptr's for the pending layout tree used in the layout transition, the flattened layout tree was accessible to changes in the pending stage of the transition, causing layout specs to sneak into the final step of the layout transition: moving subnodes.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -2211,36 +2209,6 @@
...
-			inputPaths = (
-			);
-			name = "[CP] Embed Pods Frameworks";
-			outputPaths = (
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-AsyncDisplayKitTests/Pods-AsyncDisplayKitTests-frameworks.sh\"\n";
-			showEnvVarsInLog = 0;
-		};
```
nguyenhuy: Let's revert changes in in this xcode file.

Source/ASDisplayNode+Layout.mm
```diff
@@ -1010,7 +1036,6 @@ - (void)_locked_setCalculatedDisplayNodeLayout:(const ASDisplayNodeLayout &)disp
   
   // Flatten the layout if it wasn't done before (@see -calculateLayoutThatFits:).
   if ([ASDisplayNode shouldStoreUnflattenedLayouts]) {
```
nguyenhuy: We should just remove the entire if block here since callers of this method are now responsible for flattening the layout if need to.

Source/ASDisplayNode+Layout.mm
```diff
@@ -663,6 +663,10 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
         _pendingLayoutTransition = pendingLayoutTransition = [[ASLayoutTransition alloc] initWithNode:self
                                                                                         pendingLayout:pendingLayout
                                                                                        previousLayout:previousLayout];
+        if ([ASDisplayNode shouldStoreUnflattenedLayouts]) {
```
nguyenhuy: One concern is that `newLayout` used [a few lines above](https://github.com/TextureGroup/Texture/pull/1156/files#diff-71ea742a29543cb7e08c3dffe2a2508eR656) can be unflattened, and it was set as the calculated layout. That's why we needed to flatten it in [the setter](https://github.com/TextureGroup/Texture/pull/1156/files#diff-71ea742a29543cb7e08c3dffe2a2508eR1039).

I think a better approach would be to just flatten that layout immediately and use it throughout this method. Like this

```
{
        // Grab __instanceLock__ here to make sure this transition isn't invalidated
        // right after it passed the validation test and before it proceeds
        ASDN::MutexLocker l(__instanceLock__);

        if ([ASDisplayNode shouldStoreUnflattenedLayouts]) {
          _unflattenedLayout = newLayout;
          newLayout = [newLayout filteredNodeLayoutTree];
        }
        
        // Update calculated layout
        let previousLayout = _calculatedDisplayNodeLayout;
        let pendingLayout = ASDisplayNodeLayout(newLayout,
                                                constrainedSize,
                                                constrainedSize.max,
                                                newLayoutVersion);
        [self _locked_setCalculatedDisplayNodeLayout:pendingLayout];
        
        // Setup pending layout transition for animation
        _pendingLayoutTransition = pendingLayoutTransition = [[ASLayoutTransition alloc] initWithNode:self
                                                                                        pendingLayout:pendingLayout
                                                                                       previousLayout:previousLayout];
        // Setup context for pending layout transition. we need to hold a strong reference to the context
        _pendingLayoutTransitionContext = pendingLayoutTransitionContext = [[_ASTransitionContext alloc] initWithAnimation:animated
                                                                                                            layoutDelegate:_pendingLayoutTransition
                                                                                                        completionDelegate:self];
      }
```

Source/ASDisplayNode+Layout.mm
```diff
@@ -883,6 +887,28 @@ - (void)_completePendingLayoutTransition
...
+/**
+ * In cases where a late layout tree flatten is performed, this is used to perform the transition copy and flatten.
+ * This is import to support const pointer usage instead of shared_ptr when managing layouts and transitions
+ * during the layout flatten/ transition lifecycle.
+ */
+- (ASLayoutTransition *)_locked_flattenLayoutTransitionIfNeeded:(ASLayoutTransition *)pendingLayoutTransition {
+  ASLayoutTransition *transition = pendingLayoutTransition;
+  if ([ASDisplayNode shouldStoreUnflattenedLayouts]) {
+    ASDisplayNodeLayout pendingLayout = pendingLayoutTransition.pendingLayout;
+    _unflattenedLayout = pendingLayout.layout;
```
nguyenhuy: The use case for this method isn't 100% perfect [here](https://github.com/TextureGroup/Texture/pull/1156/files#diff-71ea742a29543cb7e08c3dffe2a2508eR667), so it seems like it's only once [here](https://github.com/TextureGroup/Texture/pull/1156/files#diff-71ea742a29543cb7e08c3dffe2a2508eR879). What do you think about doing everything here inline?

AsyncDisplayKit.xcodeproj/project.pbxproj
mikezucc: Right, wasn't sure if cocoapods artifacts changes should be ingested

Source/ASDisplayNode+Layout.mm
mikezucc: Yes that sounds like a good idea, I would prefer to group everything together for clarity but there are no truly common cases.

PR #1157, opened 2018-10-02
Clean up timing of layout tree flattening/ copying of unflattened tree for Weaver

mikezucc:
Before display, a node's layout tree is flattened in order to calculate display information such as position. In almost all cases, this happens early in the layout transition lifecycle (during `calculateLayoutThatFits:` ). However, there are special cases where this layout tree is not flattened until the very last opportunity so that debugging tools (Weaver) can inspect the layout tree.

With the move to use const references instead of shared_ptr's for the pending layout tree used in the layout transition, the flattened layout tree was accessible to changes in the pending stage of the transition, causing layout specs to sneak into the final step of the layout transition: moving subnodes.

Huy simplified this diff to just a few lines.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASDisplayNode+Layout.mm
```diff
@@ -993,12 +993,6 @@ - (void)_pendingLayoutTransitionDidComplete
   _pendingLayoutTransition = nil;
 }
 
-- (void)_setCalculatedDisplayNodeLayout:(const ASDisplayNodeLayout &)displayNodeLayout
```
nguyenhuy: For the record, this method is not valled anywhere.

Source/ASDisplayNode.mm
```diff
@@ -1220,11 +1220,10 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
   }
   ASDisplayNodeLogEvent(self, @"computedLayout: %@", layout);
 
-  // Return the (original) unflattened layout if it needs to be stored. The layout will be flattened later on (@see _locked_setCalculatedDisplayNodeLayout:).
-  // Otherwise, flatten it right away.
-  if (! [ASDisplayNode shouldStoreUnflattenedLayouts]) {
-    layout = [layout filteredNodeLayoutTree];
+  if ([ASDisplayNode shouldStoreUnflattenedLayouts]) {
+      _unflattenedLayout = layout;
```
nguyenhuy: As discussed offline, there is a small risk that the layout here will not be used/applied later on, and Weaver (or any other users of this debugging feature) will end up with a wrong unflattened layout. But given it's a debugging feature and should not be used in producation, the risk is small and doesn't warrant the complexity in previous implementation.

maicki: Can you add some comment about this fact?

mikezucc:
I will rebase with the correct author, no worry

PR #1158, opened 2018-10-03
Set default tuning params

wsdwsd0829:
It's valid that user may set tuning params to zero before view is loaded and it's better to provide default values.


Source/Details/ASAbstractLayoutController.h
```diff
@@ -20,6 +22,8 @@ AS_EXTERN CGRect CGRectExpandToRangeWithScrollableDirections(CGRect rect, ASRang
 
 @interface ASAbstractLayoutController : NSObject <ASLayoutController>
 
++ (std::vector<std::vector<ASRangeTuningParameters>>)defaultTuningParameters;
```
nguyenhuy: Since this is a public header, I'm a bit worried about exposing Obj-C++ here since it's going to cause a cascade effect on every header file that imports it and every impl file that imports any of these headers. We're already seeing the effect in this PR.

How about we expose a Obj-C method here that converts the Obj-C data structure into a vector? Then maybe expose this Obj-C++ version in a **framework private** header, in case we want to mitigate any perf concern regarding the conversion.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e18b76375575b38fef549833c189e66a/raw/2618b56a9df59d80ee5222eccf5bcf9cf34644b3/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f4b1cf222e08f561ace42cd9b3c9e1f6/raw/6c339b92bf5dde8ea48b51f55d21a38002032aca/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d20d001d3434c2ffd33005e774b4d091/raw/2ae6c4f2af5f2e3ef1bdba3cb9a4ae82fbc27fec/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/fbe03f900dc86061fb2642b58436f11c/raw/d772e5971b05422ba09397e2f1794ddb9db3f245/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/330a9f5a99feb14d538494ed258b4dfb/raw/5741f87353e25dd52e2a1ef60a849b452300c833/log.txt)

Issue #1135, opened 2018-09-19
ASEditableTextNode contents not accessible to VoiceOver


ay8s:
Just had a report of this, it looks like the contents typed into an ASEditableTextNode aren't accessible when using VoiceOver. Will try and dig in tomorrow and see if I can do a pull request.

maicki:
@ay8s Hey, did you find something out yet? Thanks!

ay8s:
For now I ended up updating the accessibility value with the text from the textView in editableTextNodeDidUpdateText: 

Also digging into a few other reports. Curious if Pinterest has any issues at all with ASEditableTextNode accessibility? Seems odd it's not free as a wrapper around UITextView.

ay8s:
Digging into this one some more, there are pretty drastic differences in what VoiceOver reads between ASEditableTextNode and UITextView. In this example I simply put a ASEditableTextNode (green) at the top and a UITextView (red) below. Both are within a ASViewController with the UITextView wrapped in a ASDisplayNode.

No other settings were changed for either the ASEditableTextNode or the UITextView.

ASEditableTextNode
- Doesn't say that it is a text field.
- Doesn't say "Double tap to edit"
- Doesn't say how to access misspelled words.

When inputting characters...
- Doesn't read completed words. 
- Doesn't read insertion points. 

Will continue to dig in, now sure if there may be something conflicting with the expected VO output.

Video: https://cl.ly/49c21c61cd58

ay8s:
Digging in more to this today after exploring just using UITextView. Seems that if I use ASDK_ACCESSIBILITY_DISABLE it works correctly. Oddly ASViewController's that use ASEditableTextNode as the root node work fine, only when a ASEditableTextNode is included as a subnode default behavior is lost.

ay8s:
@maicki @nguyenhuy So I think I've figured out the cause. When a ASEditableTextNode is a root node its textView is available via "view". This isn't the case when it is a subnode. When we're collecting the accessibility elements here https://github.com/TextureGroup/Texture/blob/465abb1ded1777e7573a65e70df0ef0a2bda0df4/Source/Details/_ASDisplayViewAccessiblity.mm#L215-L234 we're not looking for the textView. 

I quickly hacked together adding the textView to the elements array and VoiceOver returned the correct information back to the user. 

Any downsides you can see adding the following?
```
  for (ASDisplayNode *subnode in node.subnodes) {
    if (subnode.isAccessibilityElement) {
      
[...]

PR #1162, opened 2018-10-04
Make ASEditableTextNode accessible to VoiceOver

ay8s:
Should resolve issues seen in #1135 where VoiceOver is unable to access the text view and correctly read out the value while inputting text.


maicki:
Could we try implement the accessibility methods within ASEditableTextNode:

```
// Return 1 here for the textView
- (NSInteger)accessibilityElementCount
{
  return 1;
}

- (id)accessibilityElementAtIndex:(NSInteger)index
{
  return self.textView;
}

- (NSInteger)indexOfAccessibilityElement:(id)element
{
  return 1;
}
```

Not sure if this is the right approach yet, but we would make the `ASEditableTextNode` an accessibility container with this.

ay8s:
@maicki Tried a couple of approaches with this. 

Making the ASEditableTextNode a container with isAccessibilityElement true, doesn't get it picked up as we don't collect the UIAccessibilityElements for non layerBacked nodes and only add the subnodes view to the array.

With isAccessibilityElement set to false we get the same outcome where only the subnodes view is added to the array.

https://github.com/TextureGroup/Texture/blob/master/Source/Details/_ASDisplayViewAccessiblity.mm#L215-L234

maicki:
If this lands: #1188 we may can just return the `textView` within `-[ASEditableTextNode accessibilityElements]` `ASEditableTextNode`. cc @ay8s

maicki:
@ay8s #1188 just landed on master. Could you pull latest and check if my suggestion above would work? Thanks!

ay8s:
On it @maicki

Source/ASEditableTextNode.mm
```diff
@@ -842,4 +842,30 @@ - (void)_delegateDidFinishEditing
     [_delegate editableTextNodeDidFinishEditing:self];
 }
 
+
+- (BOOL)isAccessibilityElement
```
maicki: We are treating `ASEditableTextNode` as a `UIAccessibilityContainer` now, so this method should return `NO`. You should also be able to just remove it as by default it will return `NO`. Some more details on `UIAccessibilityContainer`: https://developer.apple.com/documentation/uikit/accessibility/uiaccessibilitycontainer

ay8s: Good catch! Was a carry over from when I had it working previously without #1188. Removed now.

Source/ASEditableTextNode.mm
```diff
@@ -842,4 +842,24 @@ - (void)_delegateDidFinishEditing
     [_delegate editableTextNodeDidFinishEditing:self];
 }
```
maicki: Can you please add a pragma to this line: `#pragma mark - UIAccessibilityContainer`

Source/ASEditableTextNode.mm
```diff
@@ -842,4 +842,24 @@ - (void)_delegateDidFinishEditing
...
 
+- (NSInteger)accessibilityElementCount
+{
+  return 1;
+}
+
+- (NSArray *)accessibilityElements
```
maicki: We have to handle if the `self.textView` is nil. This can happen if the node was not loaded yet. We should decide if we would like to load the node at that time what should load the textView, but currently this could crash.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR modifies CHANGELOG.md, which is a generated file. Add #changelog to the title to suppress this warning.</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASEditableTextNode.mm
maicki: @nguyenhuy @Adlai-Holler  Any thoughts on if we should force loading the view in this methods or just don't return any accessibility elements like above?

CHANGELOG.md
```diff
@@ -622,4 +622,4 @@
 
 
 
-\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
```
maicki: @ay8s Can you remove all of the changelog changes.

Source/ASEditableTextNode.mm
maicki: Let's go a different route, Imho the accessibility elements should not be requested if the node is not loaded yet:

```
#pragma mark - UIAccessibilityContainer

- (NSInteger)accessibilityElementCount
{
  if (!self.isNodeLoaded) {
    ASDisplayNodeFailAssert(@"Cannot access accessibilityElementCount since ASEditableTextNode is not loaded");
    return 0;
  }
  return 1;
}

- (NSArray *)accessibilityElements
{
  if (!self.isNodeLoaded) {
    ASDisplayNodeFailAssert(@"Cannot access accessibilityElements since ASEditableTextNode is not loaded");
    return @[];
  }
  return @[self.textView];
}

- (id)accessibilityElementAtIndex:(NSInteger)index
{
  if (!self.isNodeLoaded) {
    ASDisplayNodeFailAssert(@"Cannot access accessibilityElementAtIndex: since ASEditableTextNode is not loaded");
    return nil;
  }
  return self.textView;
}

- (NSInteger)indexOfAccessibilityElement:(id)element
{
  return 0;
}
```

After that change let's get it in.

PR #1166, opened 2018-10-10
ASTextNode2 to ignore certain text alignments while calculating intrinsic size

nguyenhuy:
ASTextNode2 uses ASTextLayout to calculate its layout and bounding rect. When the constrained width that is used for layout calculation is inf/max (e.g when the node is inside a horizontal stack), ASTextLayout doesn't ignore its right/center/natural text alignment but takes it into account. That results in an unreasonable size (and position).

Fix by detecting when the node is calculating intrinsic size and force its layout alignment to be left. Other alignments should still work when the max width is finite/reasonable.

### DIFF
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -235,7 +235,7 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   _textContainer.size = constrainedSize;
   [self _ensureTruncationText];
 
-  // If the constrained size has max/inf value on the text's forward direction, the text node is calculating its intrinsic size.
+  // If the constrained size has a max/inf value on the text's forward direction, the text node is calculating its intrinsic size.
   BOOL isCalculatingIntrinsicSize;
   if (_textContainer.isVerticalForm) {
     isCalculatingIntrinsicSize = (_textContainer.size.height >= ASTextContainerMaxSize.height);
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/ASTextNode2.mm
```diff
@@ -234,9 +234,12 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 
   _textContainer.size = constrainedSize;
   [self _ensureTruncationText];
-  
+
+  // If constrained width is max/inf, the text node is calculating its intrinsic size.
+  const BOOL calculatingIntrinsicSize = (_textContainer.size.width >= ASTextContainerMaxSize.width);
```
wiseoldduck: Does Texture ever use the `isVerticalForm` property? We should probably just delete it, but if not maybe check for it here.

Source/ASTextNode2.mm
```diff
@@ -321,18 +324,31 @@ - (NSArray *)exclusionPaths
   return _textContainer.exclusionPaths;
 }
 
-- (void)prepareAttributedString:(NSMutableAttributedString *)attributedString
+- (void)prepareAttributedString:(NSMutableAttributedString *)attributedString forIntrinsicSize:(BOOL)isForIntrinsicSize
```
wiseoldduck: This method name implies the second parameter is a `CGSize`, to me something like `isForIntrinsicSize` would imply a `BOOL`.

nguyenhuy:
Awesome! thank you for reviewing, @Adlai-Holler and @wiseoldduck.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/37f500c4b791322817aebb1371edf0d6/raw/100ab4bb62037159dd3ecc9749739372163b3b1b/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f902acd27633bc0ef207150a9a4149af/raw/86a5be3ba5c80a48ba7ecb9fbafd39f6a71335fc/log.txt)

PR #1167, opened 2018-10-10
Small fix in ASTextKitRenderer #trivial

nguyenhuy:
Avoid building multiple rects that are exactly the same.


wiseoldduck:
LGTM, but I haven't actually worked with the TextKit version of ASTextNode much!

#AintNoTextNodePartyLikeAnASTextNode2Party

PR #1168, opened 2018-10-10
Improve ASTextNode2 locking in setAttributedText: #trivial

maicki:
We are holding the lock in ASTextNode2 while calling `setNeedsDisplay`. We should adapt a similar path in `ASTextNode2` as in `ASTextNode`.

### DIFF
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -271,22 +271,22 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
     attributedText = [[NSAttributedString alloc] initWithString:@"" attributes:nil];
   }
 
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5659265a4543e23ec7e2faf4a6704f1f/raw/947c9bc5808ee3cd8da05d35bd049c12ce46a0ee/log.txt)

Adlai-Holler:
Is the current behavior causing us problems? We should probably allow calling setNeedsDisplay with the lock held – invalidating the contents is part of the transaction. We should never expose a state where we have a new string, but are contents are up-to-date … we are hiding the fact that we are planning to invalidate the display soon.

maicki:
Let's abandoning the PR than.

PR #1175, opened 2018-10-17
Expose test_resetWithConfiguration: for testing #trivial

maicki:
This makes it easier for conforming UI updates within tests.


maicki:
@Adlai-Holler Ok let's go a different route. We have to expose `+test_resetWithConfiguration` otherwise you have to expose it somehow within the tests.

PR #1177, opened 2018-10-17
Expose a way to determine if a text node will truncate for a given constrained size #trivial

maicki:
- Expose textLayoutForConstraint:, unavailable for `ASTextNode` for now
- Refactor `compatibleLayoutWithContainer:text:` into a static method


maicki:
@Adlai-Holler It's not due to performance, it's due to the reference of `ASTextNode2`. There will be a change coming up soon and that code change will help.

Removed commented out code.

PR #1180, opened 2018-10-17
Move import of stdatomic to ASRecursiveUnfairLock implementation file #trivial

maicki:
This should fix a problem if a `ASRecursiveUnfairLock` is imported in a non c++ compiled file as the implementation file will not have access to `stdatomic.h`.


maicki:
@Adlai-Holler Since C11 `_Atomic` is a keyword and therefore we don't have to include the header in this case: https://en.cppreference.com/w/c/keyword

PR #1184, opened 2018-10-22
 Don’t handle touches on additional attributed message if passthrough is enabled

maicki:
This diff will change the behavior handling tapping on the additional attributed message:
If `passthroughNonlinkTouches` is set to YES, taps on the additional attributed message will *not* call the `textNodeTappedTruncationToken:` delegate method anymore. Previously tapping on the additional attributed message always called `textNodeTappedTruncationToken:` without any dependence on `passthroughNonlinkTouches`.

Another approach could be to add a new property like: `handleAdditionalTruncationString` that if set to YES will call `textNodeTappedTruncationToken:` otherwise not. This would not depend on `passthroughNonlinkTouches`.

Furthermore this diff fixes taps within the additional attributed message within `ASTextNode2` what currently didn't really work. Before this the tap was not detected if tapped directly on the additional attributed text, it was only detected if tapped a bit outside of the additional attributed text but still within the `ASTextNode2` bounds.

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -974,7 +974,7 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
   BOOL linkCrossesVisibleRange = (lastCharIndex > range.location) && (lastCharIndex < NSMaxRange(range) - 1);
 
   if (inAdditionalTruncationMessage) {
-    return YES;
+    return !_passthroughNonlinkTouches;
   } else if (range.length && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
     return YES;
   } else {
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -973,9 +973,7 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
   NSUInteger lastCharIndex = NSIntegerMax;
   BOOL linkCrossesVisibleRange = (lastCharIndex > range.location) && (lastCharIndex < NSMaxRange(range) - 1);
 
-  if (inAdditionalTruncationMessage) {
-    return !_passthroughNonlinkTouches;
-  } else if (range.length && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
+  if (range.length && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
     return YES;
   } else {
     return NO;
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -973,7 +973,7 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
   NSUInteger lastCharIndex = NSIntegerMax;
   BOOL linkCrossesVisibleRange = (lastCharIndex > range.location) && (lastCharIndex < NSMaxRange(range) - 1);
 
-  if (range.length && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
+  if (range.length > 0 && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
     return YES;
   } else {
     return NO;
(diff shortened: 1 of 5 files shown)


Source/ASTextNode.mm
```diff
@@ -974,7 +974,7 @@ - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event
   BOOL linkCrossesVisibleRange = (lastCharIndex > range.location) && (lastCharIndex < NSMaxRange(range) - 1);
 
   if (inAdditionalTruncationMessage) {
-    return YES;
+    return !_passthroughNonlinkTouches;
```
Adlai-Holler: Right above here we short-circuit `if (!_passthroughNonlinkTouches)`. So this is equivalent to `return NO.` We should ideally remove this branch entirely, and simply return YES if we're in a link and NO otherwise. Whether we're in the truncation message or not doesn't matter for this method.

maicki: @Adlai-Holler Agree and addressed

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e045fda787da0fbe5262df63b337c384/raw/dea8ad58c3442e662a94f56321634a60e0263e3d/log.txt)

Source/ASTextNode2.mm
```diff
@@ -617,6 +615,63 @@ - (id)_linkAttributeValueAtPoint:(CGPoint)point
...
+  BOOL inAdditionalTruncationMessage = NO;
+  
+  CTLineRef truncatedCTLine = layout.truncatedLine.CTLine;
+  if (truncatedCTLine != NULL && _additionalTruncationMessage != nil) {
+    CFIndex stringIndexForPosition = CTLineGetStringIndexForPosition(truncatedCTLine, point);
+    if (stringIndexForPosition != kCFNotFound) {
+      CFIndex truncatedCTLineGlyphCount = CTLineGetGlyphCount(truncatedCTLine);
+      
+      CTLineRef truncationTokenLine =
+      CTLineCreateWithAttributedString((CFAttributedStringRef)_truncationAttributedText);
```
appleguy: Bump this to the line above, and on the several lines below. No 100 char line limits in Texture, thank goodness!

Source/ASTextNode2.mm
```diff
@@ -617,6 +615,63 @@ - (id)_linkAttributeValueAtPoint:(CGPoint)point
...
+      
+      CTLineRef additionalTruncationTokenLine = 
+      CTLineCreateWithAttributedString((CFAttributedStringRef)_additionalTruncationMessage);
+      CFIndex additionalTruncationTokenLineGlyphCount =
+      additionalTruncationTokenLine ? CTLineGetGlyphCount(additionalTruncationTokenLine) : 0;   
+      
+      switch (_textContainer.truncationType) {
+        case ASTextTruncationTypeStart: {
+          if (stringIndexForPosition > truncationTokenLineGlyphCount &&
+              stringIndexForPosition < (additionalTruncationTokenLineGlyphCount + truncationTokenLineGlyphCount)) {
```
appleguy: This expression is used here and on line 648, though with the variables added together in the opposite order. Is there a descriptive name we can give for this value and put it in a variable for better readability?

examples/Kittens/Sample/AppDelegate.m
```diff
@@ -24,3 +24,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 }
 
 @end
+
+@implementation ASConfiguration (UserProvided)
+
++ (ASConfiguration *)textureConfiguration
+{
+  ASConfiguration *configuration = [ASConfiguration new];
+//  configuration.experimentalFeatures = ASExperimentalTextNode;
```
appleguy: Possibly worth landing it with this enabled :). Otherwise probably delete the commented line.

Source/ASTextNode2.mm
```diff
@@ -617,6 +615,63 @@ - (id)_linkAttributeValueAtPoint:(CGPoint)point
   return nil;
 }
 
+- (BOOL)_locked_pointInsideAdditionalTruncationMessage:(CGPoint)point withLayout:(ASTextLayout *)layout
+{
+  // Check if the range is within the additional truncation range
+  BOOL inAdditionalTruncationMessage = NO;
+  
+  CTLineRef truncatedCTLine = layout.truncatedLine.CTLine;
+  if (truncatedCTLine != NULL && _additionalTruncationMessage != nil) {
```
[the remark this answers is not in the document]
maicki: Across the text nodes where are checking `_additionalTruncationMessage != nil` if we determine if the an `_additionalTruncationMessage` is set. I would suggest we keep it as `_additionalTruncationMessage != nil`

PR #1187, opened 2018-10-23
Yoga integration improvements

maicki:
- Add thread safety for Yoga layout tree manipulation
- Support baseline alignments for ASYogaLayout
- Refactor ASLayoutElementYogaBaselineFunc to not require yogaParent (its parent style is set into a private var on ASLayoutElementStyle before layout instead)
- Only set the accessibility element to nil if the view is loaded
- Add nodeWillCalculateLayout callback to ASNodeController

### DIFF
--- a/Source/ASDisplayNode+Beta.h
+++ b/Source/ASDisplayNode+Beta.h
@@ -199,6 +199,7 @@ AS_EXTERN void ASDisplayNodePerformBlockOnEveryYogaChild(ASDisplayNode * _Nullab
 
 @property BOOL yogaLayoutInProgress;
 @property (nullable, nonatomic) ASLayout *yogaCalculatedLayout;
+@property (weak, readonly) ASDisplayNode *yogaRoot;
 
 // These methods are intended to be used internally to Texture, and should not be called directly.
 - (BOOL)shouldHaveYogaMeasureFunc;
--- a/Source/ASDisplayNode+Yoga.mm
+++ b/Source/ASDisplayNode+Yoga.mm
@@ -275,10 +275,16 @@
     return;
   }
 
+  [self enumerateInterfaceStateDelegates:^(id<ASInterfaceStateDelegate>  _Nonnull delegate) {
+    if ([delegate respondsToSelector:@selector(nodeWillCalculateLayout:)]) {
+      [delegate nodeWillCalculateLayout:rootConstrainedSize];
+    }
+  }];
+
   ASLockScopeSelf();
 
   // Prepare all children for the layout pass with the current Yoga tree configuration.
-  ASDisplayNodePerformBlockOnEveryYogaChild(self, ^(ASDisplayNode * _Nonnull node) {
+  ASDisplayNodePerformBlockOnEveryYogaChild(self, ^(ASDisplayNode *_Nonnull node) {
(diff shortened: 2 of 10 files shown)


Source/ASDisplayNode+Yoga.mm
```diff
@@ -31,8 +32,17 @@ - (ASSizeRange)_locked_constrainedSizeForLayoutPass;
 
 @implementation ASDisplayNode (Yoga)
 
+- (ASDisplayNode *)yogaRoot {
+  ASDisplayNode *yogaRoot = self;
+  while(yogaRoot.yogaParent) {
+    yogaRoot = yogaRoot.yogaParent;
```
Adlai-Holler: Can we avoid calling `yogaParent` twice per node?

Source/ASDisplayNode+Yoga.mm
```diff
@@ -261,11 +275,22 @@ - (void)calculateLayoutFromYogaRoot:(ASSizeRange)rootConstrainedSize
...
+  }];
+
   ASLockScopeSelf();
 
   // Prepare all children for the layout pass with the current Yoga tree configuration.
-  ASDisplayNodePerformBlockOnEveryYogaChild(self, ^(ASDisplayNode * _Nonnull node) {
+  ASDisplayNodePerformBlockOnEveryYogaChild(self, ^(ASDisplayNode *_Nonnull node) {
     node.yogaLayoutInProgress = YES;
+    if (node.yogaParent) {
+      node.style.parentAlignStyle = node.yogaParent.style.alignItems;
```
Adlai-Holler: Avoid getting parent twice?

Source/ASDisplayNode+Beta.h
```diff
@@ -199,6 +199,7 @@ AS_EXTERN void ASDisplayNodePerformBlockOnEveryYogaChild(ASDisplayNode * _Nullab
 
 @property BOOL yogaLayoutInProgress;
 @property (nullable, nonatomic) ASLayout *yogaCalculatedLayout;
+@property (weak, readonly) ASDisplayNode *yogaRoot;
```
Adlai-Holler: I might declare this a method, rather than an atomic weak property. Just to be more transparent.

Source/ASDisplayNode+Yoga.mm
appleguy: Yes, also a space after "while ("

Source/ASDisplayNode+Yoga.mm
```diff
@@ -31,8 +32,17 @@ - (ASSizeRange)_locked_constrainedSizeForLayoutPass;
...
+  ASDisplayNode *yogaRoot = self;
+  while(yogaRoot.yogaParent) {
+    yogaRoot = yogaRoot.yogaParent;
+  }
+  return yogaRoot;
+}
+
 - (void)setYogaChildren:(NSArray *)yogaChildren
 {
+  ASLockScope(self.yogaRoot);
```
appleguy: Is this safe? Can't the yogaRoot be changing at this time? I'm pretty sure we also need to lock self, before accessing this and also hold that lock for the duration of the method for it to be re-entrant safe.

PR #1196, opened 2018-10-26
ASTextNode2 to consider both width and height when determining if it is calculating an intrinsic size

ernestmama:
It was only considering the individual constraint width/height when calculating the intrinsic size. However, this will not work when textnode is in a vertical ASStackLayoutSpec which also try to set a flexShrink. 

If it needs to calculate either height or width based on its own content where either of them is ambiguous, it needs to apply the proper line break mode and text alignment.

Related commit: https://github.com/TextureGroup/Texture/pull/1166/commits/a139d66c5d1d90ace6ef9adcaabe86d6df62bf38

### DIFF
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -340,12 +340,9 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   [self _ensureTruncationText];
 
   // If the constrained size has a max/inf value on the text's forward direction, the text node is calculating its intrinsic size.
-  BOOL isCalculatingIntrinsicSize;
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

nguyenhuy:
Adding a change log entry isn't required because it'd be the same with the one added for https://github.com/TextureGroup/Texture/pull/1166.

PR #1197, opened 2018-10-27
Allow configuring imageCache when initializing ASPINRemoteImageDownloader.

wiseoldduck:


maicki:
@nguyenhuy @ernestmama @garrettmoon  Could someone please take a look over this PR. Thanks!

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -116,56 +115,67 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
-  NSAssert(!(configuration != nil && preconfiguredPINRemoteImageManager != nil), @"Either configuration or preconfigured image manager can be set at a time.");
-  static dispatch_once_t onceToken;
-  dispatch_once(&onceToken, ^{
-
-    if (preconfiguredPINRemoteImageManager) {
-      sharedPINRemoteImageManager = preconfiguredPINRemoteImageManager;
-    } else {
+  PINRemoteImageManager *manager = nil;
+  // Check that Carthage users have linked both PINRemoteImage & PINCache by testing for one file each
+  if (!(NSClassFromString(@"PINRemoteImageManager"))) {
```
maicki: We should put that behind a debug flag.

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -116,56 +115,67 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
-    }
-  });
-  return sharedPINRemoteImageManager;
+  return manager;
 }
 
 - (PINRemoteImageManager *)sharedPINRemoteImageManager
 {
-  return [ASPINRemoteImageDownloader sharedPINRemoteImageManagerWithConfiguration:nil preconfiguredPINRemoteImageManager:nil];
+  if (!sharedPINRemoteImageManager) {
```
maicki: Why do we need to have this check?

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -116,56 +115,67 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
-        @throw e;
-      }
-      sharedPINRemoteImageManager = [[ASPINRemoteImageManager alloc] initWithSessionConfiguration:configuration
-                                                                alternativeRepresentationProvider:[self sharedDownloader]];
+    sharedPINRemoteImageManager = [[ASPINRemoteImageManager alloc] initWithSessionConfiguration:configuration
+                                                                   alternativeRepresentationProvider:[self sharedDownloader]
+                                                                   imageCache:imageCache];
 #else
-      sharedPINRemoteImageManager = [[ASPINRemoteImageManager alloc] initWithSessionConfiguration:configuration];
+    manager = [[ASPINRemoteImageManager alloc] initWithSessionConfiguration:configuration
```
maicki: Why did this change to manager instead of `sharedPINRemoteImageManager`?

Source/Details/ASPINRemoteImageDownloader.h
```diff
@@ -47,7 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @param preconfiguredPINRemoteImageManager The preconfigured remote image manager that will be used by `sharedDownloader`
  */
-+ (void)setSharedPreconfiguredRemoteImageManager:(nullable PINRemoteImageManager *)preconfiguredPINRemoteImageManager;
++ (void)setSharedPreconfiguredRemoteImageManager:(PINRemoteImageManager *)preconfiguredPINRemoteImageManager;
```
maicki: If we remove the nullable here it does not match with the implementation. Any reason we removed it?

wiseoldduck: I should change the implementation if I didn't. It makes no sense to call this with `nil`, to support it would be to add a check right away along the lines of `if (preconfiguredPINRemoteImageManager == nil) return;` which is silly.

Source/Details/ASPINRemoteImageDownloader.m
wiseoldduck: OK

Source/Details/ASPINRemoteImageDownloader.m
wiseoldduck: because it's not shared - I funneled all the set-shared logic into as few places (two, unfortunately couldn't get it to one without changing the API of the getter) as possible to reduce complexity. This method just configures a manager and returns it to you.

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -116,56 +115,67 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
-  return sharedPINRemoteImageManager;
+  return manager;
 }
 
 - (PINRemoteImageManager *)sharedPINRemoteImageManager
 {
-  return [ASPINRemoteImageDownloader sharedPINRemoteImageManagerWithConfiguration:nil preconfiguredPINRemoteImageManager:nil];
+  if (!sharedPINRemoteImageManager) {
+    dispatch_once(&shared_init_predicate, ^{
+      sharedPINRemoteImageManager = [ASPINRemoteImageDownloader PINRemoteImageManagerWithConfiguration:nil imageCache:nil];
```
nguyenhuy: If `shared_init_predicate` was used in `setSharedPreconfiguredRemoteImageManager`, then this block shouldn't be executed here, right? So we don't need the nil check for `sharedPINRemoteImageManager`? Not a big deal though.

Source/Details/ASPINRemoteImageDownloader.m
maicki: I wondered as I reviewed before why this is the case? cc @garrettmoon

Source/Details/ASPINRemoteImageDownloader.m
maicki: Cool than we should change the other `sharedPINRemoteImageManager` to the `manager` variable too :)

Source/Details/ASPINRemoteImageDownloader.m
maicki: Yeah agree, no nil check necessary as the `dispatch_once` is guarding it anyway.

Source/Details/ASPINRemoteImageDownloader.m
wiseoldduck: I think I hit the error condition once, and didn't get the exception, and saw that, and thought it would be helpful, and it didn't have anything really to do with `PIN_ANIMATED_AVAILABLE` specifically, so I moved it out. *takes a breath*

I can see putting it behind `#if DEBUG`, although it's just going to crash either way and for a check that is guaranteed to only ever run once it doesn't seem like two `NSClassFromString`s is that expensive. But it's of limited use without a debugger attached (you would get a friendlier crash report from Apple if your released app hit here, but then presumably your app is crashing every time so ...... you hopefully have it figured out)

Source/Details/ASPINRemoteImageDownloader.m
wiseoldduck: Give me a minute to try to remember lol!

If I can't, I will write it off as a brain fart...

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -116,56 +115,67 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
+  PINRemoteImageManager *sharedManager = [self PINRemoteImageManagerWithConfiguration:configuration imageCache:imageCache];
+  [self setSharedPreconfiguredRemoteImageManager:sharedManager];
+}
+
+static dispatch_once_t shared_init_predicate;
+
++ (void)setSharedPreconfiguredRemoteImageManager:(PINRemoteImageManager *)preconfiguredPINRemoteImageManager
 {
   NSAssert(sharedDownloader == nil, @"Singleton has been created and session can no longer be configured.");
   NSAssert1(sharedPINRemoteImageManager == nil, @"An instance of %@ has been set. Either configuration or preconfigured image manager can be set at a time and only once.", [[sharedPINRemoteImageManager class] description]);
```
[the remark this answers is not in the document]
wiseoldduck: Sure why not :)

Source/Details/ASPINRemoteImageDownloader.m
wiseoldduck: OK, I took nine, and I got nothing! I am pretty sure I had worked myself into some overcomplicated concurrency scenario that cannot actually exist and wouldn't have been completely helped anyway

Source/Details/ASPINRemoteImageDownloader.h
wiseoldduck: Looks like I did actually also change the implementation so I'm happy with this as is.

Source/Details/ASPINRemoteImageDownloader.m
```diff
@@ -116,56 +115,68 @@ + (ASPINRemoteImageDownloader *)sharedDownloader NS_RETURNS_RETAINED
...
+  PINRemoteImageManager *sharedManager = [self PINRemoteImageManagerWithConfiguration:configuration imageCache:imageCache];
+  [self setSharedPreconfiguredRemoteImageManager:sharedManager];
 }
 
-+ (void)setSharedPreconfiguredRemoteImageManager:(nullable PINRemoteImageManager *)preconfiguredPINRemoteImageManager
+static dispatch_once_t shared_init_predicate;
+
++ (void)setSharedPreconfiguredRemoteImageManager:(PINRemoteImageManager *)preconfiguredPINRemoteImageManager
 {
+  NSAssert(preconfiguredPINRemoteImageManager != nil, @"setSharedPreconfiguredRemoteImageManager requires a non-nil parameter");
```
maicki: We may should use `NSParameterAssert` for the parameter check.

Source/Details/ASPINRemoteImageDownloader.h
ernestmama: Please also update the doc comments. The doc still say passing in a nil as the PINRemoteImageManager, it will create a default one.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/339b93467243320cdf126c9cec07c0be/raw/3fd4c2a0138041572e34a18896c95c7bbf56b5d2/log.txt)

PR #1209, opened 2018-11-02
Add -Wno-implicit-retain-self to podspec + smaller cleanups #trivial

maicki:
Cleanup warnings in ASDKgram and enable `CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF` to have the same settings as Xcode by default.


Adlai-Holler:
We all know it's generally bad to silence warnings, but in this case we will still get retain cycle warnings (`-Warc-retain-cycles`) and even as of Xcode 10.1, the compiler isn't smart enough to silence these warnings for noescape blocks.

I'd rather silence this semi-pedantic warning than add more friction and mess when using blocks.

PR #1216, opened 2018-11-06
ASCATransactionQueue interface trashing improvements

maicki:


nguyenhuy:
Diff looks good, but can you give a description on how this imrpoves interface trashing?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/c001d180402ec3cade15629c35d4045a/raw/d89f5fde9ecc813aaa9326bae5cc3ab1729b3586/log.txt)

Source/ASRunLoopQueue.mm
```diff
@@ -377,17 +372,13 @@ - (instancetype)init
     // __unsafe_unretained allows us to avoid flagging the memory cycle detector.
     __unsafe_unretained __typeof__(self) weakSelf = self;
     void (^handlerBlock) (CFRunLoopObserverRef observer, CFRunLoopActivity activity) = ^(CFRunLoopObserverRef observer, CFRunLoopActivity activity) {
+      while (_internalQueue.count > 0) {
```
wiseoldduck: Doesn't this defeat the point of the `weakSelf`?

maicki: Addressed

maicki:
Say before the change we are processing nodes 1, 2, 3 in queue, but the process of node 1 may do something that changes node 3's state. The while loop will handle that case, otherwise node 3's change by node 1 will be pushed to next loop.

PR #1217, opened 2018-11-06
Fix A11Y for horizontal collection nodes in Texture

maicki:
Collections have their own handling in UIKit that we shouldn't mess with. As such no nodes that contain CollectionNodes should ever be treated as AccessibilityContainers.

### DIFF
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -1971,7 +1971,8 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 
 #pragma mark - Accessibility overrides
 
-- (NSArray *)accessibilityElements {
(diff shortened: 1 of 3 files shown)


Source/ASTableView.mm
```diff
@@ -1969,4 +1969,11 @@ - (void)didMoveToSuperview
   }
 }
 
+#pragma mark - Accessibility overrides
+
+- (NSArray *)accessibilityElements {
+  [self waitUntilAllUpdatesAreCommitted];
```
nguyenhuy: I’m not familiar with accessbility and not sure when this method is called, but are you sure we want to block the (main?) thread in this case? Would it be better to somehow trigger a accessibility elements recollection once the table/collection finishes its updates.

PR #1219, opened 2018-11-07
Clean up our mutex, fix try_lock not hooking into assert mechanism

Adlai-Holler:
Might want to use split diff for this.

Just a major overhaul to clear some space. Note: AS_CHECK_LOCKING_SAFETY is gone – if we have asserts that are failing despite correct usage of our APIs, we need to comment them out and address them.

### DIFF
--- a/Source/Details/ASThread.h
+++ b/Source/Details/ASThread.h
@@ -114,13 +114,6 @@ namespace ASDN {
   class Mutex
   {
   public:
-    enum Type {
-      Plain,
-      Recursive,
-      Unfair,
-      RecursiveUnfair
-    };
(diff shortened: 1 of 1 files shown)


Source/Details/ASThread.h
```diff
@@ -9,11 +9,8 @@
 
 #import <Foundation/Foundation.h>
 
-#import <assert.h>
 #import <os/lock.h>
 #import <pthread.h>
```
Adlai-Holler: Note: pthread is still imported for ASDisplayNodeThreadIsMain function (no "main thread" in libcxx).

PR #1220, opened 2018-11-07
Push root node ref down the tree

Adlai-Holler:
Discuss.

Theory is, in the future all tree accesses will require the root lock. We could use a reader/writer mutex if we wanted. This diff uses the instance lock for convenience.

Right now, the down-propagation is murky because tree modifications *don't* require the root lock, so the ordering there is kind of unclear.

What's nice about this, is you could call a `locked_` method directly on the root node, safely.

If we do composite operations on the tree, such as ASM updating, should we hold the root the whole time or separate it, for example? Interesting stuff.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -2108,14 +2108,21 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
 // NOTE: This method must be dealloc-safe (should not retain self).
 - (ASDisplayNode *)supernode
 {
-#if CHECK_LOCKING_SAFETY
-  if (__instanceLock__.locked()) {
-    NSLog(@"WARNING: Accessing supernode while holding recursive instance lock of this node is worrisome. It's likely that you will soon try to acquire the supernode's lock, and this can easily cause deadlocks.");
-  }
-#endif
-  
   ASDN::MutexLocker l(__instanceLock__);
   return _supernode;
+}
+
+- (ASDN::UniqueLock)acquireRootLock __unused {
+  for (;; std::this_thread::yield()) {
+    if (!__instanceLock__.try_lock()) {
+      continue;
+    }
+    ASDN::UniqueLock rootLock(_rootNode->__instanceLock__, std::try_to_lock);
+    if (!rootLock.owns_lock()) {
+      continue;
+    }
+    return rootLock;
+  }
 }
 
 - (void)_setSupernode:(ASDisplayNode *)newSupernode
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
cc @wiseoldduck @wsdwsd0829

Adlai-Holler:
The `_rootNode` ivar is right now only guarded by `self->__instanceLock__` but in reality (once we add it) it is ALSO guarded by the root node's lock, since tree modifications are not allowed without the root locked.

Adlai-Holler:
@appleguy The public API wouldn't change.

@wiseoldduck Exactly

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/af6e9d7bc178c22525d576b3fced80aa/raw/d0136d8ebdf69527558f6dd6628d97fac0d51a1a/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/3e52fc8fa1927a94ffa4bbbd4dd2fb05/raw/61690bc9b633ce4133227ab3a3c304f11624214e/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/c115e7ab46de8c614e1522f9423d03f0/raw/8287ba224e8b30ff860d8f93cb81ba209fe1e0d1/log.txt)

Adlai-Holler:
The CI flaked again 😠

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/657f625892f8354a658501022eecfe3c/raw/46bfbe2cf75df5c647632f51757c130f2a207f12/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/75706ed520afc3a1c0f977414f4f7131/raw/434c1ebdcd1190589884850aee96523da4519125/log.txt)

Source/ASDisplayNode.mm
```diff
@@ -2118,6 +2119,26 @@ - (ASDisplayNode *)supernode
...
+    }
+    ASDN::UniqueLock rootLock(_rootNode->__instanceLock__, std::try_to_lock);
+    if (!rootLock.owns_lock()) {
+      continue;
+    }
+    return rootLock;
+  }
+}
+
 - (void)_setSupernode:(ASDisplayNode *)newSupernode
```
wsdwsd0829: YogaTree is different from normal node tree.  Meaning when we create a yoga tree, the rootnode will always be self.  So even during yoga layout in calculateLayouFromYogaRoot the _rootNode may not be updated timely.

PR #1221, opened 2018-11-07
Splitting up the Table view thrash tests for readability

mikezucc:
Looking into creating thrash tests for the collection view to reproduce some elusive collection view sync bugs. Referencing this test, thought I would re-enable it (since it does pass now) and separate out the sauce to be clearer


maicki:
@mikezucc Thanks for submitting. Can you not just rebase is on top of master? Thanks!

mikezucc:
yeah i did when i went from texture -> forked. lemme reset this

PR #1227, opened 2018-11-08
Have node and controller share lock

Adlai-Holler:
This prevents lock-interleaving deadlocks. The controller and node should stay in sync with each other conceptually anyway.

### DIFF
--- a/Source/ASNodeController+Beta.h
+++ b/Source/ASNodeController+Beta.h
@@ -10,18 +10,21 @@
 #import <AsyncDisplayKit/ASDisplayNode.h>
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h> // for ASInterfaceState protocol
 
+NS_ASSUME_NONNULL_BEGIN
+
 /* ASNodeController is currently beta and open to change in the future */
 @interface ASNodeController<__covariant DisplayNodeType : ASDisplayNode *>
-    : NSObject <ASInterfaceStateDelegate, NSLocking>
+    : NSObject <ASInterfaceStateDelegate>
 
-@property (nonatomic, strong /* may be weak! */) DisplayNodeType node;
+@property (strong, readonly /* may be weak! */) DisplayNodeType node;
 
 // Until an ASNodeController can be provided in place of an ASCellNode, some apps may prefer to have
 // nodes keep their controllers alive (and a weak reference from controller to node)
 
 @property (nonatomic) BOOL shouldInvertStrongReference;
 
-- (void)loadNode;
+// called on an arbitrary thread by the framework. You do not call this. Return a new node instance.
+- (DisplayNodeType)createNode;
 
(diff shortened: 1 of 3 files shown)


Source/ASNodeController+Beta.h
```diff
@@ -10,18 +10,21 @@
 #import <AsyncDisplayKit/ASDisplayNode.h>
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h> // for ASInterfaceState protocol
 
+NS_ASSUME_NONNULL_BEGIN
+
 /* ASNodeController is currently beta and open to change in the future */
 @interface ASNodeController<__covariant DisplayNodeType : ASDisplayNode *>
-    : NSObject <ASInterfaceStateDelegate, NSLocking>
+    : NSObject <ASInterfaceStateDelegate>
```
appleguy: Should we actually remove NSLocking here? Or can -lock/-unlock just do the right thing?

Adlai-Holler: It's definitely arguable. I like less code, and more transparency, but I don't feel strongly about it.

Source/ASNodeController+Beta.mm
```diff
@@ -12,33 +12,33 @@
...
 }
 
 - (ASDisplayNode *)node
 {
-  ASLockScopeSelf();
-  if (_node == nil) {
-    [self loadNode];
-  }
+  std::call_once(_nodeOnce, [self](){
+    [self setupReferencesWithNode:[self createNode]];
```
appleguy: Since we can't enforce that createNode is not called externally (or at least don't now), is it possible to only call it here if it hasn't been called yet / _node is nil?

Source/ASNodeController+Beta.mm
```diff
@@ -12,33 +12,33 @@
...
+  });
   return _node;
 }
 
 - (void)setupReferencesWithNode:(ASDisplayNode *)node
 {
-  ASLockScopeSelf();
+  ASDN::MutexLocker l(node->__instanceLock__);
```
appleguy: node can definitely be nil

appleguy:
Looks promising! Need to handle nil node, and we should probably prevent multiple createNode calls.

Also - in ELMComponent can actually replace its root node, if the materialize has a different type at the root; in that case does the node controller currently need to be able to reset .node or does it throw away the controller?

maicki:
@appleguy The node controller does not need to replace it's root. A new root controller is created every time. This was the case before this diff too, node always was treaded as `readonly` after `loadNode` was called. We should make that clear in our API though.

Adlai-Holler:
Actually no, there's a subtle race condition here. Changing…

Adlai-Holler:
There

Source/ASNodeController+Beta.h
Adlai-Holler: It's back in

PR #1229, opened 2018-11-10
Add extension points for global node behaviors, potentially replacing interface state delegates

Adlai-Holler:
right at the outset: 
- Causes no compiler warnings
- Does nothing unsafe or undocumented

Node controllers were added as a way for clients to customize node behavior globally. If you subclass ASTextNode and ASImageNode, you have to duplicate any common logic you want to add.

Node controllers however introduce additional complexity and duplication – if ASNetworkImageNode can fire off network requests, why does it need a controller? Nodes are in fact a sort of tiny view controller and we should stick with the original abstraction instead of adding a different one in.

This diff does something pretty strange, but I think useful. It modifies the node API to (1) offer a public ivar to ASDisplayNode and (2) guarantee that certain `ASDisplayNode+Subclass` category methods e.g. interface state methods are not implemented by the base class, and thus can be overridden safely in categories. To make it safe, we put in empty stubs for methods that they don't choose to implement.

I modified CatDealsCollectionView to demonstrate how this extension point could be used to add visibility logging events at the application level. 

I propose that this API *informally deprecates* the Beta `ASNodeController` and the `interfaceStateDelegates` APIs. Those systems will be removed.

### DIFF
--- a/Source/ASDisplayNode+Beta.h
+++ b/Source/ASDisplayNode+Beta.h
@@ -143,7 +143,14 @@ typedef struct {
  * this hook could be called up to 1 + (pJPEGcount * pJPEGrenderCount) times. The render count depends on how many times the downloader calls the
  * progressImage block.
  */
-- (void)hierarchyDisplayDidFinish;
+AS_CATEGORY_IMPLEMENTABLE
+- (void)hierarchyDisplayDidFinish NS_REQUIRES_SUPER;
+
+/**
+ * Only called on the root during yoga layout.
+ */
+AS_CATEGORY_IMPLEMENTABLE
+- (void)willCalculateLayout:(ASSizeRange)constrainedSize NS_REQUIRES_SUPER;
 
 /**
  * Only ASLayoutRangeModeVisibleOnly or ASLayoutRangeModeLowMemory are recommended.  Default is ASLayoutRangeModeVisibleOnly,
--- a/Source/ASDisplayNode+Subclasses.h
+++ b/Source/ASDisplayNode+Subclasses.h
@@ -69,8 +69,22 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @discussion This is the best time to add gesture recognizers to the view.
  */
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didLoad ASDISPLAYNODE_REQUIRES_SUPER;
 
+/**
+ * An empty method that you can implement in a category to add global
+ * node initialization behavior. This method will be called by [ASDisplayNode init].
+ */
+AS_CATEGORY_IMPLEMENTABLE
+- (void)baseDidInit;
+
+/**
+ * An empty method that you can implement in a category to add global
+ * node deallocation behavior. This method will be called by [ASDisplayNode dealloc].
+ */
+AS_CATEGORY_IMPLEMENTABLE
+- (void)baseWillDealloc;
 
 #pragma mark - Layout
 /** @name Layout */
@@ -88,6 +102,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @discussion Gives a chance for subclasses to perform actions after the subclass and superclass have finished laying
  * out.
  */
+AS_CATEGORY_IMPLEMENTABLE
 - (void)layoutDidFinish ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
@@ -96,6 +111,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @discussion When the .calculatedLayout property is set to a new ASLayout (directly from -calculateLayoutThatFits: or
  * calculated via use of -layoutSpecThatFits:), subclasses may inspect it here.
  */
+AS_CATEGORY_IMPLEMENTABLE
 - (void)calculatedLayoutDidChange ASDISPLAYNODE_REQUIRES_SUPER;
 
 
@@ -162,15 +178,25 @@ NS_ASSUME_NONNULL_BEGIN
   * For descriptions, see <ASInterfaceStateDelegate> definition.
   */
 
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didEnterVisibleState ASDISPLAYNODE_REQUIRES_SUPER;
+
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didExitVisibleState  ASDISPLAYNODE_REQUIRES_SUPER;
 
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didEnterDisplayState ASDISPLAYNODE_REQUIRES_SUPER;
+
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didExitDisplayState  ASDISPLAYNODE_REQUIRES_SUPER;
 
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didEnterPreloadState ASDISPLAYNODE_REQUIRES_SUPER;
+
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didExitPreloadState  ASDISPLAYNODE_REQUIRES_SUPER;
 
+AS_CATEGORY_IMPLEMENTABLE
 - (void)interfaceStateDidChange:(ASInterfaceState)newState
                       fromState:(ASInterfaceState)oldState ASDISPLAYNODE_REQUIRES_SUPER;
 
@@ -179,6 +205,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @discussion Subclasses can override this method to react to a trait collection change.
  */
+AS_CATEGORY_IMPLEMENTABLE
 - (void)asyncTraitCollectionDidChange;
 
 #pragma mark - Drawing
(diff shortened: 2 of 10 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please ensure license is correct for ASDisplayNode+CatDeals.mm:</p>
<pre><code>//
//  ASDisplayNode+CatDeals.mm
//  Texture
//
//  Copyright (c) Pinterest, Inc.  All rights reserved.
//  Licensed under Apache 2.0: http://www.apache.org/licenses/LICENSE-2.0
</code></</td>
    </tr>
  </tbody>
</table>

</p>

Adlai-Holler:
cc #1227

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ef967e6fb2b4f1d02a2a01623a32a227/raw/1fcd7d025f79dfe5b00841927bf237e4a131f602/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/9a879ac776fd791d0255c1c2c4963b6c/raw/9cc2e57588b98874acaee087aa4e3cae91bc76fd/log.txt)

appleguy:
@Adlai-Holler the rationale seems strong; Node Controllers were indeed added to allow global extension to ASDisplayNode subclasses by non-Texture / app-specific users. Indeed we have also found difficulties with Node Controllers as most infrastructure operates on e.g. ASCellNode specifically.

For some use cases it may still be required to have this state tracking done outside of the ASDisplayNode instance, if there is the possibility of that instance itself being swapped out. This introduces the potential complication of a layer we don't want to be in the layout tree, but should have jurisdiction over node children (e.g. like a component). If this approach allows for that, then we should be good!

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/60d45dd73c8353fa9e338f57b2adcc01/raw/d4df984ac688b288f3461e4cbd88e9cb8695c4e5/log.txt)

Adlai-Holler:
@appleguy Thanks for reviewing this! This approach is agnostic toward what you mentioned – if the client ever wanted to carry state information between nodes when the nodes are swapped out, it is straightforward and easy to do.

Adlai-Holler:
I'm curious to hear @maicki's take as well as @wiseoldduck and @wsdwsd0829 if they have time.

Source/ASDisplayNode.mm
```diff
@@ -247,6 +252,34 @@ + (void)initialize
...
+    class_addMethod(self, @selector(didExitPreloadState), noArgsImp, "v@:");
+    class_addMethod(self, @selector(didEnterDisplayState), noArgsImp, "v@:");
+    class_addMethod(self, @selector(didExitDisplayState), noArgsImp, "v@:");
+    class_addMethod(self, @selector(didEnterVisibleState), noArgsImp, "v@:");
+    class_addMethod(self, @selector(didExitVisibleState), noArgsImp, "v@:");
+    class_addMethod(self, @selector(hierarchyDisplayDidFinish), noArgsImp, "v@:");
+    class_addMethod(self, @selector(asyncTraitCollectionDidChange), noArgsImp, "v@:");
+    class_addMethod(self, @selector(calculatedLayoutDidChange), noArgsImp, "v@:");
+    
+    auto type0 = "v@:" + std::string(@encode(ASSizeRange));
```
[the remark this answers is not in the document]
Adlai-Holler: Yes, I don't like the boxing here either but since this code is only run once, and since `strncat` would still mean a heap allocation – although not two like we have today – the shorter code is worth it.

wiseoldduck:
I think if we can accomplish what we want via the delegates, that would be better from an OOD standpoint. The problems you describe & we have encountered are typical to a customize-through-subclasses approach and are why customization-via-composition (the pattern which the delegates approach attempts) is generally considered better than customization-via-inheritance. 

Categories help some but require the clients to understand the state of the node including locking, which seems less friendly than figuring out how to call the delegates from a consistent state where API re-entry is safe.

Source/ASDisplayNode+Subclasses.h
```diff
@@ -68,8 +68,22 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
+AS_CATEGORY_IMPLEMENTABLE
 - (void)didLoad ASDISPLAYNODE_REQUIRES_SUPER;
 
+/**
+ * An empty method that you can implement in a category to add global
+ * node initialization behavior. This method will be called by [ASDisplayNode init].
+ */
+AS_CATEGORY_IMPLEMENTABLE
+- (void)globalInit;
```
wiseoldduck: This seems like a strange name for an instance method

Adlai-Holler: Hmm good point. `configure` or `setup` or even `didInit` or `didBaseInit`?

wiseoldduck: Looks like the base class method is `_initializeInstance` which seems accurate and clear to me

Adlai-Holler:
OK I changed the names to `baseDidInit` and `baseWillDealloc` to be I think very clear. @maicki @wiseoldduck please review.

Source/ASDisplayNode+Subclasses.h
Adlai-Holler: Renamed to `baseDidInit` and `baseWillDealloc`

wiseoldduck:
Well I am a lot more friendly to this now that I have seen what a $@#%*&(^ it will be to make the delegate methods safe for re-entry (always unlocked)

Adlai-Holler:
We still should not call out with the lock held unless we document that for specific cases 😭

Source/ASDisplayNode.h
```diff
@@ -94,7 +94,14 @@ AS_EXTERN NSInteger const ASDefaultDrawingPriority;
  *
  */
 
-@interface ASDisplayNode : NSObject <ASLocking>
+@interface ASDisplayNode : NSObject <ASLocking> {
+@public
+  /**
+   * A field for clients to use as they please, for example in implementing category methods on
```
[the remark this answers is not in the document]
Adlai-Holler: Good call, done.

Source/ASDisplayNode+Beta.h
```diff
@@ -143,7 +143,14 @@ typedef struct {
  * this hook could be called up to 1 + (pJPEGcount * pJPEGrenderCount) times. The render count depends on how many times the downloader calls the
  * progressImage block.
  */
-- (void)hierarchyDisplayDidFinish;
+AS_CATEGORY_IMPLEMENTABLE
+- (void)hierarchyDisplayDidFinish NS_REQUIRES_SUPER;
+
+/**
+ * Only called on the root during yoga layout.
```
garrettmoon: Is this literally specific to Yoga?

garrettmoon:
I want to echo Huy that removing the interface state tracking in a follow up PR would cause us a bit of a burden and I think it's useful.

garrettmoon:
Additionally something to think through:

If someone decides to make a framework (let's say SuperAwesomeGridNode) that implements one of these categories, they're going to have to know that they need to explicitly warn their users not to also implement the category. And potentially create a different method that their category calls that's similar to this.

So if SuperAwesomeGridNode (ASDisplayNode) implements `didEnterPreloadState` they would have to document that they've done so, and implement another method `SAGNDidEnterPreloadState that they would have to stub out and call so clients could still hook in.

Long story short if we're going to hack this it should be documented with an `@warning` at the very least…

Adlai-Holler:
Thats true. I think its unlikely that any framework devs would be crazy enough to do that but it IS a vulnerability of this approach.

Source/ASDisplayNode.mm
```diff
@@ -433,6 +467,7 @@ - (void)onDidLoad:(ASDisplayNodeDidLoadBlock)body
 - (void)dealloc
 {
   _flags.isDeallocating = YES;
+  [self baseWillDealloc];
```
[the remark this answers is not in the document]
Adlai-Holler: Short answer: so that any calls that come out of `baseWillDealloc` will know not to retain self.

Medium length answer: The method is _base_ will dealloc, indicating the base class is going to run its dealloc routine. The flag refers to the _object_ deallocation which in fact began before this method was called, so we try to keep up with it.

Long answer: Technically flags like these are unsafe, even though they're common. The only correct way to find out if an object is deallocating (unsafe to retain) is to try and retain a weak pointer to the object. As of right now, if a subclass called something from its `dealloc`, our flag wouldn't be set and we might try to retain self. The safe thing to do is to register a weak pointer to `self` during `init` and replace this flag, with an attempt to retain that.

Source/ASDisplayNode+Beta.h
Adlai-Holler: Indeed it is.

Adlai-Holler:
Thanks for the reviews guys. We'll keep interface state delegates around and land this to start removing reliance on node controllers. I'm pretty stoked about this because it's been a long-standing limitation of the framework!

PR #886, opened 2018-04-18
Assert node did load before did enter visible way 1

wsdwsd0829:
Let's discuss which way is better #887

PR #1234, opened 2018-11-14
[ASDisplayNode] Propagate traits before loading a subnode

rcancro:
https://github.com/TextureGroup/Texture/pull/886 enforces that some nodes are loaded before they are sent a visiblity event, but it was not propagating the trait collection before forcing the load.

We are propagating traits in `_setSupernode:` (which was happening after the forced load). I moved the possible forced load to `_setSupernode:`, after propagating traits but before any visibility states are changed.


wsdwsd0829:
LGTM Thanks.

PR #1242, opened 2018-11-18
Add way to compile out ASTextNode + TextKit dependencies

maicki:
Currently we are having ASTextNode and ASTextNode2 which are adding quit a bit of size to Texture. As it becomes more and more clear that ASTextNode2 gonna be the future of text node within Texture we will provide a way to opt out of ASTextNode and make ASTextNode2 the default by default.

To be able to completly opt out of ASTextNode and all of it's TextKit dependency we are introducing a compiler flag called `AS_ENABLE_TEXTNODE`. By default the value is `YES / 1` to reflect the old behavior.

Furthermore there is a CocoaPods subspec added to disable the flag.

If the flag is disabled and ASTextNode2 is used only this will reduce around 300kb of Texture size.

### DIFF
--- a/Source/ASExperimentalFeatures.h
+++ b/Source/ASExperimentalFeatures.h
@@ -7,6 +7,7 @@
 //
 
 #import <Foundation/Foundation.h>
+#import <AsyncDisplayKit/ASAvailability.h>
 #import <AsyncDisplayKit/ASBaseDefines.h>
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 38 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/a72cc510de183e7e50f847d703459eb7/raw/5b067c05b31f7206f759e9d39814b36382361834/log.txt)

nguyenhuy:
Should we also compile out the ASTextNode2 experiment (`ASExperimentalTextNode` and friends) if only ASTextNode2 is available?

maicki:
I just updated the diff to compile out `ASExperimentalTextNode` and fix tests. Furthermore for the time being we will define `ASTextNode` to `ASTextNode2` if `AS_ENABLE_TEXTNODE` is `NO`.

PR #1245, opened 2018-11-20
Add development docs structure

garrettmoon:
This adds a structure for us add developer documentation too.


mikezucc:
Should there be some sort of link between Issues types and a section in the documentation?

PR #1247, opened 2018-11-21
add more delegate methods for monitoring network image node progress

ernestmama:
add more delegate methods for monitoring network image node progress so we can better keep track of optimization opportunity

### DIFF
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -365,14 +365,19 @@ static std::atomic_bool _useMainThreadDelegateCallbacks(true);
         [self _locked__setImage:result];
         _imageLoaded = YES;
         
+        id<ASNetworkImageNodeDelegate> delegate;
+        {
+          delegate = _delegate;
+        }
+        
         // Call out to the delegate.
         if (_delegateFlags.delegateDidLoadImageWithInfo) {
           ASUnlockScope(self);
           let info = [[ASNetworkImageLoadInfo alloc] initWithURL:url sourceType:ASNetworkImageSourceSynchronousCache downloadIdentifier:nil userInfo:nil];
-          [_delegate imageNode:self didLoadImage:result info:info];
+          [delegate imageNode:self didLoadImage:result info:info];
         } else if (_delegateFlags.delegateDidLoadImage) {
           ASUnlockScope(self);
-          [_delegate imageNode:self didLoadImage:result];
+          [delegate imageNode:self didLoadImage:result];
         }
       }
     }
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -365,11 +365,6 @@ static std::atomic_bool _useMainThreadDelegateCallbacks(true);
         [self _locked__setImage:result];
         _imageLoaded = YES;
         
-        id<ASNetworkImageNodeDelegate> delegate;
-        {
-          delegate = _delegate;
-        }
-        
         // Call out to the delegate.
         if (_delegateFlags.delegateDidLoadImageWithInfo) {
           ASUnlockScope(self);
@@ -640,6 +635,9 @@ static std::atomic_bool _useMainThreadDelegateCallbacks(true);
   [self lock];
     __weak id<ASNetworkImageNodeDelegate> delegate = _delegate;
     BOOL delegateDidStartFetchingData = _delegateFlags.delegateDidStartFetchingData;
+    BOOL delegateWillLoadImageFromCache = _delegateFlags.delegateWillLoadImageFromCache;
+    BOOL delegateWillLoadImageFromNetwork = _delegateFlags.delegateWillLoadImageFromNetwork;
+    BOOL delegateDidLoadImageFromCache = _delegateFlags.delegateDidLoadImageFromCache;
     BOOL isImageLoaded = _imageLoaded;
     NSURL *URL = _URL;
     id currentDownloadIdentifier = _downloadIdentifier;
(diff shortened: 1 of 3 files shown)


Source/ASNetworkImageNode.mm
```diff
@@ -335,6 +343,10 @@ - (void)displayWillStartAsynchronously:(BOOL)asynchronously
 {
   [super displayWillStartAsynchronously:asynchronously];
   
+  if (_delegateFlags.delegateWillStartDisplayAsynchronously) {
```
maicki: We don't grab a lock here to access the `_delegateFlags` and the `_delegate` here

Source/ASNetworkImageNode.mm
```diff
@@ -773,18 +785,31 @@ - (void)_lazilyLoadImageIfNecessary
           }
           
           if ([imageContainer asdk_image] == nil && self->_downloader != nil) {
+            if (self->_delegateFlags.delegateWillLoadImageFromNetwork) {
+              [delegate imageNodeWillLoadImageFromNetwork:self];
```
maicki: All of this methods are called with the lock held. This can cause deadlocks. We should at least document this behavior if we cannot get around this.

Source/ASNetworkImageNode.mm
ernestmama: will update

Source/ASNetworkImageNode.mm
ernestmama: I don't think these are locked. I tried putting a ASUnlockScope which return in an assertion failure.

nguyenhuy: This is inside a completion block and the lock is no longer held by the time it's executed. That's why we grabbed the lock before checking the cache sentinel a few line above. So we don't need to unlock here, *but* we need to grab the lock before accessing `_delegateFlags`. We can do so at the beginning of this method, similar to how we the `_delegateFlags.delegateDidStartFetchingData` flag.

Source/ASNetworkImageNode.mm
```diff
@@ -773,18 +797,31 @@ - (void)_lazilyLoadImageIfNecessary
...
+            }
             [self _downloadImageWithCompletion:^(id<ASImageContainerProtocol> imageContainer, NSError *error, id downloadIdentifier, id userInfo) {
               finished(imageContainer, error, downloadIdentifier, ASNetworkImageSourceDownload, userInfo);
             }];
           } else {
+            if (self->_delegateFlags.delegateDidLoadImageFromCache) {
```
nguyenhuy: Grab the flag at the beginning of this method and use it here.

Source/ASNetworkImageNode.mm
```diff
@@ -773,18 +797,31 @@ - (void)_lazilyLoadImageIfNecessary
...
+              [delegate imageNodeDidLoadImageFromCache:self];
+            }
             as_log_verbose(ASImageLoadingLog(), "Decached image for %@ img: %@ url: %@", self, [imageContainer asdk_image], URL);
             finished(imageContainer, nil, nil, ASNetworkImageSourceAsynchronousCache, nil);
           }
         };
+        
+        if (self->_delegateFlags.delegateWillLoadImageFromCache) {
```
nguyenhuy: Same here.

Source/ASNetworkImageNode.mm
```diff
@@ -773,18 +797,31 @@ - (void)_lazilyLoadImageIfNecessary
...
+          [delegate imageNodeWillLoadImageFromCache:self];
+        }
         [_cache cachedImageWithURL:URL
                      callbackQueue:[self callbackQueue]
                         completion:completion];
       } else {
+        if (self->_delegateFlags.delegateWillLoadImageFromNetwork) {
```
nguyenhuy: And here.

Source/ASNetworkImageNode.mm
```diff
@@ -346,14 +365,19 @@ - (void)displayWillStartAsynchronously:(BOOL)asynchronously
         [self _locked__setImage:result];
         _imageLoaded = YES;
         
+        id<ASNetworkImageNodeDelegate> delegate;
+        {
+          delegate = _delegate;
+        }
```
nguyenhuy: Just use cached `delegate` above.

Source/ASNetworkImageNode.mm
```diff
@@ -335,6 +343,17 @@ - (void)displayWillStartAsynchronously:(BOOL)asynchronously
 {
   [super displayWillStartAsynchronously:asynchronously];
   
+  id<ASNetworkImageNodeDelegate> delegate;
+  BOOL notifyDelegate;
+  {
+    ASLockedSelf();
```
nguyenhuy: Should be `ASLockScopeSelf`.

nguyenhuy:
@maicki Can you take another look? Thanks!

ernestmama:
Thanks @maicki and @nguyenhuy for helping on the Lock issue!

Source/ASNetworkImageNode.mm
```diff
@@ -335,6 +343,17 @@ - (void)displayWillStartAsynchronously:(BOOL)asynchronously
 {
   [super displayWillStartAsynchronously:asynchronously];
   
+  id<ASNetworkImageNodeDelegate> delegate;
+  BOOL notifyDelegate;
+  {
```
maicki: It's fine to grab the `delegate` in here with the lock held. One problem that will occur, what is not new and in fact we have it in almost all of our unlocked delegate calls, if the delegate needs more information from the node the information could already be out of date. Especially if you use this methods for logging this could introduce inconsistencies.

This is not a new problem though and we think about different solutions to find some general solution for this though.

maicki:
@ernestmama @nguyenhuy Added one comment. This should not be a blocker though as in fact we have this problem in other places too.

Source/ASNetworkImageNode.mm
nguyenhuy: Yeah, it's not a new problem, and I'd say delegates should rely on the data being passed via the delegate method. Not an ideal solution, but it's a fine trade-off.

I'm gonna land this PR now. Thanks for reviewing.

Issue #1239, opened 2018-11-16
Occasionally a crash occurs when reloading a table node


Kofktu:
// If you're looking for help, please consider joining our slack channel:
// http://asyncdisplaykit.org/slack (we'll be updating the name to Texture soon)

// The more information you include, the faster we can help you out! 
// Please include: a sample project or screenshots, code snippets
// Texture version, and/or backtraces for any crashes (> bt all). 
// Please delete these lines before posting. Thanks!

Intermittent crash occurs during reload after API communication.
Check the crash log.
ASDataController.updateWithChangeSet:
SIGKILL keeps waiting indefinitely due to dispatch_group_wait.

How can you fix it?

<img width="1133" alt="2018-11-16 12 29 55" src="https://user-images.githubusercontent.com/1860205/48596267-9ee7e380-e99b-11e8-8e43-086c67dfa0a3.png">

![2018-11-16 12 29 03](https://user-images.githubusercontent.com/1860205/48596272-a4452e00-e99b-11e8-9f70-dd3539336b52.png)

Kofktu:
PR [https://github.com/TextureGroup/Texture/pull/1251](https://github.com/TextureGroup/Texture/pull/1251)

PR #1251, opened 2018-11-23
issue-1239-crash occurs when reloading a table node

Kofktu:
Issue [https://github.com/TextureGroup/Texture/issues/1239](https://github.com/TextureGroup/Texture/issues/1239)

[issue-1239] Occasionally a crash occurs when reloading a table node

I checked where the crash occurred.

```
dispatch_group_wait(_editingTransactionGroup, DISPATCH_TIME_FOREVER);
```

This was caused by the code continuing to create a standby state.
at the bottom of the function 

```objective-c
- (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet {
    ...

    if (_usesSynchronousDataLoading) {
        [self waitUntilAllUpdatesAreProcessed];
    }
}
```

```objective-c
- (void)_scheduleBlockOnMainSerialQueue:(dispatch_block_t)block
{
  dispatch_group_wait(_editingTransactionGroup, DISPATCH_TIME_FOREVER);
  [_mainSerialQueue performBlockOnMainThread:block];
}

```

If _usersSyncDataLoading 
The waitUntilAllUpdated function was executed.
We found the same code in it.

So when I removed the code and tested it, there was no problem.
The problem of waiting indefinitely has been solved.

In fact, it is not clear if the code is required to have the `updateWithChangeSet:` function called, regardless of the `_usersSyncousDataLoading` value.
Please verify.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -2390,4 +2390,12 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   return;
 }
 
+#pragma mark - Accessibility overrides
+
+- (NSArray *)accessibilityElements
+{
+  [self waitUntilAllUpdatesAreCommitted];
+  return [super accessibilityElements];
+}
+
 @end
--- a/Source/ASNodeController+Beta.h
+++ b/Source/ASNodeController+Beta.h
@@ -10,18 +10,21 @@
 #import <AsyncDisplayKit/ASDisplayNode.h>
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h> // for ASInterfaceState protocol
 
+NS_ASSUME_NONNULL_BEGIN
+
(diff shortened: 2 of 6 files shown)


GeekTree0101:
ref: https://github.com/TextureGroup/Texture/pull/334

appleguy:
Thanks for reporting! I'm fairly sure it is not safe to remove this entirely though, as if there is a queue of multiple edit operations, other aspects of the code can't work safely.

Are you sure your app's data source is handling the difference between presentation & model index paths? The ASTableView / UIKit / presentation index paths may differ from the ASTableNode / Texture / model index paths at any given time. If you call across them without using the index path conversion methods when required, crashes will result.

Your issue could be unrelated to this and may indeed be a framework bug, but I wanted to make sure the two-index-space system is familiar and you feel confident in the app code.

strangeliu:
ref: #1266

PR #1252, opened 2018-11-26
Get StaticMutex back

nguyenhuy:
This reverts part of 5c9815f. Using `Mutex` for global C++ variables causes Texture experiments to be triggered before `main()`. Our (internal) experiment framework isn't ready that early in the process.

We can get rid of `StaticMutex` once the unfair lock experiment is shipped.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e7acb78eca808546edd12e3fb5806534/raw/4f6805d0b0faaa4efa0e887327e5e11efc99d572/log.txt)

Adlai-Holler:
Huh, I was under the impression that these new's shouldn't be executed until the variable is accessed ("magic statics" in C11). Could you post the stack trace where they're running early?

nguyenhuy:
Sure, here is the stack trace

screen shot 2018-11-26 at 10 43 44 am

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ec3e39f01ddd87d017849c56959910c4/raw/72c1628590ca0bb245136426d4cc2e34fa26189a/log.txt)

nguyenhuy:
These are run by dlyd before main().

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/70cb32de549b3918c48b1d777a2f48ad/raw/9b69b5d1ffeccf0fd758f9bc158deddab18316f6/log.txt)

Adlai-Holler:
Oh, right, magic statics only work for function-local statics not global variables. How about instead, we move the static variables into the functions in which they're used? That'll make them lazily generated. Currently each one is only used in one function.

Performance will be better if we use dispatch_once than if we rely on lazy statics, so:

```objc
- (void)myMethod {
static ASDN::Mutex *cacheMutex;
static dispatch_once_t onceToken;
dispatch_once(&onceToken, ^{
  cacheMutex = new ASDN::Mutex();
});

// is faster than

static auto *cacheMutex = new ASDN::Mutex();

// because the latter actually uses a global mutex around all accesses to all function-local statics.
}
```

nguyenhuy:
@Adlai-Holler Good call. I updated the call sites to use `dispatch_once`. Mind taking another look?

> because the latter actually uses a global mutex around all accesses to all function-local statics.

Ohh, good to know!

nguyenhuy:
@appleguy The lifecycle was a problem because `ASDN::Mutex` used to be a struct. It should be fine now, and we haven't seen any crashes related to this change.

PR #1254, opened 2018-11-27
Use interface state to manage image loading in ASMultiplexImageNode

nguyenhuy:
Similar to #1172 but for ASMultiplexImageNode


appleguy:
@nguyenhuy I wonder if we should flush the pending interface state here? It seems like in some cases like a displayIfNeeded / recursivelyEnsureDisplay preparing for an animation, skipping this call or flush could lead to the cache not being checked.

PR #1255, opened 2018-11-27
[ASMultiplexImageNode] Avoid holding on to its lock while calling out to downloader

nguyenhuy:


nguyenhuy:
I just happened to look at this method and noticed that we're doing it differently in [`ASNetworkImageNode`](https://github.com/TextureGroup/Texture/blob/master/Source/ASNetworkImageNode.mm#L363).

Issue #1161, opened 2018-10-04
accessibilityElements not cleared on relayout


ay8s:
When not using Yoga a display views accessibilityElements isn't cleared when a node's layout is changed. If you add nodes during the layout transition, new nodes aren't accessible. 

Looks like Yoga works correctly with calculateLayoutFromYogaRoot: implementing the following...
```
  // Reset accessible elements, since layout may have changed.
  ASPerformBlockOnMainThread(^{
    [(_ASDisplayView *)self.view setAccessibilityElements:nil];
  });
```

Curious where is best to call similar logic in the non Yoga layout code?

ay8s:
For now I have commented out... https://github.com/TextureGroup/Texture/blob/465abb1ded1777e7573a65e70df0ef0a2bda0df4/Source/Details/_ASDisplayViewAccessiblity.mm#L262-L267

PR #1260, opened 2018-12-01
invalidate a11y 

wsdwsd0829:
The invaidateAccessibilityElements is needed for following cases:
1. when there is a accessibilityContainer in the tree we need to make sure it's invalidated whenever insert a view or layer. 
2. If no accessibilityContainer involved, we still need to invalidate the first ancestor view to recollect new nodes. 
3. Whenever layer backed nodes changes themselves like attributedText change. 

They all covered in tests. 
Downside: it involves climbing the tree two times for 1 & 2. 
With all the work I still think it's better to just remove the cache (_accessibilityElements) in _ASDisplayView.

### DIFF
--- a/Source/Details/_ASDisplayView.mm
+++ b/Source/Details/_ASDisplayView.mm
@@ -285,6 +285,28 @@ static _ASDisplayViewMethodOverrides GetASDisplayViewMethodOverrides(Class c)
 #endif
 }
 
+- (void)insertSubview:(UIView *)view atIndex:(NSInteger)index {
+  [super insertSubview:view atIndex:index];
+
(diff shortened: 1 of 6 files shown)


Source/Details/_ASDisplayView.mm
```diff
@@ -285,6 +285,28 @@ - (void)addSubview:(UIView *)view
 #endif
 }
 
+- (void)insertSubview:(UIView *)view atIndex:(NSInteger)index {
```
wsdwsd0829: this will be called instead of addSubview above.

ay8s:
Related: https://github.com/TextureGroup/Texture/issues/1161

PR #1261, opened 2018-12-01
Move assertions so they are valid.

wiseoldduck:
If PIN_ANIMATED is enabled, the creation of a PINRemoteImageManager
will create a sharedDownloader also. Thus we cannot assert that it
doesn't exist. We will move this assertion up to the methods that
create the preconfiguredSharedManager and before said manager is
allocated.

### DIFF
--- a/Source/Details/ASPINRemoteImageDownloader.mm
+++ b/Source/Details/ASPINRemoteImageDownloader.mm
@@ -115,6 +115,7 @@ static PINRemoteImageManager *sharedPINRemoteImageManager = nil;
 
 + (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfiguration *)configuration
 {
+  NSAssert(sharedDownloader == nil, @"Singleton has been created and session can no longer be configured.");
   PINRemoteImageManager *sharedManager = [self PINRemoteImageManagerWithConfiguration:configuration imageCache:nil];
   [self setSharedPreconfiguredRemoteImageManager:sharedManager];
 }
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5e5af890affb946f0b8ffdae18319e9f/raw/096fefb044e798649557936b770364d9cf97c407/log.txt)

Source/Details/ASPINRemoteImageDownloader.mm
```diff
@@ -131,7 +133,6 @@ + (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfigurati
 + (void)setSharedPreconfiguredRemoteImageManager:(PINRemoteImageManager *)preconfiguredPINRemoteImageManager
 {
   NSAssert(preconfiguredPINRemoteImageManager != nil, @"setSharedPreconfiguredRemoteImageManager requires a non-nil parameter");
-  NSAssert(sharedDownloader == nil, @"Singleton has been created and session can no longer be configured.");
```
nguyenhuy: But removing it here means I can get away with calling this method directly, even after a singleton was created? Having said that, I don't care that much tbh.

maicki:
@nguyenhuy Could you please restart the CI. Thanks!

Source/Details/ASPINRemoteImageDownloader.mm
wiseoldduck: It's true (if you managed to create a `sharedDownloader` but without a `sharedPINRemoteImageManager`, which will assert below) you could make your way through here and miss the asserts. It's definitely wrong as-is though.

PR #1262, opened 2018-12-01
disable a11y cache

wsdwsd0829:
I think all tests would fail otherwise.


Tests/ASDisplayViewAccessibilityTests.mm
```diff
@@ -82,4 +91,173 @@ - (void)testThatContainerAccessibilityLabelOverrideStopsAggregation {
                         [node.view.accessibilityElements.firstObject accessibilityLabel]);
 }
 
+- (void)testAccessibilityLayerbackedNodesOperationInContainer {
```
wsdwsd0829: most tests doing same thing: container(text1 , text2 (text3))

PR #1265, opened 2018-12-04
Wrap supplementary node blocks to enable resizing them.

wiseoldduck:
Most ASCellNodeBlocks are wrapped by ASCollectionView.mm to add an `interactionDelegate` as well as calling `enterHierarchyState` on the node and setting its transform to a reflection if `inverted` is set on the collectionView.

This PR adds this behavior to supplementary nodes as well.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -2019,7 +2019,7 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   __weak __typeof__(self) weakSelf = self;
   return ^{
     __typeof__(self) strongSelf = weakSelf;
-    ASCellNode *node = (block ? block() : cell);
+    ASCellNode *node = block();
     ASDisplayNodeAssert([node isKindOfClass:[ASCellNode class]],
                         @"ASCollectionNode provided a non-ASCellNode! %@, %@", node, strongSelf);
 
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -2015,7 +2015,22 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     block = ^{ return cell; };
   }
 
-  return block;
+  // Wrap the node block
+  __weak __typeof__(self) weakSelf = self;
+  return ^{
+    __typeof__(self) strongSelf = weakSelf;
+    ASCellNode *node = (block ? block() : cell);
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4c68885e5b11f002c65c3a0eab0be376/raw/68e71083051d97f11a6fe7733d03be841fa2808a/log.txt)

Source/ASCollectionView.mm
```diff
@@ -2015,7 +2015,26 @@ - (ASCellNodeBlock)dataController:(ASDataController *)dataController supplementa
     block = ^{ return cell; };
   }
 
-  return block;
+  // Wrap the node block
+  BOOL disableRangeController = ASCellLayoutModeIncludes(ASCellLayoutModeDisableRangeController);
```
maicki: We may should remove that ;)

Source/ASCollectionView.mm
```diff
@@ -2015,7 +2015,22 @@ - (ASCellNodeBlock)dataController:(ASDataController *)dataController supplementa
     block = ^{ return cell; };
   }
 
-  return block;
+  // Wrap the node block
+  __weak __typeof__(self) weakSelf = self;
```
nguyenhuy: At one point we should add strongify & weakify macros.

Source/ASCollectionView.mm
```diff
@@ -2015,7 +2015,22 @@ - (ASCellNodeBlock)dataController:(ASDataController *)dataController supplementa
...
+  return ^{
+    __typeof__(self) strongSelf = weakSelf;
+    ASCellNode *node = (block ? block() : cell);
```
nguyenhuy: Isn't block is guaranteed to be non-nil after [this line](https://github.com/TextureGroup/Texture/pull/1265/files#diff-8b82b1556d1293f6f9789beb236c9271R2015)? Maybe remove that line to be consistent with the implementation of `dataController:nodeBlockAtIndexPath:`.

maicki: Concur with @nguyenhuy. The block should not be nil in here and the returning block is retaining the block too.

wiseoldduck: I thought you had to `copy` a block to actually retain it

wiseoldduck: nm I'm like stuck in 2010 😛

PR #1268, opened 2018-12-05
Forward hitTest:withEvent and piontInside:withEvent: to node within _ASCollectionViewCell

maicki:
This replaces #1268 that was unfortunately not enough to only forward the calls to the node. Extra checks need to be in place in case `UICollectionView`s animated updating API is used and hidden `UICollectionViewCell`'s are still within `UICollectionView`s subviews.

Thanks @stremsdoerfer for the initial PR!

### DIFF
--- a/Source/Details/_ASCollectionViewCell.mm
+++ b/Source/Details/_ASCollectionViewCell.mm
@@ -8,6 +8,7 @@
 //
 
 #import <AsyncDisplayKit/_ASCollectionViewCell.h>
+#import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
 
 #import <AsyncDisplayKit/ASCellNode+Internal.h>
 #import <AsyncDisplayKit/ASCollectionElement.h>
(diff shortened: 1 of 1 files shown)


Source/Details/_ASCollectionViewCell.mm
```diff
@@ -94,6 +95,29 @@ - (void)layoutSubviews
...
+  /**
+   * The documentation for hitTest:withEvent: on an UIView explicitly states the fact that:
+   * it ignores view objects that are hidden, that have disabled user interactions, or have an
+   * alpha level less than 0.01.
+   * To be able to determine if the collection view cell should skip going further down the tree
+   * based on the states above we use a valid point within the cells bounds and check the
+   * superclass hitTest:withEvent: implementation. If this returns a valid value we can go on with
+   * checking the node as it's expected to not be in one of these states.
+   */
+  if (![super hitTest:[self convertPoint:self.center fromView:self.superview] withEvent:event]) {
```
Adlai-Holler: Rather than converting the point from superview just to be simpler we should use `self.bounds.origin`

Issue #1270, opened 2018-12-06
ASTextNode2: truncation mode set in attributed text isn't respected


shamanskyh:
## Code sample:
```objc
NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
paragraphStyle.lineBreakMode = NSLineBreakByTruncatingMiddle;

ASTextNode2 *textNode = [[ASTextNode2 alloc] init];
textNode.attributedText = [[NSAttributedString alloc] initWithString:@"a really really long string" 
                                                          attributes:@{NSParagraphStyleAttributeName : paragraphStyle}];
```

## Expected results:
Text renders with truncation in middle when it does not fit within the constrained width.

## Actual results:
Text wraps when it does not fit within the constrained width.

I believe this is happening because `prepareAttributedString:isForIntrinsicSize:` calls `_truncationMode` which, in the above example, is never set. This conflicts with the docs that state `Setting a truncationMode in attributedString will override the truncation mode set here.`

## Possible fixes:
- Inspect any truncation modes set when `setAttributedText:` is called and set the underlying ivar.
- Change the header docs to reflect different behavior in ASTextNode2 if this is expected behavior.

shamanskyh:
I should add that the obvious workaround is to set the `truncationMode` property manually.

nguyenhuy:
Thanks for reporting this issue. I think we should strike to have the behavior described in the docs (which I assume is the same as ASTextNode's behavior).

PR #1272, opened 2018-12-06
Fix truncation precedence in ASTextNode2 (#1270)

shamanskyh:
Previously, setting attributed text with a truncation mode in ASTextNode2 wouldn't override the property which caused the truncation mode not to take affect. By setting the backing ivar in `setAttributedText:`, we can make the behavior match ASTextNode and what's in the docs.

### DIFF
--- a/Source/ASTextNode2.mm
+++ b/Source/ASTextNode2.mm
@@ -407,6 +407,12 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
     ASLayoutElementStyle *style = [self _locked_style];
     style.ascender = [[self class] ascenderWithAttributedString:attributedText];
     style.descender = [[attributedText attribute:NSFontAttributeName atIndex:attributedText.length - 1 effectiveRange:NULL] descender];
+
+    // Set our truncation ivar if a paragraph style is specified
+    NSParagraphStyle *paragraphStyle = [attributedText attribute:NSParagraphStyleAttributeName atIndex:attributedText.length - 1 effectiveRange:NULL];
+    if (paragraphStyle) {
+      _truncationMode = paragraphStyle.lineBreakMode;
+    }
   }
   
   // Tell the display node superclasses that the cached layout is incorrect now
--- a/Tests/ASTextNode2Tests.mm
+++ b/Tests/ASTextNode2Tests.mm
@@ -91,4 +91,27 @@
                 _textNode.defaultAccessibilityLabel, _attributedText.string);
 }
 
+- (void)testAttributedTextTruncationOverridesTruncationProperty
+{
+  // capture previous values, restore after testing
+  NSLineBreakMode prevMode = _textNode.truncationMode;
+  NSAttributedString *prevString = _textNode.attributedText;
+
+  // set the property to word-wrapping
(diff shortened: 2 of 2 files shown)


wiseoldduck:
This is a big improvement! Some trivial points:

1. We still won't support multiple paragraph style runs in a long attributed string,
2. This implementation causes the setting on the control-level ivar to change, which is divergent from how UILabel behaves.

But let's merge this for sure. I think there's another PR somewhere to correct the inverted `isTruncated` getter, so we will just see which one merges first.

maicki:
I think we should get some inspiration from `[UILabel lineBreakMode]`: https://developer.apple.com/documentation/uikit/uilabel/1620525-linebreakmode.

Changing the `truncationMode` while the user set's a new attributed string is not really obvious for the consumer of the API. It's a side effect that can lead to a lot of confusion and hard to find bugs if you are not aware what is happening. Furthermore it has the issues that @wiseoldduck pointed out.

@wiseoldduck You think we could adopt the same behavior as `UILabel` and `lineBreakMode`?

Furthermore we also should think about renaming `truncationMode` to `lineBreakMode` as we are using the `NSLineBreakMode` enum for the value and to match with UILabel, but that should be a different issue.

wiseoldduck:
I like the rename, it only inherited `truncationMode` from the old `ASTextNode`. (which, itself also used `NSLineBreakMode` so I don't know why _it_ was named that way).

I guess I liked this PR because it seemed like an incremental improvement. The true fix is to figure out what is going on in `ASTextLayout` (which is doing special casing galore around lineBreakMode) that is preventing the `lineBreakMode` from paragraph style runs being applied as the rest of the attributes are. Fixing it at that (`ASTextLayout`) layer would give us desired behavior without any of these drawbacks.

wiseoldduck:
Actually, further experimentation indicates to me that `UILabel` will also not pick up a `lineBreakMode` setting from an `NSAttributedString` - it must be set via the label property. So we are on parity with that behavior, and I would say should just leave it alone.

maicki:
@wiseoldduck You mean leave it alone in terms of not proceeding with this PR or leave the PR behavior as it is?

wiseoldduck:
I mean not proceed with this PR.

Yeah deprecation is always a PITA especially for something as minor as a name change, but it's a very simple deprecation path that can be supported with an alias for an indefinite period of time.

I think the more direct source of the name is the `CTLineTruncationType` in CoreText. I think historically that's why only those three modes (https://developer.apple.com/documentation/coretext/ctlinetruncationtype?language=objc) were correctly supported. But at the `UILabel` level (which is the `ASTextNode`/`2` analog) it becomes lineBreakMode and has six options.

maicki:
@nguyenhuy What are your thoughts?

nguyenhuy:
Great thoughts, @maicki @wiseoldduck. I agree that:
- We should be consistent with UILabel's behavior. Prior to this PR, I suggested @shamanskyh to stick to ASTextNode's behavior, but UILabel's behavior should be preferred.
- As such, let's not proceed with this PR and instead fix the root cause at ASTextLayout level.
- We should proceed with the name change. The migration path is straight-forward and I'd say we're overdue for a major release :)

shamanskyh:
Sounds good. Closing this PR for now.

PR #1273, opened 2018-12-07
Introduce ASCellLayoutMode

maicki:
`ASCellLayoutMode`s  are flexible settings for applications to improve the flexibility of integrating Texture within an app. 
Providing this mode allows for developers to debug potential issues or work towards resolving technical debt in their collection view data source, while ramping up asynchronous collection layout.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -63,6 +63,10 @@
   flowLayout ? flowLayout.property : default;                                                 \
 })
 
+// ASCellLayoutMode is an NSUInteger-based NS_OPTIONS field. Be careful with BOOL handling on the
+// 32-bit Objective-C runtime, and pattern after ASInterfaceStateIncludesVisible() & friends.
+#define ASCellLayoutModeIncludes(layoutMode) ((_cellLayoutMode & layoutMode) == layoutMode)
+
 /// What, if any, invalidation should we perform during the next -layoutSubviews.
 typedef NS_ENUM(NSUInteger, ASCollectionViewInvalidationStyle) {
   /// Perform no invalidation.
@@ -685,9 +689,10 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     if (wrapperNode.sizeForItemBlock) {
       return wrapperNode.sizeForItemBlock(wrapperNode, element.constrainedSize.max);
     } else {
-      // In this case, we should use the exact value that was stashed earlier by calling sizeForItem:, referenceSizeFor*, etc.
-      // Although the node would use the preferredSize in layoutThatFits, we can skip this because there's no constrainedSize.
-      return wrapperNode.style.preferredSize;
+      // In this case, it is possible the model indexPath for this element will be nil. Attempt to convert it,
+      // and call out to the delegate directly. If it has been deleted from the model, the size returned will be the layout's default.
+      NSIndexPath *indexPath = [_dataController.visibleMap indexPathForElement:element];
+      return [self _sizeForUIKitCellWithKind:element.supplementaryElementKind atIndexPath:indexPath];
     }
   } else {
     return [node layoutThatFits:element.constrainedSize].size;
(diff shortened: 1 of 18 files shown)


Source/ASCollectionViewProtocols.h
```diff
@@ -10,6 +10,41 @@
 #import <UIKit/UIKit.h>
 #import <AsyncDisplayKit/ASBaseDefines.h>
 
+typedef NS_OPTIONS(NSUInteger, ASCellLayoutMode) {
+  /** No options set. Default value. */
+  ASCellLayoutModeNone = 0,
```
Adlai-Holler: Document the default behavior.

Source/ASCollectionView.mm
```diff
@@ -879,9 +856,21 @@ Performing nested batch updates with super (e.g. resizing a cell node & updating
...
+    if (updates) {
+      updates();
+    }
+    [super reloadData];
+    if (completion) {
+      completion(YES);
+    }
+  } else {
+    [super performBatchUpdates:updates completion:completion];
+  }
```
Adlai-Holler: Why have this fork in two places i.e. here and also below at 1750 and 2214?

maicki: In the two places we have a different logic when to perform a batch update or reload data. In `rangeController:updateWithChangeSet:updates`: besides checking for the `ASCellLayoutModeAlwaysReloadData` flag we also checking some more conditions to decide if we should reload data or batch update.

That said I removed the "automatic" behavior within `_superPerformBatchUpdates:completion:` and we are now explicitly calling if else in places where we would like to differentiate between the calls. It was not really obvious from a caller perspective that this automatic behavior exists in the first place.

Source/ASCollectionViewProtocols.h
maicki: Tried to clarify the default values for `ASCellLayoutMode` flags a bit better.

maicki:
@Adlai-Holler Can you please give it another look. Thanks!

Source/ASCollectionView.mm
```diff
@@ -869,17 +846,29 @@ - (CGSize)_sizeForUIKitCellWithKind:(NSString *)kind atIndexPath:(NSIndexPath *)
   return size;
 }
 
+- (void)_superReloadData:(void(^)())updates completion:(void(^)(BOOL finished))completion
+{
+  if (updates) {
```
nguyenhuy: Seems like the `updates` block should be non-nil. At the very least, we want to swap the maps. If it's non-nil, we can mark it that way and skip the nil check? The main point is I don't want people to think that it can ever be nil.

Source/ASCollectionViewProtocols.h
```diff
@@ -10,6 +10,44 @@
...
+  */
+  ASCellLayoutModeAlwaysSync = 1 << 1,                // Default OFF
+  ASCellLayoutModeAlwaysAsync = 1 << 2,               // Default OFF
+  ASCellLayoutModeForceIfNeeded = 1 << 3,             // Deprecated, default OFF.
+  ASCellLayoutModeAlwaysPassthroughDelegate = 1 << 4, // Deprecated, default ON.
+  /** Instead of using performBatchUpdates: prefer using reloadData for changes for collection view */
+  ASCellLayoutModeAlwaysReloadData = 1 << 5,          // Default OFF
+  /** If flag is enabled nodes are *not* gonna be range managed. */
+  ASCellLayoutModeDisableRangeController = 1 << 6,    // Default OFF
+  ASCellLayoutModeAlwaysLazy = 1 << 7,                // Deprecated, default OFF.
```
nguyenhuy: It's strange that we need to add some flags that are deprecated right off the bat. Do we need them at all?

Source/ASSectionController.h
```diff
@@ -35,6 +35,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * Note: only called if nodeBlockForItemAtIndex: returns nil.
+ *
+ * @param index The index of the item.
+ *
+ * @return A node to display for the given item. This will be called on the main thread and should
+ *   not implement reuse (it will be called once per item).  Unlike UICollectionView's version,
+ *   this method is not called when the item is about to display.
+ */
+- (ASCellNode *)nodeForItemAtIndex:(NSInteger)index;
```
nguyenhuy: Can't find the implementation of this method.

Tests/ASCollectionViewTests.mm
```diff
@@ -1049,6 +1049,7 @@ - (void)testInitialRangeBounds
   UIWindow *window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
   ASCollectionViewTestController *testController = [[ASCollectionViewTestController alloc] initWithNibName:nil bundle:nil];
   ASCollectionNode *cn = testController.collectionNode;
+  cn.cellLayoutMode = ASCellLayoutModeAlwaysAsync;
```
nguyenhuy: Shouldn't we test the default behavior? Maybe add another test case if need to for this new mode.

Source/ASCollectionView.mm
maicki: It can be nil, if you just would like to call `[super reloadData]` with a completion block. I actually replaced all places where we call `[super reloadData]` with the call to `_superReloadData` so we have one central point to call in.

Source/ASCollectionViewProtocols.h
maicki: Yeah we need them for now at least. We will clean that up eventually.

Source/ASSectionController.h
maicki: `ASSectionController` is a protocol, so it's for clients to implement it. It's called within `collectionNode:nodeForItemAtIndexPath:`

Source/ASCollectionView.mm
nguyenhuy: Oh yeah, I overlooked those calls. 👍

Source/ASSectionController.h
nguyenhuy: lol, I'm stupid!

Tests/ASCollectionViewTests.mm
maicki: Good call! Added a test.

PR #1274, opened 2018-12-07
Disable a11y cache

wsdwsd0829:
Disable because a11y not update correctly as layer backed nodes inserted/removed/updated. 
A few other case will fail otherwise as shown in tests. 

Fixes:  #1161
Attempt to invalidate a11y (instead of disable cache) code is here: https://github.com/TextureGroup/Texture/pull/1260 while does not seem worth the effort.

### DIFF
--- a/Tests/ASDisplayViewAccessibilityTests.mm
+++ b/Tests/ASDisplayViewAccessibilityTests.mm
@@ -23,7 +23,8 @@
 
 @implementation ASDisplayViewAccessibilityTests
 
-- (void)setUp {
+- (void)setUp
+{
   ASConfiguration *config = [[ASConfiguration alloc] initWithDictionary:nil];
   config.experimentalFeatures = ASExperimentalDisableAccessibilityCache;
   [ASConfigurationManager test_resetWithConfiguration:config];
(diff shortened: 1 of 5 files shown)


Tests/ASDisplayViewAccessibilityTests.mm
```diff
@@ -14,12 +14,21 @@
...
+#import <AsyncDisplayKit/ASTextNode.h>
+#import "ASConfiguration.h"
+#import "ASConfigurationInternal.h"
 
 @interface ASDisplayViewAccessibilityTests : XCTestCase
 @end
 
 @implementation ASDisplayViewAccessibilityTests
 
+- (void)setUp {
```
nguyenhuy: Bracket on the following line please.

Tests/ASDisplayViewAccessibilityTests.mm
```diff
@@ -82,4 +91,173 @@ - (void)testThatContainerAccessibilityLabelOverrideStopsAggregation {
...
+  text3.attributedText = [[NSAttributedString alloc] initWithString:@"!!!!"];
+  text3.frame = CGRectMake(50, 400, 200, 100);
+  text3.layerBacked = YES;
+  [text2 addSubnode:text3];
+  [contianer layoutIfNeeded];
+  [contianer.layer displayIfNeeded];
+  NSArray<UIAccessibilityElement *> *updatedElements2 = contianer.view.accessibilityElements;
+  XCTAssertTrue([[updatedElements2.firstObject accessibilityLabel] isEqualToString:@"hello, world, !!!!"]);
+}
+- (void)testAccessibilityNonLayerbackedNodesOperationInContainer {
```
nguyenhuy: Same here. Also separate these methods by new lines.

Issue #23, opened 2017-04-18
automaticallyAdjustsContentOffset does not work anymore


Dandiccf:
I have to insert multiple rows in my ASTableNode on index position 0 without animation and maintaining the users' visible scroll position.

In the latest AsyncDisplayKit (2.2.1) it is working perfectly when using 
 - tableNode.view.automaticallyAdjustsContentOffset = true
 - tableNode.insertRows(at: prependIndexPaths, with: .none)

In Texture (2.3) I have following bugs
 - scroll position is jumping when inserting at index position 0  although using tableNode.view.automaticallyAdjustsContentOffset = true
 - the inserted rows have a transition animation although using UITableViewRowAnimation.none

Dandiccf:
it looks like there is no progress in this concern. Do you have it on your radar, oder is there an easy workaround?  thx

Dandiccf:
am I the only one who uses 'automaticallyAdjustsContentOffset'?

It is still not working in 2.3.3.

Any comments on that, or am I doing something completely wrong?

thx

nixzhu:
Same here, automaticallyAdjustsContentOffset is not working anymore.

cesteban:
I believe it stopped working with #317. By the time the content offset adjustment is applied, the visible nodes are still not updated (`tableView:willDisplayCell:forRowAtIndexPath:` won't be called until next layout pass), so the calculation of the offset of the first visible node in `endAdjustingContentOffsetAnimated:` just yields the same result than in `beginAdjustingContentOffset`, and nothing is done. Forcing a layout pass before `endAdjusting...` fixes the problem. We have been using this trick combined with manual contentOffset adjusting for weeks (we are in Texture 2.5). I would have a look later on what we have and consider sending a PR with this change.

I believe this is also (vaguely) related with the discussion about visibility on #788.

aalenliang:
Any news on this issue?

made2k:
If someone finds this issue and is still seeing that deleting rows doesn't work with this, I've found a few things after this change set had been merged.

When deleting items from a ASTableNode, the internal `_ASHierarchyChangeSet` always sets its `animated` property to true. I noticed there's a `TODO` on that which may fix that issue.

https://github.com/TextureGroup/Texture/blob/master/Source/Private/_ASHierarchyChangeSet.h#L108

The key issue that causes is that Texture will not work like UIKit's UITableView when the changes are animated. It just gives up on adjusting the content inset.

[...]

PR #1278, opened 2018-12-11
Set automaticallyAdjustsContentOffset to ASTableView when view is load

strangeliu:
Before ASTableView is loaded, the property `automaticallyAdjustsContentOffset` is stored on `pendingState` but never set to view.
This PR may fix this issue #23.


strangeliu:
@nguyenhuy I have updated the test suite and it passed

nguyenhuy:
Awesome. Thanks again. I’ll merge after CI finishes.

PR #1283, opened 2018-12-13
Copy yogaChildren in accessor method. Avoid using accessor method internally

maicki:


Source/ASDisplayNode+Yoga.mm
```diff
@@ -58,7 +58,7 @@ - (void)setYogaChildren:(NSArray *)yogaChildren
 
 - (NSArray *)yogaChildren
 {
-  return _yogaChildren;
+  return [_yogaChildren copy] ?: @[];
```
wiseoldduck: We already lock on the setter, why not go ahead and lock here and declare the property `atomic`?

Source/ASDisplayNode+Beta.h
```diff
@@ -189,7 +189,7 @@ AS_EXTERN void ASDisplayNodePerformBlockOnEveryYogaChild(ASDisplayNode * _Nullab
 @interface ASDisplayNode (Yoga)
 
 // TODO: Make this and yogaCalculatedLayout atomic (lock).
-@property (nullable, nonatomic) NSArray *yogaChildren;
+@property (nonatomic, copy) NSArray *yogaChildren;
```
wiseoldduck: This isn't really `copy`

Adlai-Holler: why?

wiseoldduck: I don't think there is a keyword to specify copy-on-read, this one is supposed to mean copy-on-set

Adlai-Holler: It's both

wiseoldduck: You mean the keyword means both, or this class is somehow both? Oh the second, yes, I see we do basically copy after all.

Adlai-Holler: Here's the official doc, it's pretty unclear because they assume your ivar is immutable. https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html

If your ivar is mutable and the property is atomic, there's no other choice than copying on read. Our property isn't atomic, but it should be and we badly assume that it is.

Source/ASDisplayNode+Yoga.mm
Adlai-Holler: Because this diff was copied from an 11th-hour internal hotfix! But you're right, we should do it now. @maicki mind making the change and landing it with this?

Source/ASDisplayNode+Beta.h
wiseoldduck: The official doc doesn't seem unclear at all: "Copy Properties Maintain Their Own Copies"

But in our case, we _are_ copying in our setter too (which I missed at first), so I agree `copy` is a correct attribute.

Source/ASDisplayNode+Yoga.mm
wiseoldduck: I knew I'd been here before...

Source/ASDisplayNode+Beta.h
Adlai-Holler: The docs are unclear unless you read them to mean that we are not allowed to use a mutable ivar i.e. we have to "maintain our own copy" as an immutable variable at all times. If so, then screw the docs.

wiseoldduck: The documentation happens to use immutable copies in the examples, but I don't see anything stating or implying that's required (a mutable backing var, as we can see, becomes more complicated, which is why it wouldn't have been great for example code anyway). So no screwing required!

Adlai-Holler: Because it's assumed by the example code, there's no guidance on the semantics of getters that are backed by mutable ivars. That's the problem with using it as an example!

maicki:
@Adlai-Holler @wiseoldduck I don't remember where we land on, but could you give that another look please.

Issue #1155, opened 2018-10-02
Clearing data frequently while delegate / dataSource is changing within collection view can cause crashes


maicki:
silently clearing its data while the collection view is still being used, without notifying the backing UICollectionView, puts it out-of-sync and causes mayhem next time the collection view processes a batch update. If you look at the stack trace closely, you'll notice that the crash doesn't occur on the same run loop that `clearData` was called, making it extremely tricky to investigate and identify the root cause.

![46322275-c413c480-c59c-11e8-9ede-172f435e2cde](https://user-images.githubusercontent.com/25288/46368023-020ff780-c634-11e8-869a-67b8a23dd233.png)

maicki:
The original reason for why this clearing data was added is to ensure that the interfaceState is zero’d out in the collection dealloc case.

@nguyenhuy What do you think, would it be enough to call reloadData on the collection itself in this case, or at least invalidateLayout? After really making sure the reason for the crash in the stack trace is clearing the data as we talked in the past we should look what the root cause is in the first place.

nguyenhuy:
I kind of get the issue. But I think it would be best to have a stacktrace when it occurs.

I can imagine that when the cell node are deallcoated at the end of the hosting collection view lifecycle, their interface state aren't updated properly because we never do so. A proper fix would be to directly/indirectly update those states. Directly means walking through all (visible) cell nodes and set manually reset the state. Indirectly means somehow trigger the collection view to update its dataset and ranges, for example by calling reload data as you suggested above.

PR #1284, opened 2018-12-14
Add a way to opt out of always-clear-data behavior in ASCollectionView and ASTableView

nguyenhuy:
Through crashes in the past, we established that:
1. We can't clear data if the collection/table view is still being used. (#1155)
2. We may clear data during deallocation, and even so, we want to do it if user is in `ASExperimentalClearDataDuringDeallocation` experiment.

As such, the logic change introduced in #1200 is not correct. It causes `-clearData` to **always** be executed if user is not in the experiment (which is true by default).

screen shot 2018-12-13 at 9 41 24 pm

I suspect what you wanted to do is setting up an experiment and hook up with Texture. And even then, we need to fix the implementation of `clearData`. I'm seeing a crash in production (1st screenshot below) which looks like a memory issue when `self` (i.e the collection view) is gone and accessing the editing queue (i.e `self->_editingTransactionGroup`) after that is problematic (see 2nd screenshot).

screen shot 2018-12-13 at 10 16 37 pm

screen shot 2018-12-13 at 10 08 15 pm

I think `clearData` needs to *avoid* calling `-waitUntilAllUpdatesAreProcessed` altogether. Instead, it needs to:
1. Destroy the editing queue.
2. Destroy the main serial queue.
3. Clear out the visible and pending maps.

### DIFF
--- a/Schemas/configuration.json
+++ b/Schemas/configuration.json
@@ -22,7 +22,7 @@
                     "exp_dealloc_queue_v2",
                     "exp_collection_teardown",
                     "exp_framesetter_cache",
-                    "exp_clear_data_during_deallocation"
+                    "exp_skip_clear_data"
                 ]
     		}
 		}
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -581,10 +581,8 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 {
   ASDisplayNodeAssertMainThread();
 
-  if (_asyncDataSource == nil && _asyncDelegate == nil) {
-    if (! ASActivateExperimentalFeature(ASExperimentalSkipClearData)) {
-      [_dataController clearData];
-    }
+  if (_asyncDataSource == nil && _asyncDelegate == nil && !ASActivateExperimentalFeature(ASExperimentalSkipClearData)) {
+    [_dataController clearData];
   }
 }
 
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -581,9 +581,12 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 {
   ASDisplayNodeAssertMainThread();
 
-  if (_asyncDataSource == nil && _asyncDelegate == nil && _isDeallocating
+  if (_asyncDataSource == nil && _asyncDelegate == nil
       && ASActivateExperimentalFeature(ASExperimentalClearDataDuringDeallocation)) {
-    [_dataController clearData];
+    if (_isDeallocating) {
+      [_dataController clearData];
+    }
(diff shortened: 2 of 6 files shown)


maicki:
I don't think this is the right logic either as it would change the initial behavior. In this case we would have to enable the `ASExperimentalClearDataDuringDeallocation` experimental feature to get the initial behavior we had in the first place, calling `[_dataController clearData];` every time.

I would suggest we add an overall experimental feature and change the logic to something like this:
```
if (_asyncDataSource == nil && _asyncDelegate == nil) {
    // Overall experiment to test if different clearing behaviors will fix some issues
    if (ASActivateExperimentalFeature(ASExperimentalTestClearDuringDeallocation)) {
        // If this experiment is activated we only clear if _isDeallocating == TRUE
        if (ASActivateExperimentalFeature(ASExperimentalClearDataDuringDeallocation)) {
            if (_isDeallocating) {
                [_dataController clearData];          
            }
        // Here is room for some other cases we would like to test
        } else {
            // In this case we would not clear at all
        }
    } else {
        // Default behavior as before
        [_dataController clearData];  
    }
}
```

nguyenhuy:
The thing is we don’t want to keep the original behavior (i.e always call clearData). The implementation of clearData is problematic, both while the collection view is being used and when it’s being deallocated.

nguyenhuy:
Ok, I've just updated the logic to check the experiment first. It means you'll have room to test other cases in which clearing data is needed. I still think it should be opt-in as we don't want to keep the original problematic behavior.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/95796e257d29ed34ab5f3ddd692514e0/raw/0a3730c6d486012430b0beb107286097a2264aae/log.txt)

maicki:
@nguyenhuy Would be great to have this PR not change the standard behavior where it clears by default. Applications still rely on this behavior and we can not make it opt in. We should make the new behavior (clear only on deallocation) opt in it until all bigger clients of Texture can agree what the final behavior we should strive for.

nguyenhuy:
@maicki I updated the diff to always clear data (the original behavior) unless a client explicitly opts out (and Pinterest is going to). This essentially reverts all recent changes to get us back to #1136. LGTY?

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/7dd5c90f4deb9edce295a9a3bb59bd3c/raw/82c7f374ddc1e39ebe418a12a43b3dfa2c50f4a1/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/6b3f2f9d2195282e7a95f1f487113d96/raw/02df50ece505f60ac20b4b305f25ef1cbe01433e/log.txt)

PR #1150, opened 2018-09-29
ASTextNode2 rendering corrections.

wiseoldduck:
ASTextNode2 was only setting truncationMode (lineBreakMode) on existing paragraph styles in attributedString (thus having no effect for the two non-truncating modes if there were not any existing paragraph style runs).

ASTextLayout (essentially YYTextLayout) was not rendering the two non-tail truncation lineBreakModes correctly. There's not much history on github but it appears to me that it was set up correctly at one time and then some additional code was added for unclear reasons that assumed any truncation was at the end of the string.

This commit corrects both issues.

PR #1287, opened 2018-12-18
Allow truncation mode (head/tail/middle) for attribute string

ernestmama:
Truncation mode (Head/Tail/Middle) override has not been applied when preparing for attributed string. It was default to WordWrapping. Those truncation mode is currently rendering correctly. This change is to let the override happens for all lineBreakMode.


ernestmama:
This is the commit which added support for the truncation: https://github.com/TextureGroup/Texture/pull/1150

ernestmama:
@wiseoldduck What do you think? It seems like your previous commit fix the truncation problem. Is it safe to remove the switch?

wiseoldduck:
I do not believe it is. The ASTextLayout class uses the "inner" mode to break up the lines, and then applies the "real" mode to the final line.

nguyenhuy:
From @wiseoldduck's comment, it seems like we should close this issue. Feel free to reopen it. Thanks all.

PR #1289, opened 2018-12-18
Remove unnecessary dispatch_once calls.

dmaclach:
Dispatch once is not necessary for most static variable initialization in C++.
C++ already guarantees static initialization safety.


dmaclach:
Abort. Thanks to Huy I learned that static initialization is slower than dispatch_once.

PR #1290, opened 2018-12-18
Early exit editing transaction queue consumption during deallocation

mikezucc:
This is in response to a hard to reproduce crash where `performBatchUpdates` is sent on a valid pointer, however the internal UIKit collection view components such as _UICollectionViewData later reference a nil shared indexPath store during their many block invokes for a collection view update. Looking in hopper and fabric, these crashes are spread across many points in the _UICollectionView interfaces, but they all crash on using the indexPath as a key for a dictionary. This is what leads me to think that the indexPaths, which are referenced through to the parent _UICollectionView or _UICollectionViewData have been nilled, where there is still a long series of block invocations happening

This could be attributed the editingQueue still consuming its tasks while the collection view is being async deallocated. We want to early return at editing transation block invocation time since we can not destroy queued blocks.

### DIFF
--- a/Source/Details/ASDataController.h
+++ b/Source/Details/ASDataController.h
@@ -113,6 +113,19 @@ AS_EXTERN NSString * const ASCollectionInvalidUpdateException;
  */
 - (void)dataController:(ASDataController *)dataController updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet updates:(dispatch_block_t)updates;
 
+/**
+ * Communicate to the child whether or not its parent is scheduled for deallocation.
+ *
+ * @discussion This is in response to a hard to reproduce crash where
+ * `performBatchUpdates` is sent on a valid pointer, however the internal UIKit
+ * collection view components such as _UICollectionViewData later reference a nil shared
+ * indexPath store during their many block invokes for a colleciton view update.
+ * This could be attributed the editingQueue still consuming its tasks
+ * while the collection view is being async deallocated. We want to early return
+ * at editing transation block invocation time since we can not destroy queued blocks;
+ */
+- (BOOL)isDeallocating;
+
 @end
 
 @protocol ASDataControllerLayoutDelegate <NSObject>
--- a/Source/Details/ASDataController.mm
+++ b/Source/Details/ASDataController.mm
@@ -629,6 +629,10 @@ typedef void (^ASDataControllerSynchronizationBlock)();
   Class<ASDataControllerLayoutDelegate> layoutDelegateClass = [self.layoutDelegate class];
   ++_editingTransactionGroupCount;
   dispatch_group_async(_editingTransactionGroup, _editingTransactionQueue, ^{
+    if (_delegate.isDeallocating) {
+      as_log_debug(ASCollectionLog(), "Editing queue consumed during dealloc");
+      return;
+    }
     __block __unused os_activity_scope_state_s preparationScope = {}; // unused if deployment target < iOS10
     as_activity_scope_enter(as_activity_create("Prepare nodes for collection update", AS_ACTIVITY_CURRENT, OS_ACTIVITY_FLAG_DEFAULT), &preparationScope);
 
@@ -653,7 +657,15 @@ typedef void (^ASDataControllerSynchronizationBlock)();
     // Step 4: Inform the delegate on main thread
     [_mainSerialQueue performBlockOnMainThread:^{
       as_activity_scope_leave(&preparationScope);
+      if (_delegate.isDeallocating) {
+        as_log_debug(ASCollectionLog(), "datacontroller batchupdate during dealloc");
+        return;
+      }
       [_delegate dataController:self updateWithChangeSet:changeSet updates:^{
+        if (_delegate.isDeallocating) {
+          as_log_debug(ASCollectionLog(), "updateWithChangeSet during dealloc");
+          return;
+        }
         // Step 5: Deploy the new data as "completed"
         //
         // Note that since the backing collection view might be busy responding to user events (e.g scrolling),


mikezucc:
Here is one of the candidate blocks where `setObject:forKey:` killed the process
```
int ___51-[UICollectionView _viewAnimationsForCurrentUpdate]_block_invoke1586(int arg0, int arg1) {
    var_A0 = arg0;
    rdi = *(*(arg0 + 0x20) + *ivar_offset(_currentUpdate))->_newModel;
    r15 = *_objc_msgSend;
    r12 = [[rdi indexPathForItemAtGlobalIndex:arg1] retain];
    var_98 = r12;
    r14 = [[*(*(arg0 + 0x20) + *ivar_offset(_layout)) initialLayoutAttributesForAppearingItemAtIndexPath:r12] retain];
    rax = *(arg0 + 0x20);
    rax = rax->_currentUpdate;
    rcx = *ivar_offset(_newModel);
    rdx = r12;
    rbx = r15;
    r13 = [[*(rax + rcx) layoutAttributesForItemAtIndexPath:rdx] retain];
    r12 = r14;
    if (r14 != 0x0) goto loc_8b7e5a;

loc_8b7e1f:
    r14 = rbx;
    rbx = (r14)(r13, @selector(copy), rdx, rcx);
    [r12 release];
    rsi = @selector(setAlpha:);
    (r14)(rbx, rsi, rdx, rcx);
    if (rbx == 0x0) goto loc_8b80ea;

loc_8b7e54:
    r12 = rbx;
    rbx = r14;
    goto loc_8b7e5a;

loc_8b7e5a:
    r15 = var_A0;
    rsi = r12;
    rax = [&var_70 frame];
    goto loc_8b7e74;

loc_8b7e74:
    r14 = r13;
    if (r14 != 0x0) {
            rsi = r14;
            rax = [&var_90 frame];
    }
    else {
            var_80 = intrinsic_movaps(var_80, 0x0);
            var_90 = intrinsic_movaps(var_90, 0x0);
    }
    CGRectUnion(&var_48, rsi);
    if (CGRectIntersectsRect(&var_48, rsi) != 0x0) {
            if (([r12 isHidden] == 0x0) || ([r14 isHidden] == 0x0)) {
                    var_A8 = r12;
                    r12 = (rbx)(*(r15 + 0x20), @selector(_createPreparedCellForItemAtIndexPath:withLayoutAttributes:applyAttributes:), var_98, r12, 0x1);
                    r13 = [(rbx)(*(*(r15 + 0x20) + *ivar_offset(_currentUpdate))->_newModel, @selector(layoutAttributesForItemAtIndexPath:), var_98, *ivar_offset(_newModel)) retain];
                    [r14 release];
                    var_B0 = [(rbx)(*(*(r15 + 0x20) + *ivar_offset(_layout)), @selector(_animationForReusableView:toLayoutAttributes:type:), r12, r13, 0x2) retain];
                    (rbx)(@class(UICollectionViewAnimation), @selector(alloc));
                    intrinsic_movsd(xmm1, *0xafb288);
                    r14 = (rbx)(rax, @selector(initWithView:viewType:finalLayoutAttributes:startFraction:endFraction:animateFromCurrentPostion:deleteAfterAnimation:customAnimations:));
                    (rbx)(*(r15 + 0x28), @selector(addObject:), r14);
                    rdi = *(r15 + 0x30);
                    r15 = r12;
                    (rbx)(rdi, @selector(setObject:forKey:), r15, var_98);
                    r12 = var_A8;
                    [r14 release];
                    [var_B0 release];
                    [r15 release];
                    r14 = r13;
            }
    }
    [r14 release];
    [r12 release];
    rax = [var_98 release];
    return rax;
```

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/35c13a56b90c67f52a23f76bcfdc1899/raw/1f087b467dbdf5e0a6489dc24a7214c64359dce8/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/9c0593857d180067fd295a926782be04/raw/2a020874eec11b2826119f21d1644e2147d3bce8/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/abdf58cc58ebbcb9f461560cc2c2fd35/raw/edc414c7b8552817af0ecbfb1df8838b54db6de0/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/dd9878eb27ead432e2139e4197c51b00/raw/d64207a69b9786fecb91c5f621d13a8c6c1c83eb/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/13b0a16d8a52c19c277c2a7fa1d3d0f6/raw/9533b7448087b8a3ea253bd333d009d99371065a/log.txt)

mikezucc:
Ok finally passed 👍

Source/Details/ASDataController.h
```diff
@@ -115,6 +115,30 @@ AS_EXTERN NSString * const ASCollectionInvalidUpdateException;
...
+
+@protocol ASDeallocateStatusProtocol <NSObject>
+
+/**
+ * Communicate to the child whether or not its parent is scheduled for deallocation.
+ *
+ * @discussion This is in response to a hard to reproduce crash where
+ * `performBatchUpdates` is sent on a valid pointer, however the internal UIKit
+ * collection view components such as _UICollectionViewData later reference a nil shared
+ * indexPath store during their many block invokes for a colleciton view update.
```
mikezucc: typo * `collection`

Source/ASCollectionView.mm
```diff
@@ -333,6 +334,10 @@ - (void)dealloc
   ASPerformBackgroundDeallocation(&_rangeController);
 }
 
+- (BOOL)isDeallocating {
+  return _isDeallocating;
```
nguyenhuy: This method can be called on multiple thread (i.e the editing queue of `ASDataController`). We should make sure it's thread-safe.

Source/Details/ASDataController.h
```diff
@@ -115,6 +115,30 @@ AS_EXTERN NSString * const ASCollectionInvalidUpdateException;
...
+ *
+ * @discussion This is in response to a hard to reproduce crash where
+ * `performBatchUpdates` is sent on a valid pointer, however the internal UIKit
+ * collection view components such as _UICollectionViewData later reference a nil shared
+ * indexPath store during their many block invokes for a collection view update.
+ * This could be attributed the editingQueue still consuming its tasks
+ * while the collection view is being async deallocated. We want to early return
+ * at editing transation block invocation time since we can not destroy queued blocks;
+ */
+- (BOOL)isDeallocating;
```
nguyenhuy: Should we merge this protocol with `ASDataControllerDelegate`? Most of the time data controller only cares about whether its delegate is still around or not.

Source/Details/ASDataController.mm
```diff
@@ -629,6 +629,10 @@ - (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet
   Class<ASDataControllerLayoutDelegate> layoutDelegateClass = [self.layoutDelegate class];
   ++_editingTransactionGroupCount;
   dispatch_group_async(_editingTransactionGroup, _editingTransactionQueue, ^{
+    if (_deallocDelegate.isDeallocating) {
```
nguyenhuy: Check to make sure `_deallocDelegate` isn't `nil` because if it is, `isDeallocating` will return `NO`. We really don't want to proceed in that case.

nguyenhuy: Also, we should weakify & strongify `self`, and bail early if `self` is nil. It may not help with the crash, but it'll at least allow the data controller to be de-allocated much earlier (which is beneficial because it owns all of the cell nodes).

Source/Details/ASDataController.mm
```diff
@@ -653,7 +657,15 @@ - (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet
     // Step 4: Inform the delegate on main thread
     [_mainSerialQueue performBlockOnMainThread:^{
       as_activity_scope_leave(&preparationScope);
+      if (_deallocDelegate.isDeallocating) {
+        as_log_debug(ASCollectionLog(), "datacontroller batchupdate during dealloc");
+        return;
+      }
       [_delegate dataController:self updateWithChangeSet:changeSet updates:^{
```
nguyenhuy: Another question is, should we move the `isDeallocating` check to the implementation of the delegate? That means that `ASDataController` wouldn't mind about whether or not its delegate is being de-allocated, as long as the delegate is still around.

Source/Details/ASDataController.h
appleguy: Yeah, ideally let's either merge it or generalize it by moving out to one of our utility-ish directories. I think it would also be appropriate to reference Texture PR #1290 for details and skip the full description, particularly because we haven't confirmed this theory yet (although it is very well-substantiated).

Source/Details/ASDataController.mm
appleguy: Forgive me if I'm misunderstanding, but wouldn't _deallocDelegate itself be a sufficient indication of whether the object is deallocating? As a weak pointer, the contract is that the pointer should become nil before the dealloc method is entered. This is essential for objc4 to guarantee because a retain during dealloc is forbidden, and if a weak pointer were still available, a retain would be easily possible.

Source/ASCollectionView.mm
appleguy: If possible let's remove the isDeallocating method and instance variable itself, and just check the weak pointer for nil-ness.

Source/Details/ASDataController.mm
mikezucc: hm I assumed it would still be useful since `_delegate` as a weak pointer references the same object, and (given the root cause assumption in this PR) `_delegate` can perform a valid `objcmsg_send`

mikezucc:
@appleguy so my assumption here is that the weak pointer is still around as the `_delegate` reference can still perform an `objc_msgsend` of `performBatchUpdate`. The ultimate strategy here is to prevent UICollectionView data source reads from happening during and after `dealloc` which can still happen if the collection view is destroyed during asmainserialqueue before the main dispatch queue can consume the batch update blocks. is that right?

Adlai-Holler:
Hey @mikezucc, a few thoughts:

- The collection view is deallocated on the main thread, and these callbacks are on the main thread so the two can't happen at the same time.
- Message sends to weak variables actually do a retain under the hood, and call the method on the temporary strong variable that's created: `id strongObj = [weakObj tryRetain]; [strongObj doMethod]; [strongObj release];` so if the collection view had been marked for deallocation, strongObj will be nil and the method will nop.
- The only safe way to check if an object is still retainable from the outside is to try and retain a weak pointer. Flags like these don't work because by the time we enter `-dealloc` it's too late.
- If we look at some stack traces of the crashes, it should be easy to determine if the collection view is in `-dealloc`, since everything must happen on the main thread. Do you have some?

Another circumstance that could cause symptoms like these is, if the data source had changed (or died) in the meantime. Say, if the real data source is a view controller that deallocates while the collection view survives (someone else is retaining), and it set the collection node data source to nil, which updates the uicollectionview's data source, it may now point to an empty data source out from under us. So possibly we should re-check _that_ (that the data source/data did not change in the meantime) when we regain the main thread.

mikezucc:
screen shot 2019-01-30 at 3 17 02 pm
screen shot 2019-01-30 at 3 20 49 pm
screen shot 2019-01-30 at 3 21 04 pm


The call stacks after they get into UIKit begin to differ. They could be from many different root causes. I wonder what sort of situations could lead to the crash logs that show a nil shared global indexPaths trying to be accessed by a `_UICollectionViewUpdate` which is an ivar of the `_UICollectionViewData` which is an ivar of `_UICollectionView`.

Also I think it is important to note that the dealloc queue is batch consumed as the last step in the default run loop before returning control back to GCD to consume the main queue on the main thread. Is it possible for CRUD blocks for the collection view or its data source to be scheduled on the main queue (but no executed) **before** something on the dealloc queue (ASMainSerialQueue) yet be executed after? see the diagram in my response to Garrett in the comments above

PR #1291, opened 2018-12-19
Reduce startup time.

dmaclach:
Get rid of file scope initialization that causes a premain startup hit.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/7e47e78b32bf6ff8221597375276ea2d/raw/52e55dbd389263cba1e970b8e7b720ec923291d4/log.txt)

PR #1296, opened 2018-12-21
Add some snapshot tests for ASTextNode2 truncation modes.

wiseoldduck:
Should have added these sooner, looks like there is a regression in NSLineBreakModeMiddle :-(


Tests/ASTextNode2SnapshotTests.mm
```diff
@@ -45,13 +120,86 @@ - (void)testTextContainerInset_ASTextNode2
...
+  [container addSubview:uiLabelLabel];
+  [container addSubview:description];
+
+  NSArray<LineBreakConfig *> *c = [LineBreakConfig configs];
+  for (LineBreakConfig *config in c) {
+    reference.lineBreakMode = textNode.truncationMode = config.lineBreakMode;
+    reference.numberOfLines = textNode.maximumNumberOfLines = config.numberOfLines;
+    description.text = config.description;
+    [container setNeedsLayout];
+    NSString *identifier = [NSString stringWithFormat:@"%lu_%lu", (unsigned long)config.lineBreakMode, (unsigned long)config.numberOfLines];
```
nguyenhuy: Instead of using the line break mode int value, maybe it’d be easier to understand if we use a string representation of the mode here?

wiseoldduck: okay

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5fbffb1d8c1b4cd4d041fac65a1284ec/raw/e0ac4af8e656d104d61d703426a00bbb18b9a872/log.txt)

PR #1302, opened 2019-01-05
Move ASSectionController and ASSupplementaryNodeSource method to be optional

maicki:
After #1273 was merged it's possible to implement either `nodeBlockForItemAtIndex:` or `nodeForItemAtIndex:` in the `ASSectionController` case and `nodeBlockForSupplementaryElementOfKind:atIndex:` or `nodeForSupplementaryElementOfKind:atIndex` in the `ASSupplementaryNodeSource` case.

As it's an or case now instead of having the required we move them to optional and throw an assertion if no one of the two is implemented.

### DIFF
--- a/Source/ASSectionController.h
+++ b/Source/ASSectionController.h
@@ -16,12 +16,13 @@ NS_ASSUME_NONNULL_BEGIN
 @class ASBatchContext;
 
 /**
- * A protocol that your section controllers should conform to,
- * in order to be used with AsyncDisplayKit.
+ * A protocol that your section controllers should conform to, in order to be used with Texture.
  *
  * @note Your supplementary view source should conform to @c ASSupplementaryNodeSource.
  */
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
I'm surprised the `nodeForItemAtIndex:` method exists! Can we remove it? It's only on ASCollectionView for legacy reasons so my original intention was not to have it for section controllers.

maicki:
@Adlai-Holler Yeah I know that this method should not exist at all as they are deprecated, removing them is a larger undertaking though.

buithuyen:
Hi you guys.
How can I got this pull request from cocoapod ?
I have tried to got master `pod 'Texture', :git => 'https://github.com/TextureGroup/Texture.git'`
But I have many dependency, so it was not successful.
Do you have any plan for new release.

PR #1303, opened 2019-01-06
Make ASCollectionElement Public

JonathanDowning:
I'm having a problem with using AsyncDisplayKit with Carthage.

I've previously used Cocoapods with my project, but whilst changing to Carthage, I noticed that ASCollectionElement.h is private.

I presume this is a mistake as ASCollectionElement is referenced in other public headers in this project.


dskuza:
I noticed this PR was merged into master, but is still missing from the `2.8.1` release (which is dated post-merge). Is there insight into when this will be available? I'd like to stick to semver rather than "master" in my Cartfile.

PR #1306, opened 2019-01-07
Fix deprecated implementations warning

maicki:
Seems like Xcode 10.1 introduces some new warnings. Either we should prevent them via Xcode settings or like in this PR guard them via a compiler flag.

### DIFF
--- a/Source/Details/ASThread.h
+++ b/Source/Details/ASThread.h
@@ -197,11 +197,11 @@ namespace ASDN {
     }
 
     void AssertHeld() {
-      ASDisplayNodeCAssert(_owner == std::this_thread::get_id(), @"Thread should hold lock");
+//      ASDisplayNodeCAssert(_owner == std::this_thread::get_id(), @"Thread should hold lock");
     }
     
     void AssertNotHeld() {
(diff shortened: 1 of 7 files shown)


Source/Details/ASThread.h
```diff
@@ -197,11 +197,11 @@ namespace ASDN {
     }
 
     void AssertHeld() {
-      ASDisplayNodeCAssert(_owner == std::this_thread::get_id(), @"Thread should hold lock");
+//      ASDisplayNodeCAssert(_owner == std::this_thread::get_id(), @"Thread should hold lock");
     }
     
     void AssertNotHeld() {
-      ASDisplayNodeCAssert(_owner != std::this_thread::get_id(), @"Thread should not hold lock");
+//      ASDisplayNodeCAssert(_owner != std::this_thread::get_id(), @"Thread should not hold lock");
```
Adlai-Holler: Mistake?

maicki: Most definitely :)

maicki:
@Adlai-Holler Want to give that another look please. Thank!

PR #1310, opened 2019-01-10
[ASCollectionView] Supplementary nodes should not enter ASHierarchyStateRangeManaged.

appleguy:
As of today, ASRangeController doesn't manage the interfaceState for supplementary nodes. This bug was introduced with a refactoring of this code a couple months ago.

With this change, the supplementary nodes will automatically enter each interfaceState bit when they are added to a view hierarchy.


appleguy:
Although I had done the incorrect change internally, it looks like it just got published a month ago: https://github.com/TextureGroup/Texture/pull/1282

PR #1311, opened 2019-01-11
Add experiments to skip waiting for updates of collection and table views

nguyenhuy:
- `exp_skip_a11y_wait`: An experiment that skips a wait when accessibility elements of a collection/table view is collected. The wait was introduced in #1217. We suspect this causes some perf regressions and need to confirm via an experiment.
- `exp_skip_default_cell_layout_mode`: An experiment that skips a new default behavior of `ASCellLayoutModeNone`, introduced in #1273. The new behavior blocks main thread when 1) the collection/table view reloads data (including the initial load and any subsequent ones), 2) there is a small number of async cells to be laid out, or 3) the collection/table view hasn't had enough content to fill the visible viewport. While I agree that this new behavior may help with display latency when it's more important than FPS (i.e during app startup), I'm not sure if this is the right approach and, even if it is, we need to measure its perf implications within our app. My main concern is that blocking the main thread during app startup prevents the app from performing other equally important tasks that are intentionally scheduled during that time. The new behavior also affects apps that optimize the first page's size to fill the viewport, but doesn't necessarily fill it entirely due to various reasons. In such case, it's actually quite bad to block main thread for the second batch update, because users may (and are welcome to) interact with the content of the first page.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -2465,7 +2465,9 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 
 - (NSArray *)accessibilityElements
 {
-  [self waitUntilAllUpdatesAreCommitted];
+  if (!ASActivateExperimentalFeature(ASExperimentalSkipAccessibilityWait)) {
+    [self waitUntilAllUpdatesAreCommitted];
+  }
   return [super accessibilityElements];
 }
 
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -1869,14 +1869,20 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     return NO;
   }
   // The heuristics below apply to the ASCellLayoutModeNone.
-  // If we have very few ASCellNodes (besides UIKit passthrough ones), match UIKit by blocking.
-  if (changeSet.countForAsyncLayout < 2) {
-    return YES;
-  }
-  CGSize contentSize = self.contentSize;
-  CGSize boundsSize = self.bounds.size;
-  if (contentSize.height <= boundsSize.height && contentSize.width <= boundsSize.width) {
-    return YES;
+  if (!ASActivateExperimentalFeature(ASExperimentalSkipDefaultCellLayoutMode)) {
+    // Reload data is expensive, don't block main while doing so.
+    if (changeSet.includesReloadData) {
+      return NO;
+    }
+    // If we have very few ASCellNodes (besides UIKit passthrough ones), match UIKit by blocking.
+    if (changeSet.countForAsyncLayout < 2) {
+      return YES;
+    }
+    CGSize contentSize = self.contentSize;
+    CGSize boundsSize = self.bounds.size;
+    if (contentSize.height <= boundsSize.height && contentSize.width <= boundsSize.width) {
+      return YES;
+    }
   }
   return NO;
 }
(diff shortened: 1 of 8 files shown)


Source/ASCollectionView.mm
```diff
@@ -1869,14 +1869,20 @@ - (BOOL)dataController:(ASDataController *)dataController shouldSynchronouslyPro
...
-  if (changeSet.countForAsyncLayout < 2) {
-    return YES;
-  }
-  CGSize contentSize = self.contentSize;
-  CGSize boundsSize = self.bounds.size;
-  if (contentSize.height <= boundsSize.height && contentSize.width <= boundsSize.width) {
-    return YES;
+  if (!ASActivateExperimentalFeature(ASExperimentalSkipDefaultCellLayoutMode)) {
+    // Reload data is expensive, don't block main while doing so.
+    if (changeSet.includesReloadData) {
```
nguyenhuy: Added this check because `countForAsyncLayout` doesn't account for reload data.

ernestmama:
LGTM

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/345a5c17b37e046334ca428642823f8c/raw/7a3020b32a84c587645e072436da902f7e8b19fb/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ed8ac94db2cb3f293e8fea19b1d1abcf/raw/2a8519406d3c16d56f94622069229c91499ddb22/log.txt)

Tests/ASCollectionViewTests.mm
```diff
@@ -1071,19 +1071,15 @@ - (void)testInitialRangeBoundsWithCellLayoutMode:(ASCellLayoutMode)cellLayoutMod
   // Trigger the initial reload to start 
   [window layoutIfNeeded];
 
-  // Test the APIs that monitor ASCollectionNode update handling if collection node should
-  // layout asynchronously
-  if (![cn.view dataController:nil shouldSynchronouslyProcessChangeSet:nil]) {
```
nguyenhuy: @maicki Heads up that I think we need to remove this line.

I understand why we need to call `waitUntilAllUpdatesAreProcessed` if the collection view doesn't synchronously process the change set: to make sure the initial load finishes before we can proceed. However, in order to reliably do so, we need to pass in a valid change set that contains a `reloadData` change, instead of `nil`.

We can either do that, or always call `waitUntilAllUpdatesAreProcessed` like what I'm proposing here.
- If the change set is processed asynchronously, calling `waitUntilAllUpdatesAreProcessed` will be required here.
- If it's processed synchronously, calling `waitUntilAllUpdatesAreProcessed` here will be super fast and won't do any harm.

nguyenhuy:
@Adlai-Holler Thanks for the review. Reverting back would be ideal actually. I was assuming that you were aware of the new default behavior.

Adlai-Holler:
Let's do this – would you be willing to add a new mode named e.g. `ASCellLayoutModeSyncForSmallContent` and have the experiment _enable_ that mode as the default? Then ASCellLayoutModeNone will stay on and mean the previous default behavior?

appleguy:
I'm hoping to look at this in more detail today, however, I may not have time and don't want to block this from moving forward. 

Running the experiment sounds good for sure. I'm a little more worried about changing the default again unless we are carefully coordinated because it would be dangerous for one of our big clients to accidentally ship that change. :) 

If Huy is confident testing this flag as an option, than we can go from there?

nguyenhuy:
@appleguy I think it's better for the new behavior to be opt-in, at least until all of us are confident that it's beneficial and everyone should use it.
@Adlai-Holler I updated the diff with a new `ASCellLayoutModeSyncForSmallContent` mode. Please make sure to enable it on your end. Thank you.

PR #1318, opened 2019-01-16
Correct attributes lookup for supplementary elements.

wiseoldduck:


Source/ASCollectionView.mm
```diff
@@ -1995,7 +1995,13 @@ - (BOOL)dataController:(ASDataController *)dataController presentedSizeForElemen
     ASDisplayNodeFailAssert(@"Data controller should not ask for presented size for element that is not presented.");
     return YES;
```
appleguy: I think the condition above, indexPathForNode:, may also need to be updated to check e.g. indexPathForSupplementaryElement: ofKind: -- or maybe we can implement (or use if existing) indexPathForElement:

wiseoldduck: Yeah `indexPathForNode:` does appear to lead to `indexPathForElement:`, which itself just looks up in `_elementToIndexPathMap`, which I don't know what is but isn't an OS call so maybe it's just fine?

Adlai-Holler: Yep `indexPathForNode:` is fine here, it works for supplementaries and for items.

PR #1319, opened 2019-01-16
Dev docs for threading in Texture

mikezucc:
Looking for more things to add!

outline:
Threading
  - At a glance
  - Run Loop, Threads, and Queues
     - Dispatch Queue Playgrounds
  - ASMainSerialQueue
  - ASRunLoopQueue
  - Locks and Safety
  - Thread Contention
  - Other Threading Conventions
  - Threading by example


docs/_docs/development/threading.md
```diff
@@ -4,4 +4,357 @@ layout: docs
...
+        uint64_t elapsed = end - start;
+        uint64_t nanosec = elapsed * s_timebase_info.numer / s_timebase_info.denom;
+        printf("%llu ns block %i\n", nanosec / kOneMillion, i);
+        usleep(10000);
+    });
+}
+```
+Looking at the logs, we can see how GCD reaches a maximum background thread count, varied by CPU. This example was run on the 8 core Macbook Pro
+```
+1 ns block 62
```
mikezucc: ms* not ns

docs/_docs/development/threading.md
```diff
@@ -4,4 +4,357 @@ layout: docs
...
+1 ns block 63
+10 ns block 65
+10 ns block 66
+10 ns block 64
+```
+
+## ASMainSerialQueue
+
+The `ASMainSerialQueue` ensures that work can be performed serially on the main thread without being interrupted. The key difference between this and purely using `dispatch_async(dispatch_get_main_queue, block)` is that the main thread can be interrupted between execution of blocks in its queue, where as this interface will execute everything possible in its queue on the main thread before returning control back to the OS.
```
garrettmoon: Are you saying that OS will not context switch away from the main thread while a block is being run by ASMainSerialQueue? I'm not sure that's the case.

If I remember correctly, the purpose of ASMainSerialQueue is that a block submitted off the main thread to the main thread will be run before another block submitted on the main thread if it was submitted after. If we just used `ASPerformBlockOnMainThread` a block submitted on the main thread *after* a block submitted off the main thread would be run before the block submitted (temporally) before.

mikezucc: yes thats what i was trying to get at with my explanation. here i made a sweet google drawing to hopefully communicate my understanding of this class
asmainserialqueue

mikezucc: also I'll change that from `returning control back to the OS` to `returning control back to GCD scheduler`

mikezucc: In a ridiculous, scenario, is it possible for C to never execute if background threads are hammering ASMainSerialQueue with new blocks at rate that is equal to or faster than the rate at which main thread is consuming those blocks?

mikezucc:
* add mention to use the `ASDN` namespace mutexes

garrettmoon:
Another thought: It may be worth separating this out a bit? Possibly having the non-texture specific stuff in a separate page we link at the top of the texture specific stuff?

nguyenhuy:
Thanks for working on this, @mikezucc. Let's land it now and follow up if need to.

PR #1324, opened 2019-01-19
Deadlock fixes

wiseoldduck:
1) lock to root for tree
2) lock self to change parent (& consequently root)
3) Implement ASLocking (tryLock) on ASNodeController
4) add lockPair to try-lock node & controller together
5) lock controllers if they exist in lockToRoot...


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d385a982806a70342bc21e034d47c9de/raw/5458cacf5d93452dc2d38e799d8ce0cccaed6f90/log.txt)

PR #1326, opened 2019-01-22
[Text Nodes] Maintain isAccessibilityElement setting on text nodes when updating text

smeis:
I'd like to propose a change in the way `isAccessibilityElement` is updated when setting the `attributedText` property on `ASTextNode` and `ASTextNode2`. 

Currently this always happen when the `attributedText` property is updated. This makes sense in most cases. But when you set a string and you configure the text node's `isAccessibilityElement` property to `NO`, it will change back to yes if you update the string. Meaning you always need to update `isAccessibilityElement` after updating the text. 

I think that in the case where the `attributedText` property has been set, and the `isAccessibilityElement` has been explicitly set to `NO`, and you update the text (not set the property to `nil`) it makes sense to respect this setting and making the user of the text node responsible for changing it if that's necessary.

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -462,12 +462,16 @@ static NSArray *DefaultLinkAttributeNames() {
     attributedText = [[NSAttributedString alloc] initWithString:@"" attributes:nil];
   }
 
+  NSAttributedString *oldAttributedText = nil;
+  
   {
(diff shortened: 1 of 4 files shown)


smeis:
Hi @nguyenhuy @maicki , would it be possible to get a review for this?

wsdwsd0829:
LGTM

PR #1327, opened 2019-01-24
Fix a11y update when insert view

wsdwsd0829:


wsdwsd0829:
As I tested the insertView will be called instead of addSubview below even for the view hierarchy is created first time. Another case is when view tree is updated by removing and inserting a new image node for example,  the new image node will not have a11y info without this change.

Adlai-Holler:
That makes sense but if we put the code in `_addSubnode:` (not `addSubview:`) on ASDisplayNode, I think we can cover layer backed nodes and wrapper nodes at the same time. Something along the lines of:

```objc
- (void)_addSubnode:(ASDisplayNode *)node {
  // other code
  if (node->_layer) {
    // Find nearest view above (self or parent).
    nearestView.accessibilityElements = nil;`
  }
}
```

PR #1330, opened 2019-01-29
Update layout2-layoutspec-types.md

knowsudhanshu:
`justifyContent` definition updated.

### DIFF
--- a/docs/_docs/layout2-layoutspec-types.md
+++ b/docs/_docs/layout2-layoutspec-types.md
@@ -84,7 +84,7 @@ they will be resolved again, causing justifyContent and alignItems to be updated
 - `verticalAlignment`. Specifies how children are aligned vertically. Depends on the stack direction, setting the alignment causes either
  justifyContent or alignItems to be updated. The alignment will remain valid after future direction changes.
  Thus, it is preferred to those properties.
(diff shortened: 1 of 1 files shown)


docs/_docs/layout2-layoutspec-types.md
```diff
@@ -84,7 +84,7 @@ they will be resolved again, causing justifyContent and alignItems to be updated
 - `verticalAlignment`. Specifies how children are aligned vertically. Depends on the stack direction, setting the alignment causes either
  justifyContent or alignItems to be updated. The alignment will remain valid after future direction changes.
  Thus, it is preferred to those properties.
-- `justifyContent`. The amount of space between each child.
+- `justifyContent`. It defines the alignment along the main axis.
```
nguyenhuy: While both descriptions are valid, I agree that the new one is clearer because there are cases ("start", "end" and "center") in which the space between each child is zero and yet the alignments are different. Plus, the new description is similar to the one in [CSS Flexbox Spec](https://www.w3.org/TR/css-flexbox-1/#justify-content-property).

Nit: How about "Alignment of children along main axis"? It's consistent with the other descriptions in this page.

nguyenhuy:
Cool, thank you, @knowsudhanshu.

PR #1344, opened 2019-02-20
Add forwarding of UIAccessibilityAction methods

smeis:
I was working on accessibility in our app and noticed these weren't implemented yet. There's no checking if the node implements the method because `NSObject` provides default implementations.

### DIFF
--- a/Tests/ASDisplayViewAccessibilityTests.mm
+++ b/Tests/ASDisplayViewAccessibilityTests.mm
@@ -18,6 +18,49 @@
 #import <AsyncDisplayKit/ASConfiguration.h>
 #import <AsyncDisplayKit/ASConfigurationInternal.h>
 
+@interface ASAccessiblityForwardingTestDisplayNode : ASDisplayNode
+
+@property (nonatomic, readonly) BOOL didReceiveAccessibilityActivate;
+@property (nonatomic, readonly) BOOL didReceiveAccessibilityIncrement;
+@property (nonatomic, readonly) BOOL didReceiveAccessibilityDecrement;
+@property (nonatomic, readonly) BOOL didReceiveAccessibilityScroll;
+@property (nonatomic, readonly) BOOL didReceiveAccessibilityPerformEscape;
+@property (nonatomic, readonly) BOOL didReceiveAccessibilityPerformMagicTap;
+
+@end
+
+@implementation ASAccessiblityForwardingTestDisplayNode
+
+- (BOOL)accessibilityActivate {
+  _didReceiveAccessibilityActivate = YES;
+  return YES;
+}
+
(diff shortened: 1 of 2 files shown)


smeis:
Hi @maicki @appleguy , I can't add reviewers but you've worked a lot on these components, could you take a look at it?

Tests/ASDisplayViewAccessibilityTests.mm
```diff
@@ -269,4 +312,25 @@ - (void)testAccessibilityUpdatesWithElementsChanges
...
+  [view accessibilityIncrement];
+  XCTAssertTrue(node.didReceiveAccessibilityIncrement);
+  [view accessibilityDecrement];
+  XCTAssertTrue(node.didReceiveAccessibilityDecrement);
+  [view accessibilityScroll:UIAccessibilityScrollDirectionDown];
+  XCTAssertTrue(node.didReceiveAccessibilityScroll);
+  [view accessibilityPerformEscape];
+  XCTAssertTrue(node.didReceiveAccessibilityPerformEscape);
+  [view accessibilityPerformMagicTap];
+  XCTAssertTrue(node.didReceiveAccessibilityPerformMagicTap);
```
Adlai-Holler:
```suggestion
id mockNode = OCMPartialMock(node);
OCMExpect([node accessibilityActivate]);
[view accessibilityActivate];
OCMExpect([node accessibilityIncrement]);
[view accessibilityIncrement];
// …
OCMVerifyAll(mockNode);
```

smeis: @Adlai-Holler fixed!

Adlai-Holler:
Thanks for the fixup @smeis , hope we see more of you!!

fruitcoder:
Is there anything specific to watch out for? I tried subclassing an ASScrollNode and the overridden `accessibiliyActivate` isn't called

maicki:
@fruitcoder As `ASScrollNode` uses `UIScrollView` and not `_ASDisplayView` as its view this methods will not be passed through to the node. We would have to add them to the `ASScrollView` class.

fruitcoder:
I think I tried that as well but it didn't work since I can't provide a `ASScrollView` subclass that overrides the `accessibilityActivate` so I just added an a11y proxy node as the only scroll node's child when accessibility is enabled

maicki:
You should be able to do this by subclassing `ASScrollView` and `ASScrollNode` and instead of creating the `ASScrollView` as view of the `ASScrollNode` within the initializer you are initialize your custom `ASScrollView`.  I'm not sure right now which one takes presedence but you also should be able to archive the same by overriding the `+[ASDisplayNode viewClass]` method

PR #1345, opened 2019-02-21
Add an experimental flag to use native dispatch_apply

Adlai-Holler:
This has shown performance wins against our current core_count*2 approach in some profiling, possibly due to reduced context switching.


Adlai-Holler:
Great – the wins we're seeing on old devices are significant so give it a go!

PR #1349, opened 2019-02-23
Experiment with different strategies for image downloader priority

nguyenhuy:
Right now when an image node enters preload state, we kick off an image request with the default priority. Then when it enters display state, we change the priority to "imminent" which is mapped to the default priority as well. This means that requests from preload and display nodes have the same priority and are put to the same pool. The right behavior would be that preload requests should have a lower priority from the beginning.

Another problem is that, due to the execution order of -didEnter(Preload|Display|Visible)State calls, a node may kick off a low priority request when it enters preload state even though it knows that it's also visible. By the time -didEnterVisibleState is called, the low priority request may have already been consumed and the download/data task won't pick up the new higher priority, or some work needs to be done to move it to another queue. A better behavior would be to always use the current interface state to determine the priority. This means that visible nodes will kick off high priority requests as soon as -didEnterPreloadState is called.

The last (and smaller) issue is that a node marks its request as preload/low priority as soon as it exits visible state. I'd argue that this is too agressive. It may be reasonble for nodes in the trailing direction. Even so, we already handle this case by (almost always) have smaller trailing buffers. So this diff makes sure that nodes that exited visible state will have imminent/default priority if they remain in the display range.

All of these new behaviors are wrapped in an experiment and will be tested carefully before being rolled out.

### DIFF
--- a/Source/ASMultiplexImageNode.mm
+++ b/Source/ASMultiplexImageNode.mm
@@ -96,7 +96,6 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   
   //set on init only
   BOOL _cacheSupportsClearing;
-  BOOL _isInImageDownloaderPriorityExperiment;
 }
 
 //! @abstract Read-write redeclaration of property declared in ASMultiplexImageNode.h.
@@ -184,8 +183,6 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   
   _shouldRenderProgressImages = YES;
 
-  _isInImageDownloaderPriorityExperiment = ASActivateExperimentalFeature(ASExperimentalImageDownloaderPriority);
-  
   self.shouldBypassEnsureDisplay = YES;
 
   return self;
@@ -302,7 +299,7 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
 - (void)didExitDisplayState
 {
   [super didExitDisplayState];
-  if (_isInImageDownloaderPriorityExperiment) {
+  if (ASActivateExperimentalFeature(ASExperimentalImageDownloaderPriority)) {
     [self _updatePriorityOnDownloaderIfNeededWithDefaultPriority:ASImageDownloaderPriorityPreload];
   }
 }
@@ -488,20 +485,19 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
 - (void)_updatePriorityOnDownloaderIfNeededWithDefaultPriority:(ASImageDownloaderPriority)defaultPriority
 {
   ASAssertUnlocked(_downloadIdentifierLock);
-  if (_downloaderFlags.downloaderImplementsSetPriority) {
-    id downloadIdentifier;
-    {
-      ASDN::MutexLocker l(_downloadIdentifierLock);
-      downloadIdentifier = _downloadIdentifier;
-    }
 
-    if (downloadIdentifier != nil) {
+  if (_downloaderFlags.downloaderImplementsSetPriority) {
+    // Read our interface state before locking so that we don't lock super while holding our lock.
+    ASInterfaceState interfaceState = self.interfaceState;
+    ASDN::MutexLocker l(_downloadIdentifierLock);
+
+    if (_downloadIdentifier != nil) {
       ASImageDownloaderPriority priority = defaultPriority;
-      if (_isInImageDownloaderPriorityExperiment) {
-        priority = ASImageDownloaderPriorityWithInterfaceState(self.interfaceState);
+      if (ASActivateExperimentalFeature(ASExperimentalImageDownloaderPriority)) {
+        priority = ASImageDownloaderPriorityWithInterfaceState(interfaceState);
       }
 
-      [_downloader setPriority:priority withDownloadIdentifier:downloadIdentifier];
+      [_downloader setPriority:priority withDownloadIdentifier:_downloadIdentifier];
     }
   }
 }
@@ -875,7 +871,7 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
 
     id downloadIdentifier;
     if (strongSelf->_downloaderFlags.downloaderImplementsDownloadWithPriority
-        && strongSelf->_isInImageDownloaderPriorityExperiment) {
+        && ASActivateExperimentalFeature(ASExperimentalImageDownloaderPriority)) {
 
       /*
        Decide a priority based on the current interface state of this node.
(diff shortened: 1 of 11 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/711364327b00d733c9161f71edfc3ef8/raw/b7dc67578c178c4ba0fcd8543b240a60dd20d348/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/05ea44cc48628c9cf6306a1bb4db5c25/raw/8760b2eae131736676da2d4db22c5d092a7edc39/log.txt)

Source/ASNetworkImageNode.mm
```diff
@@ -98,12 +100,14 @@ - (instancetype)initWithCache:(id<ASImageCacheProtocol>)cache downloader:(id<ASI
...
   _downloaderFlags.downloaderImplementsAnimatedImage = [downloader respondsToSelector:@selector(animatedImageWithData:)];
   _downloaderFlags.downloaderImplementsCancelWithResume = [downloader respondsToSelector:@selector(cancelImageDownloadWithResumePossibilityForIdentifier:)];
+  _downloaderFlags.downloaderImplementsDownloadWithPriority = [downloader respondsToSelector:@selector(downloadImageWithURL:priority:callbackQueue:downloadProgress:completion:)];
 
   _cacheFlags.cacheSupportsClearing = [cache respondsToSelector:@selector(clearFetchedImageFromCacheWithURL:)];
   _cacheFlags.cacheSupportsSynchronousFetch = [cache respondsToSelector:@selector(synchronouslyFetchedCachedImageWithURL:)];
   
   _shouldCacheImage = YES;
   _shouldRenderProgressImages = YES;
+  _isInImageDownloaderPriorityExperiment = ASActivateExperimentalFeature(ASExperimentalImageDownloaderPriority);
```
Adlai-Holler: Let's not cache this 😱  – ASActivateExperimentalFeature is fast enough to be called repeatedly so let's leave it as the source of truth.

Source/ASMultiplexImageNode.mm
```diff
@@ -835,30 +846,67 @@ - (void)_downloadImageWithIdentifier:(id)imageIdentifier URL:(NSURL *)imageURL c
...
+      /*
+       Kick off a download with default priority.
+       The actual "default" value is decided by the downloader.
+       ASBasicImageDownloader and ASPINRemoteImageDownloader both use ASImageDownloaderPriorityImminent
+       which is mapped to NSURLSessionTaskPriorityDefault.
+
+       This means that preload and display nodes use the same priority
+       and their requests are put into the same pool.
+       */
+      downloadIdentifier = [strongSelf->_downloader downloadImageWithURL:imageURL
```
wsdwsd0829: can we call this we a priority also so that it do not need to be called differently in if-else?

Source/ASMultiplexImageNode.mm
```diff
@@ -271,49 +279,34 @@ - (BOOL)placeholderShouldPersist
...
   [super didExitVisibleState];
-  
-  if (_downloaderImplementsSetPriority) {
-    ASDN::MutexLocker l(_downloadIdentifierLock);
-    if (_downloadIdentifier != nil) {
-      [_downloader setPriority:ASImageDownloaderPriorityPreload withDownloadIdentifier:_downloadIdentifier];
-    }
-  }
-  
+  [self _updatePriorityOnDownloaderIfNeededWithDefaultPriority:ASImageDownloaderPriorityPreload];
```
wsdwsd0829: Just reminder that in thrash case (without coalescing) a view may exit and enter immediately. It seems ok in this case though.

Source/ASNetworkImageNode.mm
```diff
@@ -583,24 +596,55 @@ - (void)_downloadImageWithCompletion:(void (^)(id <ASImageContainerProtocol> ima
...
+      /*
+        Kick off a download with default priority.
+        The actual "default" value is decided by the downloader.
+        ASBasicImageDownloader and ASPINRemoteImageDownloader both use ASImageDownloaderPriorityImminent
+        which is mapped to NSURLSessionTaskPriorityDefault.
+
+       This means that preload and display nodes use the same priority
+        and their requests are put into the same pool.
+      */
+      downloadIdentifier = [self->_downloader downloadImageWithURL:url
```
wsdwsd0829: same as above, if we can pass ASImageDownloaderPriorityImminent to priority, we can extract the call out of if else.

wsdwsd0829:
Thanks for the pr, it makes sense,  curious how will we measure the effeteness of this experiment. Any metrics? I wonder if we can do the measure somehow from framework level instead of app level.

wiseoldduck:
LGTM as well. A couple of these issues (not setting a real priority until interface state changes, and the side effects of interfaceStateDelegate call order) seem like just-bugs that should not even need an experiment, but I understand the cautious approach.

Source/ASMultiplexImageNode.mm
```diff
@@ -493,19 +485,38 @@ - (UIImage *)_bestImmediatelyAvailableImageFromDataSource:(id *)imageIdentifierO
...
- @note: This should be called without _downloadIdentifierLock held. We will lock
- super to read our interface state and it's best to avoid acquiring both locks.
- */
+- (void)_updatePriorityOnDownloaderIfNeededWithDefaultPriority:(ASImageDownloaderPriority)defaultPriority
+{
+  ASAssertUnlocked(_downloadIdentifierLock);
+  if (_downloaderFlags.downloaderImplementsSetPriority) {
+    id downloadIdentifier;
+    {
+      ASDN::MutexLocker l(_downloadIdentifierLock);
```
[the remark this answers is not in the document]
appleguy: @nguyenhuy is there a codepath where the downloadIdentifier could become invalid or set by a different thread first, then clobbered by this thread when resumed? Seems like an unlikely scenario, but it may need to be a locked transaction for correctness.

Source/ASMultiplexImageNode.mm
nguyenhuy: Good call. I think the right fix is coalescing in such case to avoid jumping between interface states unnecessarily.

Source/ASMultiplexImageNode.mm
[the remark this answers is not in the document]
nguyenhuy: You two are right. This is a change in behavior. At first it seemed to me that we should retain the lock as short as possible, but on a second thought, it's much safer to hold the lock for the entire transaction. Gonna update the diff.

Source/ASMultiplexImageNode.mm
nguyenhuy: I'd like to let the concrete downloader to decide what the default priority is. Right now both `ASBasicImageDownloader` and `ASPINRemoteImageDownloader` have the same default value, but other custom ones may have a different one.

Source/ASNetworkImageNode.mm
nguyenhuy: Ah yeah, my bad. I didn't look into the implementation of `ASActivateExperimentalFeature`.

I assumed that it calls out to the configuration delegate every time. Our experiment manager is known to have measurable perf impacts if it's repeatedly called many times within a frame.

But since `ASActivateExperimentalFeature` calls its delegate only a new experiment is triggered, we don't need to cache this. Gonna update the diff.

nguyenhuy:
Alright, I’m landing this diff. Thanks everyone!

PR #1350, opened 2019-02-26
Optimize ASCATransactionQueue

Adlai-Holler:
This queue is extremely busy, since every single node enters it on every single interface state change on the main thread.

Avoid waking up the run loop for every single node.
Avoid a ton of NSPointerArray overhead that we don't need.
Avoid retain/release traffic on the singleton by using an inline function. I confirmed that in release mode, the static __strong is correctly inlined and no ARC traffic is incurred.

I will post some traces in a little while.

### DIFF
--- a/Source/ASRunLoopQueue.mm
+++ b/Source/ASRunLoopQueue.mm
@@ -301,9 +301,10 @@ static void runLoopSourceCallback(void *info) {
 
   if (!foundObject) {
     [_internalQueue addPointer:(__bridge void *)object];
-
-    CFRunLoopSourceSignal(_runLoopSource);
-    CFRunLoopWakeUp(_runLoop);
+    if (_internalQueue.count == 1) {
+      CFRunLoopSourceSignal(_runLoopSource);
+      CFRunLoopWakeUp(_runLoop);
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
Want approval from @wsdwsd0829 before landing.

Adlai-Holler:
Rebased to kick the ci

Source/ASRunLoopQueue.mm
```diff
@@ -342,22 +352,15 @@ @implementation ASCATransactionQueue
...
-  return sharedQueue;
-}
-
 - (instancetype)init
 {
   if (self = [super init]) {
-    _runLoop = CFRunLoopGetMain();
-    NSPointerFunctionsOptions options = NSPointerFunctionsStrongMemory;
-    _internalQueue = [[NSPointerArray alloc] initWithOptions:options];
+    _internalQueueHashSet = CFSetCreateMutable(NULL, 0, NULL);
```
wsdwsd0829: I think the order is important here, will CFSet maintain the order? to make sure the queue process node in FIFO order.

Source/ASRunLoopQueue.mm
```diff
@@ -342,22 +352,15 @@ @implementation ASCATransactionQueue
...
+    
+    // This is going to be a very busy queue – every node in the preload range will enter this queue.
+    // Save some time on first render by reserving space up front.
+    _internalQueue.reserve(64);
```
wsdwsd0829: What does this mean? (64 bits for each pointer?)

Source/ASRunLoopQueue.mm
wsdwsd0829: I see what you doing, resolved.

wsdwsd0829:
LGTM, thanks for the optimize.

Source/ASRunLoopQueue.mm
Adlai-Holler: No it just tells the vector to allocate space in advance for 64 items.

By default, when you add items to a std::vector, it will first reserve 1 spot. Then if you add another, it will reserve 2, move the 1, and delete the old buffer. Then if you add another, it'll reserve 4, move the 2, and delete the old buffer and so-on. To save all these allocations, we skip over the 1 2 4 8 16 and 32 steps.

maicki: We could remove the magic number in here to make it more obvious.

Adlai-Holler: Done.

PR #1352, opened 2019-02-27
Optimize _assertSubnodeState

Adlai-Holler:
This method is actually pretty painful in today's world. In one iPad mini trace, the first page of nodes spent 6.6ms in this call, just in time profiler.

This only affects Yoga users currently.

### DIFF
--- a/Source/ASDisplayNode+Layout.mm
+++ b/Source/ASDisplayNode+Layout.mm
@@ -931,16 +931,21 @@ ASLayoutElementStyleExtensibilityForwarding
   ASDN::MutexLocker l(__instanceLock__);
   NSArray<ASLayout *> *sublayouts = _calculatedDisplayNodeLayout.layout.sublayouts;
   unowned ASLayout *cSublayouts[sublayouts.count];
-  [sublayouts getObjects:cSublayouts];
-  NSUInteger i = 0;
-  BOOL matches = YES;
-  for (ASDisplayNode *subnode in _subnodes) {
-    if (subnode != cSublayouts[i].layoutElement) {
-      matches = NO;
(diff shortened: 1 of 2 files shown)


Source/ASDisplayNode+Layout.mm
```diff
@@ -914,35 +916,40 @@ - (void)_assertSubnodeState
...
-
-  for (ASLayout *sublayout in sublayouts) {
-    id <ASLayoutElement> layoutElement = sublayout.layoutElement;
-    ASDisplayNodeAssert([layoutElement isKindOfClass:[ASDisplayNode class]],
-                        @"All calculatedLayouts should be flattened and only contain nodes!");
-    [layoutSubnodes addObject:(ASDisplayNode *)layoutElement];
+  ASDN::MutexLocker l(__instanceLock__);
+  NSArray<ASLayout *> *sublayouts = _calculatedDisplayNodeLayout.layout.sublayouts;
+  unowned ASLayout *cSublayouts[sublayouts.count];
+  [sublayouts getObjects:cSublayouts range:NSMakeRange(0, AS_ARRAY_SIZE(cSublayouts))];
```
maicki: Should `AS_ARRAY_SIZE(cSublayouts) == sublayouts.count` in this case and could drop `AS_ARRAY_SIZE` in this and also in the `if` check below?

Adlai-Holler: I wanted to avoid adding a new local variable or calling the count method multiple times. Since AS_ARRAY_SIZE is crazy fast I think it hits a good balance.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f94253ae8d47281cb40ee950ad226343/raw/ae610a0aa189f9140a3bad89f066caf69d53fad6/log.txt)

PR #1354, opened 2019-03-01
Add UIDataSourceModelAssociation to ASTableView and ASCollectionView

farktronix:
This adds two new API methods to `ASTableView` and `ASCollectionView` to match the `UIDataSourceModelAssociation` protocol. 

For reference, `UIDataSourceModelAssociation` is
```
@protocol UIDataSourceModelAssociation
- (nullable NSString *) modelIdentifierForElementAtIndexPath:(NSIndexPath *)idx inView:(UIView *)view;
- (nullable NSIndexPath *) indexPathForElementWithModelIdentifier:(NSString *)identifier inView:(UIView *)view;
@end
```

This protocol allows UIKit's state restoration feature to save and resume to the proper scroll position in a table or collection view even if the elements have changed position.

This patch adds node-based versions of those calls that implementers of `AS{Table,Collection}View` can use.

The one odd bit in this patch is that UIKit appears to use a class-based check for conformance to the `UIDataSourceModelAssociation` protocol, which means `ASDelegateProxy` wasn't getting its instance-based `conformsToProtocol:` method called. I had to add the protocol to the proxies directly in order to get UIKit to call the modelIdentifier methods on `ASTableViewProxy` and `ASCollectionViewProxy`.


Source/ASCollectionView.mm
```diff
@@ -805,6 +809,31 @@ - (void)invalidateFlowLayoutDelegateMetrics
...
+    } else {
+        return nil;
+    }
+}
+
+- (nullable NSIndexPath *)indexPathForElementWithModelIdentifier:(NSString *)identifier inView:(UIView *)view {
+    if (_asyncDataSourceFlags.modelIdentifierMethods) {
+        GET_COLLECTIONNODE_OR_RETURN(collectionNode, nil);
+        NSIndexPath *result = [_asyncDataSource indexPathForElementWithModelIdentifier:identifier inNode:collectionNode];
+        result = [self convertIndexPathToCollectionNode:result];
```
Adlai-Holler: The data source should return us an index path in the collection node space, so we can return without converting.

farktronix: Thanks! I've made this change on both `ASTableView` and `ASCollectionView`

Adlai-Holler: Excellent!

PR #1356, opened 2019-03-02
Lock up to yogaRoot during layout to avoid deadlocks.

wiseoldduck:
Yoga layout will try to acquire locks from above node during layout, which is Bad under our conventions. So we must acquire locks for all ancestor nodes before beginning so we do not create a deadlock with some other thread trying to lock downward (which is Good).

1) lock to root for tree
2) lock self to change parent (& consequently root)
3) add lockPair to try-lock node & controller together
4) lock controllers if they exist in lockToRoot...
5) disable some asserts due to lock to root. :(


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5f17f1313979c1dca97c4ecd86bbe1bd/raw/5d38629568ec9034ef59e6466bd6e1ef2f27357f/log.txt)

PR #1358, opened 2019-03-03
Remove redundant accessibility values and clean up

Adlai-Holler:
Our display node instances are big. One (small) reason for this is that accessibility properties are currently stored in two places: in the pending state, before view load, and in the node itself, for layer-backed nodes.

There's no need for this redundancy. The pending state is a fine place to act as the storage for such properties, and we use it that way for other UIView-only properties like `userInteractionEnabled`.

This diff aligns our accessibility storage strategy with those other properties and gets rid of the duplicate storage.

### DIFF
--- a/Source/Private/ASDisplayNode+UIViewBridge.mm
+++ b/Source/Private/ASDisplayNode+UIViewBridge.mm
@@ -49,6 +49,7 @@
 /// returns NO. Otherwise, the pending state can be scheduled and flushed *before* you get a chance
 /// to apply it.
 ASDISPLAYNODE_INLINE BOOL ASDisplayNodeShouldApplyBridgedWriteToView(ASDisplayNode *node) {
+  ASAssertLocked(node->__instanceLock__);
   BOOL loaded = _loaded(node);
   if (ASDisplayNodeThreadIsMain()) {
     return loaded;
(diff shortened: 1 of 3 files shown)


Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -67,10 +68,10 @@ ASDISPLAYNODE_INLINE BOOL ASDisplayNodeShouldApplyBridgedWriteToView(ASDisplayNo
...
 
-#define _setToViewOnly(viewAndPendingViewStateProperty, viewAndPendingViewStateExpr) BOOL shouldApply = ASDisplayNodeShouldApplyBridgedWriteToView(self); \
-if (shouldApply) { _view.viewAndPendingViewStateProperty = (viewAndPendingViewStateExpr); } else { ASDisplayNodeGetPendingState(self).viewAndPendingViewStateProperty = (viewAndPendingViewStateExpr); }
+#define _setToViewOnly(viewAndPendingViewStateProperty, viewAndPendingViewStateExpr) \
+  if (_view && ASDisplayNodeShouldApplyBridgedWriteToView(self)) { _view.viewAndPendingViewStateProperty = (viewAndPendingViewStateExpr); } else { ASDisplayNodeGetPendingState(self).viewAndPendingViewStateProperty = (viewAndPendingViewStateExpr); }
 
-#define _getFromViewOnly(viewAndPendingViewStateProperty) _loaded(self) ? _view.viewAndPendingViewStateProperty : ASDisplayNodeGetPendingState(self).viewAndPendingViewStateProperty
+#define _getFromViewOnly(viewAndPendingViewStateProperty) (_view ? _view.viewAndPendingViewStateProperty : ASDisplayNodeGetPendingState(self).viewAndPendingViewStateProperty)
 
 #define _getFromLayer(layerProperty) _loaded(self) ? _layer.layerProperty : ASDisplayNodeGetPendingState(self).layerProperty
```
maicki: We may should do the same for `_getFromLayer` now like we do for `_getFromViewOnly` etc.  and check for the `_layer`.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/cd68a1db3598e3710dc7dd37a79f7624/raw/ae8e954ab9c1e5ee669b2b43989892eaa271f456/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ea990390a02b9994c84d0983cd96f63c/raw/c44672f4992dcb8dcdc4657a7116c1dad4774b6d/log.txt)

Adlai-Holler:
Update: this diff won't work. The reason is, we throw pending states away and clear their data even for layer-backed nodes. So if we want these accessibility properties to stick around, we need to rethink this policy. Few thoughts:

- The quick fix is to rethink this class as bridged-state instead of pending state, indicating that it holds all data that mirrors CALayer & UIView properties, whether they're pending or not.
- We could replace it with two C++ classes, `ASDN::LayerState` and `ASDN::ViewState`. The latter would only store _view-only_ bridged data.
- Then when we get a view, we can get rid of `ASDN::ViewState` until the next bridged write when it's needed. When we get a layer, we can get rid of `ASDN::LayerState` until the next bridged write.
- In our unconditional clearing of the bridged state [here](https://github.com/TextureGroup/Texture/blob/257f5a9a6de3395b00f670763075677452027e3e/Source/ASDisplayNode.mm#L3467) we would instead rely on the fact that bridged state is always cleared whenever it is applied to a real view or layer.

Thoughts @maicki @garrettmoon @nguyenhuy ?

PR #1361, opened 2019-03-04
Add `setNeedsLayout` to yoga tree changes.

wiseoldduck:
I didn't write this so ask @Adlai-Holler "not Tim Rowe" if there's any questions 😛

### DIFF
--- a/Source/ASDisplayNode+Layout.mm
+++ b/Source/ASDisplayNode+Layout.mm
@@ -19,6 +19,19 @@
 #import <AsyncDisplayKit/ASNodeController+Beta.h>
 #import <AsyncDisplayKit/ASDisplayNode+Yoga.h>
 
+@interface ASDisplayNode (ASLayoutElementStyleDelegate) <ASLayoutElementStyleDelegate>
+@end
+
+@implementation ASDisplayNode (ASLayoutElementStyleDelegate)
+
+#pragma mark <ASLayoutElementStyleDelegate>
+
+- (void)style:(ASLayoutElementStyle *)style propertyDidChange:(NSString *)propertyName {
+  [self setNeedsLayout];
+}
+
+@end
+
 #pragma mark - ASDisplayNode (ASLayoutElement)
 
 @implementation ASDisplayNode (ASLayoutElement)
@@ -42,8 +55,9 @@
 
 - (ASLayoutElementStyle *)_locked_style
 {
+  ASAssertLocked(__instanceLock__);
   if (_style == nil) {
-    _style = [[ASLayoutElementStyle alloc] init];
+    _style = [[ASLayoutElementStyle alloc] initWithDelegate:self];
   }
   return _style;
 }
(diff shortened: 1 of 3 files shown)


Source/ASDisplayNode+Layout.mm
```diff
@@ -19,6 +19,19 @@
...
 
+@interface ASDisplayNode (ASLayoutElementStyleDelegate) <ASLayoutElementStyleDelegate>
+@end
+
+@implementation ASDisplayNode (ASLayoutElementStyleDelegate)
+
+#pragma mark <ASLayoutElementStyleDelegate>
+
+- (void)style:(ASLayoutElementStyle *)style propertyDidChange:(NSString *)propertyName {
+  [self setNeedsLayout];
```
garrettmoon: Can you add some comments or context here? This is causing us infinite layout loops. Is it applicable to non-yoga?

garrettmoon: It actually doesn't seem unreasonable, but it breaks ASTextNode which sets its style during layout.

maicki: Every time `style:propertyDidChange:` is called one of the layout properties changed. If some layout related properties (or some other related properties like e.g. attributed text` on an `ASTextNode` are changing the node needs to be informed that a new layout pass should happen. The most obvious way is to do it in central place like here.

We should fix the `ASTextNode` issue though ... we should try getting setting any style properties out of size calculation. `calculateSizeThatFits:` should just use the current state and return the size based on it and don't adjust anything in the world around it as you could just call `calculateSizeThatFits:` without actually changing anything on the node therefore the next time it would draw with a different bounds as the `constrainedSize` passed in the ascender and descender could be wrong?

@rcancro Any idea how we could get adjusting the ascender / descender out of `calculateSizeThatFits:`. We should try to get the current renderer in `setAttributedText:` or if the bounds changes and update the ascender / descender with it?

garrettmoon: Until a fix for ASTextNode is made available this should be reverted?

Adlai-Holler: - `-setAscender:` only informs the delegate, if the style actually changed.
- `ASTextNode` actually sets it twice, to different values, during layout so that's why we have an infinite loop.
- Fixing the above would still mean double-layout. 
- I think it's correct to say that layout invalidations that happen DURING layout should always be ignored, *if* all operations on the tree used the same lock.
- Since it doesn't, can we hack around this by using the ASLayoutElementContext thing? Does the presence of that signal that we are calculating a layout? @maicki 
- If we can't come up with anything before tomorrow morning, I'm OK with reverting this for now since ASTextNode is such a core type.

maicki: - We don't have anything like `ASLayoutElementContext` setup
- I will get something going that does mean we don't have to revert it completly

garrettmoon: Awesome, thanks y'all!

PR #1362, opened 2019-03-04
_ASCollectionViewCell - The point isn't converted before to send to node, impossible to touch button into the node hierarchy

Idefix60:
Hi,

The first step of this modification was:
- The node can have a transformation, the point is not converted to the scaled node.

Second step:
use convertPoint
`
CGPoint pointOnNode = [self.node.view convertPoint:point fromView:self]
return [self.node hitTest:pointOnNode withEvent:event];
`

But...
With this situation
- _ASCollectionViewCell calls ASDisplayNode
- ASDisplayNode calls ASDisplayView1
- ASDisplayView1 calls ASDisplayNode
- ASDisplayNode calls ASDisplayView1
- ASDisplayView1 calls Super.... finished
We need to calculate the new point (for node with view only, because touch for layerBacked is not supported)

This this modification, we win 1 loop for nothing...
- _ASCollectionViewCell calls ASDisplayView
- ASDisplayView1 calls ASDisplayNode
- ASDisplayNode calls ASDisplayView1
- ASDisplayView1 calls Super.... finished
And the transformation is computed by the OS.

Now, with the call to 'super', [self.node hitTest:point withEvent:event] is called, but with the converted point.

### DIFF
--- a/Source/Details/_ASCollectionViewCell.mm
+++ b/Source/Details/_ASCollectionViewCell.mm
@@ -110,12 +110,12 @@
     return nil;
   }
 
-  return [self.node hitTest:point withEvent:event];
+  return [super hitTest:point withEvent:event];
 }
 
 - (BOOL)pointInside:(CGPoint)point withEvent:(nullable UIEvent *)event
 {
-  return [self.node pointInside:point withEvent:event];
+  return [super pointInside:point withEvent:event];
 }
 
 @end
--- a/Source/Details/_ASCollectionViewCell.mm
+++ b/Source/Details/_ASCollectionViewCell.mm
@@ -110,22 +110,14 @@
     return nil;
   }
 
-  if ([self.node isNodeLoaded] && self.node.view) {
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
cc #1268 for the history here.

The key is to handle your case (convert the point) and also to ensure the node gets a chance to apply `hitTestSlop`. So I think your initial pass – convert the point and forward to node – is the way to go.

Idefix60:
Finally, i think the solution to call self.node.view is the best.
If the view is not created when the user tap on the collectionViewCell, we can create it on the mainThread right here.

Or another solution is:
For pointInside, if node is not Loaded && self.node.view is nil, return CGRectContainsPoint(self.bounds, point);
And for hitTest, if node is not Loaded && self.node.view is nil, return [self pointInside:point withEvent:event] ? self : nil;
Because, if the node is not ready, the final view to hitTest is _ASCollectionCellView :)

PR #1365, opened 2019-03-05
Add clang-format for a common source code format

maicki:
@Adlai-Holler @appleguy @nguyenhuy @garrettmoon I would appreciate for any input on the code format. For now I added a `.clang-format` file to discuss and iterate on.

Preview: https://github.com/TextureGroup/Texture/compare/master...MSClangFormatPreview

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,7 +1,37 @@
 BasedOnStyle:  Google
 
-# Objc allows 120 instead of the traditional 80.
+## Indentation
 ColumnLimit: 120
+IndentWidth: 2
+
+## Line Break
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass:      true
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  true
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      true
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+  SplitEmptyFunction: false
+
+## ObjC
+ObjCBinPackProtocolList: Always
+ObjCSpaceAfterProperty: false
+
+## Comments
+AlignTrailingComments: true
+
+## Arguments / Parameters
+
+# If false, a function call’s or function definition’s arguments will either
+# all be on the same line or will have one line each.
+BinPackArguments: false
 
 # If false, a function call’s or function definition’s parameters will either
 # all be on the same line or will have one line each.
@@ -14,3 +44,10 @@ BinPackParameters: false
 #             bar,
 #             baz);
 AllowAllParametersOfDeclarationOnNextLine: true
+
+## Single Line
+AllowShortBlocksOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,16 @@
+BasedOnStyle:  Google
+
+# Objc allows 120 instead of the traditional 80.
+ColumnLimit: 120
+
+# If false, a function call’s or function definition’s parameters will either
+# all be on the same line or will have one line each.
+BinPackParameters: false
+
+# Allow putting all parameters of a function declaration onto the next line even
+# if BinPackParameters is false.
+# For example this should be ok:
+# someFunction(foo,
+#             bar,
+#             baz);
+AllowAllParametersOfDeclarationOnNextLine: true
--- a/format
+++ b/format
@@ -1,0 +1,42 @@
+#!/bin/bash
+#
+# Script uses clang-format to format all of the source files in a
+# default format.
+#
+
+# The absolute path of the directory containing this script.
+DIR="$( cd "$( dirname "$0" )" && pwd)"
+# Define top level project directory
+PROJECT_DIR="${DIR}"
+
+METHOD="GIT"
+if [ -n "$1" ]; then
+ METHOD=$1
+fi
+
+# Paths to format
+MONITOR=()
+MONITOR+=( "${PROJECT_DIR}/Source" )
+
+if [ "$METHOD" == "ALL" ]; then
+    # Format all Source files
+    for FILE in $(find ${MONITOR[*]} -type f \( -iname "*.mm" -o -iname "*.m" -o -iname "*.h" \))
+    do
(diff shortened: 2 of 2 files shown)


format
```diff
@@ -0,0 +1,20 @@
...
+PROJECT_DIR="${DIR}"
+
+# Paths to format
+MONITOR=()
+MONITOR+=( "${PROJECT_DIR}/Source" )
+
+# Format all files
+for FILE in $(find ${MONITOR[*]} -type f \( -iname "*.mm" -o -iname "*.m" -o -iname "*.h" \))
+do
+    clang-format -style=file "$FILE" > tempfile && mv tempfile "$FILE"
```
Adlai-Holler: - Looks like `clang-format` supports a `-i` for inline updating of the file. Let's use that.
- clang-format lets you invoke it for multiple files in one invocation. If we can find a way to do that, it may be more performant. Particularly if we can query git for the list of files modified in this branch and pass only that.

Ideally a developer would be able to `./format` constantly while working (or even as some kind of trigger on saving) so performance is critical.

maicki: Changed a bit:
- Use `-i` for files
- Added different modes for format. `ALL` just goes through all files within `Source/` currently and formats them. `GIT` only goes through added and modified files based on git.

.clang-format
```diff
@@ -0,0 +1,16 @@
+BasedOnStyle:  Google
```
appleguy: Is this necessary? Not a blocker, but also seems irrelevant for most framework users / authors.

.clang-format
```diff
@@ -0,0 +1,16 @@
...
+
+# Objc allows 100 instead of the traditional 80.
+ColumnLimit: 100
```
appleguy: IMO, 120 is a better value here to encourage descriptive method and variable names. I think 140 would be even better than that, considering how universally-wide monitors are these days (things derived on / close to 80 column limits should have died a decade ago!)

This would also reduce the "noise" in the PR, e.g. less disruption to the blame log.

appleguy:
@maicki One of the main uses of blame is to see when a given method was introduced to the codebase — especially in a codebase as large as and old as Texture, I think it makes sense to avoid arbitrary changes in brace style.

In particular, if we skipped moving the open bracket and had a longer line length in this change, there would be much less impact to the practical usability of blame / history. Thoughts on doing that, and going ahead with the codebase-wide cleanup PR (rather than waiting for it to be individually applied on modified files?)

.clang-format
maicki: Yes it's basically the style used for all options not specifically set in the configuration: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options

.clang-format
maicki: I think it would make sense to go with 120.

maicki: Changed to 120.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/dd1ab6af4e5cd4085b8bed9e4d1ed61e/raw/d66531d0382a20304bcddd71e9cb74309ef9fd5d/log.txt)

.clang-format
```diff
@@ -0,0 +1,16 @@
+BasedOnStyle:  Google
+
+# Objc allows 100 instead of the traditional 80.
```
nguyenhuy:
```suggestion
# Objc allows 120 instead of the traditional 80.
```

format
```diff
@@ -0,0 +1,42 @@
...
+
+# Paths to format
+MONITOR=()
+MONITOR+=( "${PROJECT_DIR}/Source" )
+
+if [ "$METHOD" == "ALL" ]; then
+    # Format all Source files
+    for FILE in $(find ${MONITOR[*]} -type f \( -iname "*.mm" -o -iname "*.m" -o -iname "*.h" \))
+    do
+        clang-format -style=file -i "$FILE" &
```
nguyenhuy:
```suggestion
        clang-format -style=file -i "$FILE"
```

format
```diff
@@ -0,0 +1,42 @@
...
+METHOD="GIT"
+if [ -n "$1" ]; then
+ METHOD=$1
+fi
+
+# Paths to format
+MONITOR=()
+MONITOR+=( "${PROJECT_DIR}/Source" )
+
+if [ "$METHOD" == "ALL" ]; then
```
nguyenhuy: Are we planning to run this script for "all" anytime soon? If not then maybe don't include this mode for now for simpler script?

nguyenhuy: I guess, the right question would be: do we want to run this script for "all" anytime soon? Having a huge diff that touches so many files will affect the usability of git blame/history. I think a much more lenient approach would be to have it run on new PRs for a while, and then run "all" months/a year from now. What do you think?

nguyenhuy: Also, can you share a new diff with the latest format? a08cc7b is outdated right?

maicki: The plan to run it all at once together with the introduction of the script.

Correct it will affect to the blame / history, but having it one at for all I think would make more sense as gradually and having problems with blame / history one file after another.

https://github.com/TextureGroup/Texture/commit/a08cc7b850950d6202a6c2e997086d1a7c51c7f3 is the latest, I updated the diff with the latest changes morning.

format
maicki: The `&` is there on purpose. It runs the command in the background and does not wait for every file.

nguyenhuy: ahh, got it!

format
nguyenhuy: ok. In that case, what do you think about keeping the open brackets as is?

Adlai-Holler:
Agree that the spaces-before-assignments should be kept, and same with the spaces before inline comments.

The reflowing of block comments is great. I like the on-the-line brace style a lot. 120 line length seems good. Property attributes with no space after `@property` is good.

maicki:
@Adlai-Holler @nguyenhuy @appleguy @garrettmoon Ok I think I found the format that is like our current one with a couple of small changes like line length and no space after `@property`.

You can see the diff here: https://github.com/TextureGroup/Texture/compare/master...MSClangFormatPreview

Adlai-Holler:
Sent an email with a straw poll for brace style. Please respond! We'll go with the answer Thursday at noon.

format
wiseoldduck: I think `git blame -w` would sidestep most of the pain, but that's only when running it from the command-line, for i.e. github, sourcetree, IntelliJ, where I think most people actually use it, there's no way to inject that flag into the command :(

nguyenhuy:
LGTM now. Thanks for working on this, @maicki.

PR #1366, opened 2019-03-05
Rename ASDN C++ namespace to AS

Adlai-Holler:
Referring to the framework as ASDisplayNode is pretty out-dated and verbose. See CoreAnimation which uses CA for their internal namespace.

As we use more C++, having a visually clean namespace will matter more.

PR #1367, opened 2019-03-05
Replace ASDynamicCast macros with inline C++ functions

Adlai-Holler:
Macros bad, functions good. I've confirmed that when inlined, ARC has full optimization control so it can avoid retains and releases just like in a macro.

The namespace I used for this function is `AS` and not `ASDN` like mutex. #1366 changes mutex to match – I believe that's a better name for our C++ root namespace.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -38,6 +38,9 @@
 #import <AsyncDisplayKit/ASLayout.h>
 #import <AsyncDisplayKit/ASThread.h>
 
+using AS::DynamicCast;
+using AS::DynamicCastStrict;
+
 /**
  * A macro to get self.collectionNode and assign it to a local variable, or return
  * the given value if nil.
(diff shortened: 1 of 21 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/7b7e42fb606ca6f8dff5b7639f315174/raw/8b98bee1ef744fcef4847c77d533027f5b513749/log.txt)

Adlai-Holler:
The CI failure is from an old Xcode version (CI runs 9.0). IIRC that compiler bug was fixed in 9.3 (downloading to verify). If we can bump the CI to 9.4.1 without too much trouble, that would be great. Might fix the spurious failures too. @garrettmoon

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/74f2fa3b1c790ae155dd58d4bcb36ffd/raw/b87340f757707a7d8c42011e22fe9e44c56b7318/log.txt)

Adlai-Holler:
ARC just isn't smart enough for this. Even inlined functions have to do the ARC handoff dance. Closing.

PR #1368, opened 2019-03-06
Allow ASDisplayNode subclasses to modify _contentsScaleForDisplay

mikezucc:
Expose content scale for ASImageNodeDrawParameters so that Texture consumers can override content scale in a subclassed ASImageNode.

We would like to experiment with imposing limits to resource allocation for certain device families. One strategy is to experiment with reducing allocation used by the ASImageNode when preparing its image for draw in the render loop.

### DIFF
--- a/Source/ASImageNode.h
+++ b/Source/ASImageNode.h
@@ -12,7 +12,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class ASWeakMapEntry;
 @protocol ASAnimatedImageProtocol;
 
 /**
@@ -24,23 +23,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
 
-/**
- *  Weakly reference the cached, processed image.
- */
-typedef void (^ASImageNodeDrawParametersBlock)(ASWeakMapEntry *entry);
-
-/**
- *  @abstract Specific description of draw parameters only for image nodes.
- *
- *  @discussion The `_contentsScale` is exposed here so that consumers of Texture
- *  may impose special limits to the drawn image sized based on the
- *  draw parameters.
- */
-@interface ASImageNodeDrawParameters : NSObject {
-  @public CGFloat _contentsScale;
-}
-
-@end
 
 /**
  * @abstract Draws images.
--- a/Source/ASImageNode.h
+++ b/Source/ASImageNode.h
@@ -25,11 +25,6 @@ NS_ASSUME_NONNULL_BEGIN
 typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
 
 /**
- *  Weakly reference the cached, processed image.
- */
-typedef void (^ASImageNodeDrawParametersBlock)(ASWeakMapEntry *entry);
-
-/**
  *  @abstract Specific description of draw parameters only for image nodes.
  *
  *  @discussion The `_contentsScale` is exposed here so that consumers of Texture
--- a/Source/ASImageNode.h
+++ b/Source/ASImageNode.h
@@ -25,6 +25,11 @@ NS_ASSUME_NONNULL_BEGIN
 typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
 
 /**
+ *  Weakly reference the cached, processed image.
+ */
+typedef void (^ASImageNodeDrawParametersBlock)(ASWeakMapEntry *entry);
+
+/**
  *  @abstract Specific description of draw parameters only for image nodes.
  *
  *  @discussion The `_contentsScale` is exposed here so that consumers of Texture
--- a/Source/ASImageNode.h
+++ b/Source/ASImageNode.h
@@ -12,6 +12,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class ASWeakMapEntry;
 @protocol ASAnimatedImageProtocol;
 
 /**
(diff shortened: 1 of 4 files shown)


maicki:
I would rather have us not exposing any internal logic. Maybe I'm missing something but `[ASImageNodeDrawParameters contentScale] = [ASDisplayNode contentsScaleForDisplay]`. Why can we not use `[ASDisplayNode contentsScaleForDisplay]` not sufficient to use?

mikezucc:
The flexibility is specific for the case of rendering an image of an `ASImageNode`. I don't see a case where all `ASDisplayNode` and its subclasses would need to render to a different scale than the screen.

Also `contentsScaleForDisplay` would be exposing for too many cases. Exposing just this one property allows consumers to do something like this in an `ASImageNode` subclass:
```objc
// MyCoolImageNode.m
- (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
{
    //...
    ASImageNodeDrawParameters *drawParams = (ASImageNodeDrawParameters *)[super drawParametersForAsyncLayer:layer];
    drawParams->_contentsScale = [Experimenter screenScaleForDrawParameters:drawParams];
    return drawParams;
}
```

Source/ASImageNode.h
```diff
@@ -23,6 +24,23 @@ NS_ASSUME_NONNULL_BEGIN
  */
 typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
 
+/**
+ *  Weakly reference the cached, processed image.
+ */
+typedef void (^ASImageNodeDrawParametersBlock)(ASWeakMapEntry *entry);
```
mikezucc: this doesnt need to be public anymore

Source/ASImageNode.h
```diff
@@ -23,6 +24,18 @@ NS_ASSUME_NONNULL_BEGIN
...
 typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
 
+/**
+ *  @abstract Specific description of draw parameters only for image nodes.
+ *
+ *  @discussion The `_contentsScale` is exposed here so that consumers of Texture
+ *  may impose special limits to the drawn image sized based on the
+ *  draw parameters.
+ */
+@interface ASImageNodeDrawParameters : NSObject {
```
garrettmoon: Could we expose this in a +beta header?

mikezucc: Sure!

nguyenhuy:
I don't think we should expose a private ivar like this. 

I think the root issue lies at [this line](https://github.com/TextureGroup/Texture/blob/master/Source/ASImageNode.mm#L304):

```
drawParameters->_contentsScale = _contentsScaleForDisplay;
```

Because we access the `_contentsScaleForDisplay` ivar directly here, the only way to have another value picked up is by using its setter beforehand. Would it be better to change the line to use the property's getter instead?

```
 drawParameters->_contentsScale = self.contentsScaleForDisplay;
```

Subclasses can then override the getter and return whatever it thinks is the best value.

nguyenhuy:
Hm, I think I understand what you meant by "`contentsScaleForDisplay` would be exposing for too many cases". Exposing `ASImageNodeDrawParameters` is fine, but exposing `_contentsScale` directly like this isn't ideal IMO. Can we set up a readwrite property for it at least?

maicki:
`contentsScaleForDisplay` is a readwrite property on `ASDisplayNode` so you can set that from outside and it will be used in the time display is happening.

Something like this within init:
```
_someImageNodeProperty = [[ASImageNode alloc] init];
_someImageNodeProperty. contentsScaleForDisplay = [Experimenter screenScaleForImageScreenScaleExperiment];
```
As you don't expose more properties as screen scale it seems like you also don't need the drawing parameters as input to the Experimenter.

mikezucc:
Maybe I'm looking at an old commit, but I thought `contentsScaleForDisplay` is `readonly` in public header `ASDisplayNode+Subclasses.h`?

mikezucc:
I think I can just get away with redeclaring as a readwrite prop in the subclass

maicki:
@mikezucc That should work.

mikezucc:
I'd have to move the `@protected` declaration from ASDisplayNodeInter.h -> ASDisplayNode.h to get this to work tho

mikezucc:
Ok cool so now I moved the `@protected` declaration for `_contentsScaleForDisplay` to `ASDisplayNode.h` so that subclasses can create their own readwrite version if necessary. I like this solution much better.

mikezucc:
Revealing the draw parameters was not very clean heh

nguyenhuy:
I’m actually fine with making the property readwrite for subclasses. That way clinets won’t need to re-declare it.

maicki:
Concur with @nguyenhuy  we should make it readwrite for subclasses in `ASDisplayNode+Subclasses.h` and remove it publicly from `ASDisplayNode.h`

mikezucc:
Ah yeah right. 👍

nguyenhuy:
Also, just to make sure we don't introduce an unnecessary change: wouldn't `-setNeedsDisplayAtScale:` be enough? The only thing I can think of that makes you want a setter is that you don't want to trigger a display pass (`-setNeedsDisplay`) every time you set a difference contents scale.

maicki:
We could also use `- (void)setNeedsDisplayAtScale:(CGFloat)contentsScale;` that would have the side effect of a `setNeedsDisplay`.

mikezucc:
whats the impact of the additional `setNeedsDisplay` ? In the experiment I'm looking at, the `_contentsScaleForDisplay` is written to during init or just after init, but not during a render loop

nguyenhuy:
@mikezucc: well then the impact is immeasurable, if any. Let’s use it then.

mikezucc:
By immeasureable you mean that using the setter would be robust/safe? It seems to do some additional flag setting and dispatching. I would only be able to safely use the setter outside of the initialization, which isn't a deal breaker.

maicki:
Calling `setNeedsDisplay` on a node that is not loaded yet should be pretty cheap. We don't call through to the view or layer's `setNeedsDisplay` as well as we don't schedule it for display as the interface state does not include display.

So calling `setNeedsDisplayAtScale:` in `init` should be fine.

mikezucc:
Ok cool I'll use that then 😌

nguyenhuy:
Yeah, @maicki explained it much better than me.

PR #1297, opened 2018-12-22
Don't add extraneous truncation token during kCTLineTruncationMiddle.

wiseoldduck:

PR #1371, opened 2019-03-06
Revert "Don't add extraneous truncation token during kCTLineTruncationMiddle."

Adlai-Holler:
Reverts TextureGroup/Texture#1297

The CI passed before that landed but it broke some snapshot tests. cc @garrettmoon @wiseoldduck


wiseoldduck:
Yeah I added snapshot tests for ASTextNode2 that will need to be updated

Adlai-Holler:
cc #1375 it's currently running CI so let's aim to just land that fix-forward and skip this.

PR #1373, opened 2019-03-06
Assert for context creation failure during displayBlock.

wiseoldduck:
Nothing good can possibly happen if we continue without one.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/8c835ac16c5b4f779ed713c4103a0c9e/raw/c73283afddbbae39617fe7e5055440ff70015c45/log.txt)

PR #1377, opened 2019-03-07
Expose initial constrained size before layout...

wiseoldduck:
in case anyone is interested later during the layout on the same thread.

Optionally, clients are expected to clean it up after usage to avoid accessing outdated data.


wiseoldduck:
(This is @wsdwsd0829 's original work)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/fc260418af029b6642670cf356e0e665/raw/44093f340e223b478aac4546ef70c39a8c4b4c22/log.txt)

Adlai-Holler:
Clients don't need to clean it up, we clean it up on line 176.

This is definitely a bit of a hack but it's a really high-quality one! @garrettmoon @nguyenhuy if you've ever needed to know the initial constrained size for a cell node at init time you can use this.

PR #1378, opened 2019-03-07
Correct View vs. Layer here (tho it doesn't make any difference)

wiseoldduck:
Older FBSnapshot versions, however, don't use VerifyViewOrLayer


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0fdf9741c43cf691efe435e38bb2ecaf/raw/ab7c7467be632cb49c858622319f941bca2a8946/log.txt)

PR #1369, opened 2019-03-06
Launches switching ASNetworkImageNode callbacks to global queue.

garrettmoon:
This showed performance gains for us on startup when the main thread was under heavy contention.

PR #1382, opened 2019-03-07
Revert "Launches switching ASNetworkImageNode callbacks to global queue."

Adlai-Holler:
Reverts TextureGroup/Texture#1369

This is a pretty huge API change! I'm uncomfortable doing it so quietly. Some users undoubtedly do main-only things in these callbacks. For example, they might do `if (self.hidden) { self.hidden = NO; }` which would be accessing a bridged property off-main and boom.

I propose instead adding an opt-in class flag to trigger the feature. It's still an optional feature, it's just not an experimental one anymore. Like so:

```objc
// ASImageNode.h
AS_EXTERN void ASImageNodeEnableBackgroundCallbacks(void);
```


wsdwsd0829:
BTW, we had ASNetworkImageNode.useMainThreadDelegateCallbacks to optionally diable setting to NO.  to control image delegate call backs.

garrettmoon:
I don't think this is an API change, it's used entirely internally as far as I can tell?

Adlai-Holler:
Oh, you're right, my bad! let's go forward with this.

PR #1383, opened 2019-03-07
Remove experimental features (exp_skip_a11y_wait && exp_new_default_cell_layout_mode)

ernestmama:
We were suspecting these two changes were impact the performance. Result shows that it does not regress the performance.


Source/ASCollectionView.mm
```diff
@@ -1886,10 +1881,6 @@ - (BOOL)dataController:(ASDataController *)dataController shouldSynchronouslyPro
     return NO;
   }
   if (ASCellLayoutModeIncludes(ASCellLayoutModeSyncForSmallContent)) {
-    // Reload data is expensive, don't block main while doing so.
-    if (changeSet.includesReloadData) {
```
maicki: Why was this removed?

Source/ASTableView.mm
```diff
@@ -1678,20 +1678,14 @@ - (BOOL)dataControllerShouldSerializeNodeCreation:(ASDataController *)dataContro
 
 - (BOOL)dataController:(ASDataController *)dataController shouldSynchronouslyProcessChangeSet:(_ASHierarchyChangeSet *)changeSet
 {
-  if (ASActivateExperimentalFeature(ASExperimentalNewDefaultCellLayoutMode)) {
-    // Reload data is expensive, don't block main while doing so.
-    if (changeSet.includesReloadData) {
```
maicki: Why was this removed?

Source/ASCollectionView.mm
ernestmama: This code was not there when the new layout changes was introduced. This is added when we created the experiment afterward. I checked with @nguyenhuy that we did not experiment without this block of code. I will put it back.

Source/ASTableView.mm
ernestmama: This code was not there when the new layout changes was introduced. This is added when we created the experiment afterward. I checked with @nguyenhuy that we did not experiment without this block of code. I will put it back.

Source/ASCollectionView.mm
```diff
@@ -317,12 +317,7 @@ - (instancetype)_initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionV
   
   [self _configureCollectionViewLayout:layout];
   
-  if (ASActivateExperimentalFeature(ASExperimentalNewDefaultCellLayoutMode)) {
-    _cellLayoutMode = ASCellLayoutModeSyncForSmallContent;
-  } else {
-    _cellLayoutMode = ASCellLayoutModeNone;
-  }
-  
+  _cellLayoutMode = ASCellLayoutModeSyncForSmallContent;
```
maicki: This one can go away as the default one will be `ASCellLayoutModeNone`. In fact the `ASCellLayoutModeSyncForSmallContent` is not needed at all anymore

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0e7b368848d2f55cdca123b0b0dc9370/raw/8971a3f378a50ebddf5eb6412a9767dcc4e1ab32/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4541d982fa57fefe256ca686ca15175f/raw/d34ec3e9b77d3455c9b4040d4cfee3ea0c5e13c6/log.txt)

PR #1384, opened 2019-03-07
Adds support for using UIGraphicsImageRenderer in ASTextNode.

garrettmoon:
In many cases this reduces the backing store of text nodes by 1/2.

### DIFF
--- a/Schemas/configuration.json
+++ b/Schemas/configuration.json
@@ -25,7 +25,8 @@
                     "exp_disable_a11y_cache",
                     "exp_skip_a11y_wait",
                     "exp_new_default_cell_layout_mode",
-                    "exp_dispatch_apply"
+                    "exp_dispatch_apply",
+                    "exp_text_drawing"
                 ]
     		}
 		}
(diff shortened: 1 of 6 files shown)


garrettmoon:
I think the biggest change that *won't* be captured by an experiment is the switch from `drawRect:` to `displayWithParameters:`. Anyone have any concerns with that?

garrettmoon:
If this goes well I'll likely attempt this in ASImageNode as well.

maicki:
We still support iOS 9 within our Podspec and UIGraphicsImageRenderer is iOS10+  so I think we need to guard against it for the time.

garrettmoon:
@Adlai-Holler Yes, if all experiments are successful 🤞 then I'll keep moving this up the stack.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4c794e647c96f85ccaaa582730479ad2/raw/d06bf440ed08f93913b569515198846a463c02dc/log.txt)

PR #1385, opened 2019-03-07
Whoops, someone forgot an else!

garrettmoon:
Luckily it seems this didn't cause huge memory bloat because there's
a check below that keeps the display flag from being set in low memory
conditions when the node is off the screen.


Adlai-Holler:
<- This guy!

PR #1390, opened 2019-03-08
ASDisplayNode view and layer properties are nullable

### LINKED ISSUE #1389: clang analyzer issue when returning nil
When running the clang static analyzer, the following warnings are generated:

```
~/Texture/Source/ASDisplayNode.mm:646:5: warning: nil returned from a method that is expected to return a non-null value
    return nil;
    ^~~~~~~~~~
~/Texture/Source/ASDisplayNode.mm:685:5: warning: nil returned from a method that is expected to return a non-null value
    return nil;
    ^~~~~~~~~~
2 warnings generated.

```

This code was refactored with ef2ed54, but the issue existed since the initial commit (15565873). However, assuming `nonnull` was added to `ASDisplayNode` with 814f725. 

Should `ASDisplayNode.view` and `ASDDisplayNode.layer` be `nullable`? @garrettmoon @Adlai-Holler @maicki Any opinions? Thanks!

bolsinga:
Fixes #1389

### DIFF
--- a/Source/ASDisplayNode.h
+++ b/Source/ASDisplayNode.h
@@ -209,7 +209,7 @@ AS_EXTERN NSInteger const ASDefaultDrawingPriority;
  * @warning The first access to it must be on the main thread, and should only be used on the main thread thereafter as 
  * well.
  */
-@property (readonly) UIView *view;
+@property (readonly, nullable) UIView *view;
 
 /** 
  * @abstract Returns whether a node's backing view or layer is loaded.
@@ -234,7 +234,7 @@ AS_EXTERN NSInteger const ASDefaultDrawingPriority;
  * @warning The first access to it must be on the main thread, and should only be used on the main thread thereafter as 
  * well.
  */
-@property (readonly) CALayer * layer;
+@property (readonly, nullable) CALayer * layer;
 
 /**
  * Returns YES if the node is – at least partially – visible in a window.


maicki:
We have to be careful with this. The reason we didn’t change that in the past is that in Swift everyone who is using Texture and accesses the view / layer needs to unwrap and this could have a huge change for a possible large app.

garrettmoon:
@maicki wouldn't they hit an exception if they tried to access it in Swift though? Or does the runtime protect against that in some way?

maicki:
That would hit an exception you are right, but if you look into the accessors of view and layer in `ASDisplayNode *` this case is happening only if `_locked_shouldLoadViewOrLayer` returns `YES` and this means either the node is within the deallocating process or it's rasterized. Another reason we didn't add nullable, in theory it can be null and the static analyzer is right, but in a practical case I didn't see a case where view / layer was nil in an issue ... you may also can argue that if you access the view / layer in the case of `_locked_shouldLoadViewOrLayer == NO` it should crash.

bolsinga:
I agree this is a major version change, and would require Swift users to handle `nil` here. As it stands now, Texture **may** actually return `nil`, but the Swift apps will not handle it properly and likely crash.

If, as suggested, that the `view` and `layer` properties cannot be `nil`, if the code that returns `nil` in these property accessors is removed, will any unit tests fail?

bolsinga:
When is the next major version bump expected? Are there a group of changes that would apply to a major version bump?

PR #1393, opened 2019-03-10
Make experiment checks faster

Adlai-Holler:
- Use `dispatch_once` when not debugging.
- Clean up the singleton – replace the method + macro with an inline function.

### DIFF
--- a/Source/ASConfigurationInternal.mm
+++ b/Source/ASConfigurationInternal.mm
@@ -12,13 +12,14 @@
 #import <AsyncDisplayKit/ASConfigurationDelegate.h>
 #import <stdatomic.h>
 
+static ASConfigurationManager *ASSharedConfigurationManager;
+static dispatch_once_t ASSharedConfigurationManagerOnceToken;
+
 NS_INLINE ASConfigurationManager *ASConfigurationManagerGet() {
-  static ASConfigurationManager *inst;
-  static dispatch_once_t onceToken;
-  dispatch_once(&onceToken, ^{
-    inst = [[ASConfigurationManager alloc] init];
+  dispatch_once(&ASSharedConfigurationManagerOnceToken, ^{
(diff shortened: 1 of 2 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/dbea7eaf8a02d04aca38b0284a44b64f/raw/d66531d0382a20304bcddd71e9cb74309ef9fd5d/log.txt)

Source/ASConfigurationInternal.mm
```diff
@@ -103,12 +99,12 @@ + (void)test_resetWithConfiguration:(ASConfiguration *)configuration
 
 @end
 
-BOOL ASActivateExperimentalFeature(ASExperimentalFeatures feature)
+BOOL _ASActivateExperimentalFeature(ASExperimentalFeatures feature)
```
maicki: Should we call this `ASActivateExperimentalFeatures`, add an "s" at the end. I think that makes it more clear that we are passing in multiple features.

Source/ASConfigurationInternal.mm
```diff
@@ -103,12 +99,12 @@ + (void)test_resetWithConfiguration:(ASConfiguration *)configuration
...
 {
-  return [ASGetSharedConfigMgr() activateExperimentalFeature:feature];
+  return [ASConfigurationManagerGet() activateExperimentalFeature:feature];
```
maicki: Just a nit: Also we may should rename `activateExperimentalFeature:` to `activateExperimentalFeatures:`.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/3f2f74192bdf8fb280359f38fac077e0/raw/ebb35f933259ffac8e2c8da9f31753576bf8f186/log.txt)

PR #1395, opened 2019-03-11
Experiment with releasing pending state of all nodes

garrettmoon:
Instead of specifically just range managed. We're seeing ASPendingState become a large portion of our memory footprint when digging deep into view controller stacks. If this significantly reduces our OOM we may want to look at how we can more effectively choose when to clear these out vs when to release them entirely.

If we're lucky we'll also be able to check if this regresses scroll performance. If it doesn't we may even just want to ship this as is.


garrettmoon:
I'm not sure this is going to do what I hoped it would, let's not merge yet.

PR #1396, opened 2019-03-11
Add layer-action support to nodes

Adlai-Holler:
- Make `_ASDisplayView` forward `actionForLayer:forKey:` to the node, so now nodes get that call for view and layer backed nodes.
- Remove `_ASDisplayView will/didMoveToWindow` since we can derive that information from `actionForLayer:` just like layer-backed nodes do.
- Add a base-implementable node method `layerActionForKey:` that the node will call.
- Add bridging for the `CALayer.actions` property so you can set up your actions off-main.

This makes it possible to add pretty broad animation support outside of the `transitionLayoutWithSizeRange:` pipeline.

### DIFF
--- a/Source/Details/_ASDisplayView.mm
+++ b/Source/Details/_ASDisplayView.mm
@@ -159,10 +159,7 @@ static _ASDisplayViewMethodOverrides GetASDisplayViewMethodOverrides(Class c)
 
   // Even though the UIKit action will take precedence, we still unconditionally forward to the node so that it can
   // track events like kCAOnOrderIn.
-  id<CAAction> nodeAction = (id)kCFNull;
-  if (ASDisplayNode *node = _asyncdisplaykit_node) {
-    nodeAction = [node actionForLayer:layer forKey:event];
-  }
+  id<CAAction> nodeAction = [_asyncdisplaykit_node actionForLayer:layer forKey:event];
 
   // If UIKit specifies an action, that takes precedence. That's an animation block so it's explicit.
   if (uikitAction && uikitAction != (id)kCFNull) {
--- a/Source/Details/_ASDisplayView.mm
+++ b/Source/Details/_ASDisplayView.mm
@@ -153,22 +153,22 @@ static _ASDisplayViewMethodOverrides GetASDisplayViewMethodOverrides(Class c)
 
 #pragma mark - UIView Overrides
 
-- (void)willMoveToWindow:(UIWindow *)newWindow
+- (id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event
 {
-  ASDisplayNode *node = _asyncdisplaykit_node; // Create strong reference to weak ivar.
-  BOOL visible = (newWindow != nil);
-  if (visible && !node.inHierarchy) {
-    [node __enterHierarchy];
-  }
-}
+  id<CAAction> uikitAction = [super actionForLayer:layer forKey:event];
 
-- (void)didMoveToWindow
-{
-  ASDisplayNode *node = _asyncdisplaykit_node; // Create strong reference to weak ivar.
-  BOOL visible = (self.window != nil);
-  if (!visible && node.inHierarchy) {
-    [node __exitHierarchy];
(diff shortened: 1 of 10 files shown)


Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -944,6 +944,18 @@ - (void)setInsetsLayoutMarginsFromSafeArea:(BOOL)insetsLayoutMarginsFromSafeArea
...
+- (NSDictionary<NSString *,id<CAAction>> *)actions
+{
+  _bridge_prologue_read;
+  return _getFromLayer(actions);
+}
+
+- (void)setActions:(NSDictionary<NSString *,id<CAAction>> *)actions
+{
+  _bridge_prologue_write;
+  _setToLayer(actions, [actions copy]);
```
maicki: I don't think you need to copy here as CALayer's actions is: `@property(copy) NSDictionary<NSString *,id<CAAction>> *actions;`

Source/Private/_ASPendingState.mm
```diff
@@ -1287,69 +1305,7 @@ - (BOOL)hasSetNeedsDisplay
...
-  || flags.setAccessibilityLanguage
-  || flags.setAccessibilityElementsHidden
-  || flags.setAccessibilityViewIsModal
-  || flags.setShouldGroupAccessibilityChildren
-  || flags.setAccessibilityIdentifier
-  || flags.setAccessibilityNavigationStyle
-  || flags.setAccessibilityHeaderElements
-  || flags.setAccessibilityActivationPoint
-  || flags.setAccessibilityPath);
+  return memcmp(&_flags, &kZeroFlags, sizeof(ASPendingStateFlags));
```
maicki: Nice

Source/Private/ASDisplayNode+UIViewBridge.mm
Adlai-Holler: True, done.

Source/Private/_ASPendingState.mm
Adlai-Holler: yeah buddy!

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/795189854281f1cc83c39c8180e4f980/raw/fd7b7eb860a2f1c3353767cff6d947ae0ab10b58/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/414e20f25377daf72daaee892c556563/raw/97951110aaf64a09d50cd431cab9a28c054ac59f/log.txt)

Adlai-Holler:
@garrettmoon It looks like the CI isn't taking to 9.4.1 too good? Cut off during asset catalog copy: https://gist.githubusercontent.com/pinterest-ios/414e20f25377daf72daaee892c556563/raw/97951110aaf64a09d50cd431cab9a28c054ac59f/log.txt

Source/Details/_ASDisplayView.mm
```diff
@@ -153,22 +153,22 @@ - (NSString *)description
...
 {
-  ASDisplayNode *node = _asyncdisplaykit_node; // Create strong reference to weak ivar.
-  BOOL visible = (newWindow != nil);
-  if (visible && !node.inHierarchy) {
-    [node __enterHierarchy];
+  id<CAAction> uikitAction = [super actionForLayer:layer forKey:event];
+
+  // Even though the UIKit action will take precedence, we still unconditionally forward to the node so that it can
+  // track events like kCAOnOrderIn.
+  id<CAAction> nodeAction = (id)kCFNull;
```
tnorman42: Returning kCFNull stops the action search (meaning it will not check the layers actions array), so it's not clear that this is the correct default here.

Adlai-Holler: You're right – the goal was to avoid CALayers default actions, but adding kCFNull to defaultActionForKey: was enough to get that done. Fixed.

PR #1399, opened 2019-03-12
Naming ivar context can conflict with subclasses

garrettmoon:
Naming this context will cause us a bunch of problems. Can we rename before we release this?


Adlai-Holler:
Ah jeez. Maybe we should wrap all the ivars in a C++ struct, since this has happened before.

CALayer uses `_attrs` for example. To ask the rude question – is the migration pretty big?

garrettmoon:
@Adlai-Holler Yes. Remember context logging?

Adlai-Holler:
Ha! True. Alright this is all good, approving.

PR #1409, opened 2019-03-15
Avoid an unnecessary lock & unlock pair in ASMainSerialQueue

nguyenhuy:
There is no point to lock after `runBlocks` only to unlock right afterward.

### DIFF
--- a/Source/Details/ASMainSerialQueue.mm
+++ b/Source/Details/ASMainSerialQueue.mm
@@ -40,14 +40,12 @@
 
 - (void)performBlockOnMainThread:(dispatch_block_t)block
 {
-
-  AS::UniqueLock l(_serialQueueLock);
-  [_blocks addObject:block];
   {
-    l.unlock();
-    [self runBlocks];
-    l.lock();
+    AS::UniqueLock l(_serialQueueLock);
+    [_blocks addObject:block];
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/9e2dc8a9d5a3496fb097df6a664ef6d9/raw/75e3feb46a9cbe081750bb4d969c027a8fe7caa2/log.txt)

Source/Details/ASMainSerialQueue.mm
```diff
@@ -40,14 +40,12 @@ - (NSUInteger)numberOfScheduledBlocks
...
 - (void)performBlockOnMainThread:(dispatch_block_t)block
 {
-
-  AS::UniqueLock l(_serialQueueLock);
-  [_blocks addObject:block];
   {
-    l.unlock();
-    [self runBlocks];
-    l.lock();
+    AS::UniqueLock l(_serialQueueLock);
```
maicki: You don't need a `AS::UniqueLock` here anymore. Let's just use `AS::MutexLocker`

nguyenhuy: Done.

PR #1410, opened 2019-03-15
Fix dealloc on bg

wsdwsd0829:
The node can be deallocated on background and causing crash when it try to access view/gesture recognizers in dealloc.
ASMainThreadDeallocation.h should handle the ivar deallocation, thus we do not need to do it ourselves here


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/cb7f184ff790569c97145fce9757f9c5/raw/071b4ac788b56b923fc352baa6b47340ce4ea7d2/log.txt)

PR #1413, opened 2019-03-19
Add support for reverse direction to yoga layouts

wiseoldduck:


Source/Layout/ASStackLayoutDefines.h
```diff
@@ -15,6 +15,10 @@ typedef NS_ENUM(NSUInteger, ASStackLayoutDirection) {
   ASStackLayoutDirectionVertical,
   /** Children are stacked horizontally */
   ASStackLayoutDirectionHorizontal,
+  /** Children are stacked vertically, but in reverse. Only used by Yoga spec. */
```
maicki: We may should compile this part out for now via the `Yoga` flag. Same in `ASStackLayoutSpec.mm` where we use the flags.

wiseoldduck: You mean where we just assert on them? lol

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/78c8d11e95be0eefcd17ff5f02348532/raw/e19ec7c0c5aedc051e50ba80f026babef812957a/log.txt)

PR #882, opened 2018-04-12
Match interfacestate update sequence to uikit

wsdwsd0829:

PR #1418, opened 2019-03-22
Fix a bug in ASRangeController that causes cell nodes to render unnecessarily

nguyenhuy:
When a visible collection/table view is about to become invisible, we tell its range controller to update (see https://github.com/TextureGroup/Texture/pull/882). However, [an optimization in ASRangeController](https://github.com/TextureGroup/Texture/commit/90a1bb23469a38820adc681fca4aac23ea5c77ed) doesn't expect that. Instead, it assumes that the view is **invisible** and going to be **visible** soon. As a result, it tells cell nodes in *all* ranges to start preloading and rendering, including nodes that weren't even in the display range. To react to that, those nodes [put themselves into the rendering queue](https://github.com/TextureGroup/Texture/blob/master/Source/ASDisplayNode.mm#L2955) which causes them to load their back store (view/layer).

Since ASRangeController is such a fundamental class, I wrapped this fix in an experiment in case I didn't cover an edge case, and to measure the impact on memory consumption in our app.

Here is what happens in control group - without the fix:
Screen Shot 2019-03-21 at 5 27 20 PM
Screen Shot 2019-03-21 at 5 27 56 PM
Screen Shot 2019-03-21 at 5 28 09 PM
Screen Shot 2019-03-21 at 5 29 06 PM

Here is enabled group - with the fix:
Screen Shot 2019-03-21 at 5 31 29 PM
Screen Shot 2019-03-21 at 5 32 05 PM

### DIFF
--- a/Schemas/configuration.json
+++ b/Schemas/configuration.json
@@ -25,7 +25,8 @@
                     "exp_disable_a11y_cache",
                     "exp_dispatch_apply",
                     "exp_image_downloader_priority",
-                    "exp_text_drawing"
+                    "exp_text_drawing",
+                    "exp_fix_range_controller"
                 ]
     		}
 		}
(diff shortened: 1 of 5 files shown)


nguyenhuy:
Yes, unit test will be added in a follow up diff - hoping to get this in before the next branch cut.

Adlai-Holler:
I hear that – gl

nguyenhuy:
Thanks for reviewing!

PR #1419, opened 2019-03-22
Follow up on the ASRangeController fix in #1418

nguyenhuy:
The changes in #1418 is a bit too aggressive when it comes to nodes that are in display range. It forces those nodes to not preload.

Also update the changes to avoid diluting the experiment data by triggering too broadly (i.e avoid triggering when the old and new implementations yield the same result leading to no behavior change).


nguyenhuy:
For reviewers: might be easier to review the end state of this diff with the parent commit of #1418: https://github.com/TextureGroup/Texture/compare/74de765b62ef988bc773f0b08f9e705da579a97e...nguyenhuy:HN-Fix-RangeController?expand=1

PR #1424, opened 2019-03-25
[Experiment] Dont forget about these collection view background deallocs

mikezucc:
Follow up to the #1420 where collection node is sending data/range controller to this API. Removing these calls removes child node leaks in removing collection nodes


mikezucc:
Note: not experimenting for the ASImageNode and friends since they do not intercept the `-dealloc` flow

PR #1425, opened 2019-03-26
Replace +load initializers with __attribute__((constructor)) functions

farktronix:
With the latest version of Xcode installed I was getting this dyld error at app launch time:

```Swift class extensions and categories on Swift classes are not allowed to have +load methods```

It looks like it's possible to replace `+load` methods with a static `__attribute__((constructor))` function without making dyld angry.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -284,7 +284,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 }
 
 #if !AS_INITIALIZE_FRAMEWORK_MANUALLY
-+ (void)load
+__attribute__((constructor)) static void ASLoadFrameworkInitializer(void)
 {
   ASInitializeFrameworkMainThread();
 }
--- a/Source/Private/ASTipsController.mm
+++ b/Source/Private/ASTipsController.mm
@@ -37,9 +37,9 @@
 
 #pragma mark - Singleton
 
-+ (void)load
+__attribute__((constructor)) static void ASLoadTipsControllerNotification(void)
 {
-  [NSNotificationCenter.defaultCenter addObserver:self.shared
+  [NSNotificationCenter.defaultCenter addObserver:ASTipsController.shared
                                          selector:@selector(windowDidBecomeVisibleWithNotification:)
                                              name:UIWindowDidBecomeVisibleNotification
                                            object:nil];
--- a/Source/ASConfigurationDelegate.h
+++ b/Source/ASConfigurationDelegate.h
@@ -27,8 +27,8 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Texture framework initialized. This method is called synchronously
  * on the main thread from ASInitializeFrameworkMainThread if you defined
- * AS_INITIALIZE_FRAMEWORK_MANUALLY or from the default initialization point
- * (currently +load) otherwise.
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: Source/ASDisplayNode.mm [hunk 1]
```diff
@@ -284,7 +284,7 @@ + (void)initialize
 }
 
 #if !AS_INITIALIZE_FRAMEWORK_MANUALLY
-+ (void)load
+__attribute__((constructor)) static void ASLoadFrameworkInitializer(void)
 {
   ASInitializeFrameworkMainThread();
```
nguyenhuy: Agree that using static attribute initializer is the best option here.

We should update the documentation of `-[ASConfigurationDelegate textureDidInitialize]` to state that it's called pre-main instead of +load.

@Adlai-Holler @maicki Do you know if `-textureDidInitialize` is useful for any clients? If it can be removed, and if we can shut down `ASExperimentalLayerDefaults` experiment (i.e gather `allowsGroupOpacityFromUIKitOrNil` and `allowsEdgeAntialiasingFromUIKitOrNil` on demand), then we can remove `ASInitializeFrameworkMainThread()` completely.

Source/Private/ASTipsController.mm
```diff
@@ -37,9 +37,9 @@ @implementation ASTipsController
 
 #pragma mark - Singleton
 
-+ (void)load
```
nguyenhuy: While we're on this, can we wrap this initializer in the `AS_ENABLE_TIPS` compiler flag and get back a tiny bit of pre-main time?

Source/ASDisplayNode.mm
farktronix: Thanks for catching that, @nguyenhuy! I noticed that comment when I was debugging this but forgot to update it with this patch. I'll push an update in a minute.

Source/Private/ASTipsController.mm
farktronix: All of the code in this file is wrapped in `#if AS_ENABLE_TIPS` already, so I think we're good here.

farktronix:
One thing about this that concerned me is that the [clang attributes documentation](https://clang.llvm.org/docs/AttributeReference.html) doesn't mention the `constructor` attribute anywhere. 
It exists in the [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc-4.3.0/gcc/Function-Attributes.html#index-g_t_0040code_007bconstructor_007d-function-attribute-2122)

However, clang has [tests](https://github.com/llvm-mirror/clang/blob/master/test/Sema/constructor-attribute.c) for using `constructor` and `destructor` attributes, so I think it's safe to depend on this.

maicki:
@farktronix Should be fine to use it. Apple even mentions it within it's WWDC talks about startup optimization.

Source/ASDisplayNode.mm
nguyenhuy: Awesome, thanks!

Source/Private/ASTipsController.mm
nguyenhuy: 👍 Good call. I missed that.

Source/ASDisplayNode.mm
Adlai-Holler: We do rely on `-textureDidInitialize` currently so I don't want to get rid of it right now but let me think of how we can remove our reliance on it.

PR #1432, opened 2019-03-29
Have image nodes draw into opaque contexts automatically if possible

Adlai-Holler:
If the image is opaque, and it fills the context, the context can be opaque regardless of whether they set the flag on the node.


Adlai-Holler:
I like this diff, but as I bet Pinterest realized, it doesn't play well with image/context modification blocks that expect a transparent context. Revert for now.

PR #1434, opened 2019-03-29
[ASDisplayNode] Stop infinite layout in _u_measureNodeWithBoundsIfNecessary

rcancro:
We came across an infinite layout loop in `_u_measureNodeWithBoundsIfNecessary`. After requesting a layout from above, the sizes between pending and caluclated layout still do not match. We continue to prefer to use the pending layout and ask for another layout loop from above. We can’t seem to break out of this loop. The solution (thanks to Huy for the guidance) was to nil out the pending layout we get from requesting the layout from above.

I was only able to reproduce this when working with a node in a `UINavigationBarItem’s` `titleView`. I think that UIKit must be doing something sneaky with setting the frame on the view. While I was not able to create a unit test to catch this issue (I tried for a long time, and can post what I’ve come up with to see if anyone has any suggestions), I was able to create a pretty simple example project that shows the behavior:
https://github.com/rcancro/TextureLayoutLoopExample

### DIFF
--- a/Source/ASDisplayNode+Layout.mm
+++ b/Source/ASDisplayNode+Layout.mm
@@ -444,6 +444,16 @@ ASLayoutElementStyleExtensibilityForwarding
       __instanceLock__.lock();
     }
 
+    // If we request that our root layout we may generate a new _pendingDisplayNodeLayout.layout which has
+    // requestedLayoutFromAbove set to NO. If the pending layout has a different constrained size than nextLayout's
+    // and the layout sizes don't change we could end up back here asking the root to layout again causing an
+    // infinite layout loop. Instead, we nil out the _pendingDisplayNodeLayout.layout here because it can be
+    // considered an undesired artifact of the layout request. nextLayout will become _calculatedDisplayNodeLayout
+    // when the pending layout transition which will be created later in this method is applied.
+    // We will use _calculatedLayout the next time around, so requestedLayoutFromAbove will be set to YES and we
+    // will break out of this layout loop.
+    _pendingDisplayNodeLayout.layout = nil;
+    
     // Update the layout's version here because _u_setNeedsLayoutFromAbove calls __setNeedsLayout which in turn increases _layoutVersion
     // Failing to do this will cause the layout to be invalid immediately
     nextLayout.version = _layoutVersion;


nguyenhuy:
> The comment below says _u_setNeedsLayoutFromAbove bumps the _layoutVersion of the node; that should mark the _pendingDisplayNodeLayout as totally useless which is the same as doing this. 

`_pendingDisplayNodeLayout` actually has the increased layout version because it's a bi-product of the layout pass performed during `-_u_setNeedsLayoutFromAbove`. Then we update the layout version of `nextLayout` to the same value and set it to `_calculatedDisplayNodeLayout` when `-_completePendingLayoutTransition` is called (either at the end of `-_u_measureNodeWithBoundsIfNecessary` or when the parent node finishes an on-going layout transition). And then this happens on the next layout pass:

Inf layout

The node picks up the bi-product (`_pendingDisplayNodeLayout`) because it has the same version and a different constrained size, only to later realizes that it can't use it and calls `_u_setNeedsLayoutFromAbove` again.

> There must be some place (maybe _locked_constrainedSizeForLayoutPass?) where we're not validating the version.

It's true that `_locked_constrainedSizeForLayoutPass` doesn't check the layout version, and I think it should. It wouldn't help in this case, see the screenshot below:

Inf layout 1

Source/ASDisplayNode+Layout.mm
```diff
@@ -444,6 +444,16 @@ - (void)_u_measureNodeWithBoundsIfNecessary:(CGRect)bounds
       __instanceLock__.lock();
     }
 
+    // If we request that our root layout we may generate a new _pendingDisplayNodeLayout.layout which has
+    // requestedLayoutFromAbove set to NO. If the pending layout has a different constrained size than nextLayout's
+    // and the layout sizes don't change we could end up back here asking the root to layout again causing an
+    // infinite layout loop. Instead, we nil out the _pendingDisplayNodeLayout.layout here because it can be
+    // considered an undesired artifact of the layout request. nextLayout will become _calculatedDisplayNodeLayout
+    // when the pending layout transition which will be created later in this method is applied.
+    // We will use _calculatedLayout the next time around, so requestedLayoutFromAbove will be set to YES and we
```
nguyenhuy:
```suggestion
    // We will use _calculatedDisplayNodeLayout the next time around, so _u_setNeedsLayoutFromAbove will be set to YES and we
```

nguyenhuy:
Let’s get this in. Thanks so much everyone.

PR #1441, opened 2019-04-03
Fix the stub for layerActionForKey: to let CA continue the search

Adlai-Holler:
As the comment says, we return kCFNull from +defaultActionForKey:. Stubbing this to nil is important so that CA will search the actions and style.actions dictionaries for layer actions.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/eb909bfedc903266c4a35a116575fd0a/raw/8cdb9eac22a3fe866e35cd134e0068cb5fae4764/log.txt)

Adlai-Holler:
Test failure looks flaky.

PR #1443, opened 2019-04-04
Revert "Have image nodes draw into opaque contexts automatically if possible"

Adlai-Holler:
Reverts TextureGroup/Texture#1432

I like this diff, but as I bet Pinterest realized, it doesn't play well with image/context modification blocks that expect a transparent context. Revert for now.


wiseoldduck:
😕

nguyenhuy:
Thanks for reverting it!

PR #1455, opened 2019-04-15
Try to remove global lock when initialising TextKit components

zhongwuzw:
This is a PR, but I want to come up this discussion, is this really need? if needed, we need to create function to lock when initialize TextKit components, because we have others places to do initialize things in Texture, see https://github.com/TextureGroup/Texture/search?q=%5B%5BNSTextStorage+alloc%5D&unscoped_q=%5B%5BNSTextStorage+alloc%5D.

Seems this lock code is from `ComponentKit`, but I can't find any infos based rdar.

From Apple docs, I only find they should access in a single thread at the same time.
https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/CustomTextProcessing/CustomTextProcessing.html#//apple_ref/doc/uid/TP40009542-CH4-SW1
image


Adlai-Holler:
There's an adage "never tear down a fence until you know why it was built."

If you're willing, I think this would be a good place for an experiment. My company doesn't use TextKitComponents right now, but if you or Pinterest has a large enough user base to put this under experiment and watch for an increase in crashes before rolling this out to everyone, I think that's the way to go. cc @nguyenhuy

zhongwuzw:
@Adlai-Holler Thanks for review, sorry I don't have enough users to experiment, my opinion is it may already prove not produces much crashes, because we also do some initializing things at other places, so at least only add lock at here is not sufficient, I think we need to add a global lock to all places we used? or try to experiment to remove this lock? 🤔

Adlai-Holler:
Let's use an experiment to remove this lock, unless stakeholders at large companies that use TextNode1 like @nguyenhuy or @ay8s feel differently. It should be quick to converge after only a few days of experimentation at scale, to see if there's an appreciable difference in crashes.

Schemas/configuration.json
```diff
@@ -27,7 +27,9 @@
                     "exp_image_downloader_priority",
                     "exp_text_drawing",
                     "exp_fix_range_controller",
-                    "exp_oom_bg_dealloc_disable"
+                    "exp_oom_bg_dealloc_disable",
+                    "exp_transaction_operation_retain_cycle",
+                    "exp_adding_lock_to_textkit_initialising"
```
Adlai-Holler: The experiment should remove the lock so that current behavior is maintained ("opt-in".)

Source/Layout/ASAbsoluteLayoutSpec.h
```diff
@@ -13,7 +13,7 @@
 typedef NS_ENUM(NSInteger, ASAbsoluteLayoutSpecSizing) {
   /** The spec will take up the maximum size possible. */
   ASAbsoluteLayoutSpecSizingDefault,
-  /** Computes a size for the spec that is the union of all childrens' frames. */
+  /** Computes a size for the spec that is the union of all children's frames. */
```
Adlai-Holler: 👍

Source/TextKit/ASTextKitContext.mm
```diff
@@ -34,14 +34,15 @@ - (instancetype)initWithAttributedString:(NSAttributedString *)attributedString
...
-    AS::MutexLocker l(*mutex);
+    static AS::Mutex *mutex = NULL;
+    if (mutex != NULL || ASActivateExperimentalFeature(ASExperimentalAddingLockToTextKitInitialising)) {
+      static dispatch_once_t onceToken;
+      // Concurrently initialising TextKit components crashes (rdar://18448377) so we use a global lock.
+      dispatch_once(&onceToken, ^{
+        mutex = new AS::Mutex();
+      });
+      mutex->lock();
+    }
```
Adlai-Holler: Move all the logic inside the dispatch_once.

zhongwuzw:
cc. @Adlai-Holler

Source/TextKit/ASTextKitContext.mm
```diff
@@ -34,14 +34,17 @@ - (instancetype)initWithAttributedString:(NSAttributedString *)attributedString
...
     dispatch_once(&onceToken, ^{
-      mutex = new AS::Mutex();
+      if (!ASActivateExperimentalFeature(ASExperimentalRemoveTextKitInitialisingLock)) {
+        mutex = new AS::Mutex();
+      }
     });
-    
-    // Concurrently initialising TextKit components crashes (rdar://18448377) so we use a global lock.
-    AS::MutexLocker l(*mutex);
+    if (mutex) {
```
nguyenhuy: Nit: Let's use `mutex != NULL` to be consistent with the check at the end of this scope.

appleguy:
FYI, I'm fairly sure this is *not* safe to remove on iOS 10 / iOS 9, and possibly newer versions. We should proceed with caution. There used to be crashes in CoreText primitives when initialized concurrently, especially Framesetter.

zhongwuzw:
@appleguy Thanks for your valuable infos. If we can't initialized concurrently, I think we also need do safe guard at some [other places](https://github.com/TextureGroup/Texture/search?q=%5B%5BNSTextStorage+alloc%5D&unscoped_q=%5B%5BNSTextStorage+alloc%5D )? And I also think maybe we only can initialize them on main thread, because we can't ensure users may also use UIKit UILabel/UITextView... directly when they use Texture at the same time.

bolsinga:
We just wrote a test that creates 2000 threads and this works on iOS10 in the simulator. We need to find an iOS9 to test some more.

nguyenhuy:
Actually, we ran this experiment internally and the lock does indeed help to prevent a rather serious crash. At one point it was our top crasher (BUG-97509). It also occurs on iOS 13 so we likely need to keep the global lock around for some time. I think the next step is to actually keep the global lock and remove the experiment code.

Here is a sample crash report if anyone is interested in:
```
# OS Version: 13.1.2 (17A860)

#10. Crashed: org.AsyncDisplayKit.ASDisplayLayer.displayQueue
0  libsystem_malloc.dylib         0x191be7e40 nanov2_allocate_from_block$VARIANT$mp.cold.1 + 36
1  libsystem_malloc.dylib         0x191bced14 nanov2_find_block_and_allocate$VARIANT$mp + 510
2  libsystem_malloc.dylib         0x191bcdf98 nanov2_allocate$VARIANT$mp + 124
3  libsystem_malloc.dylib         0x191bce480 nanov2_calloc$VARIANT$mp + 136
4  libsystem_malloc.dylib         0x191bdc924 malloc_zone_calloc + 140
5  libsystem_malloc.dylib         0x191bdd208 calloc + 32
6  UIFoundation                   0x1954756c8 _NSGlyphTreeInsertNodeAfterCurrentNode + 416
7  UIFoundation                   0x1954764bc _NSGlyphTreeInsertGlyphs + 164
8  UIFoundation                   0x195446d4c -[NSLayoutManager setGlyphs:properties:characterIndexes:font:forGlyphRange:] + 200
9  UIFoundation                   0x195437210 -[NSGlyphGenerator generateGlyphsForGlyphStorage:desiredNumberOfCharacters:glyphIndex:characterIndex:] + 3564
10 UIFoundation                   0x195437570 -[NSLayoutManager(NSPrivate) _fillGlyphHoleForCharacterRange:startGlyphIndex:desiredNumberOfCharacters:] + 700
11 UIFoundation                   0x195479cdc _NSFastFillAllGlyphHolesForCharacterRange + 900
12 UIFoundation                   0x195437aa8 -[NSLayoutManager glyphRangeForCharacterRange:actualCharacterRange:] + 56
13 UIFoundation                   0x19542615c -[NSLayoutManager(NSPrivate) _fillLayoutHoleForCharacterRange:desiredNumberOfLines:isSoft:] + 700
14 UIFoundation                   0x19542abb4 -[NSLayoutManager(NSPrivate) _fillLayoutHoleAtIndex:desiredNumberOfLines:] + 196
15 UIFoundation                   0x195432640 -[NSLayoutManager(NSPrivate) _markSelfAsDirtyForBackgroundLayout:] + 424
16 UIFoundation                   0x19542d1d8 -[NSLayoutManager(NSPrivate) _invalidateLayoutForExtendedCharacterRange:isSoft:invalidateUsage:] + 2332
17 UIFoundation                   0x19542d46c -[NSLayoutManager textStorage:edited:range:changeInLength:invalidatedRange:] + 212
18 UIFoundation                   0x19542d318 -[NSTextStorage _notifyEdited:range:changeInLength:invalidatedRange:] + 184
19 UIFoundation                   0x195433c0c -[NSTextStorage edited:range:changeInLength:] + 320
20 Foundation                     0x1921f13b0 -[NSConcreteMutableAttributedString replaceCharactersInRange:withAttributedString:] + 352
21 UIFoundation                   0x1954b4b98 __71-[NSConcreteTextStorage replaceCharactersInRange:withAttributedString:]_block_invoke + 44
22 UIFoundation                   0x1954b47c8 __NSConcreteTextStorageLockedForwarding + 56
23 UIFoundation                   0x1954397dc -[NSConcreteTextStorage replaceCharactersInRange:withAttributedString:] + 72
24 Pinterest                      0x1018b1b44 -[ASTextKitContext initWithAttributedString:lineBreakMode:maximumNumberOfLines:exclusionPaths:constrainedSize:] + 61 (ASTextKitContext.mm:61)
25 Pinterest                      0x1018b6b28 -[ASTextKitRenderer initWithTextKitAttributes:constrainedSize:] + 63 (ASTextKitRenderer.mm:63)
26 Pinterest                      0x10181b18c rendererForAttributes(ASTextKitAttributes, CGSize) + 133 (ASTextNode.mm:133)
27 Pinterest                      0x10181b0b4 -[ASTextNodeDrawParameter rendererForBounds:] + 178 (ASTextNode.mm:178)
28 Pinterest                      0x10181d0d8 +[ASTextNode displayWithParameters:isCancelled:] + 566 (ASTextNode.mm:566)
```

vovasty:
shouldn't `performBlockWithLockedTextKitComponents` be also called within the lock, cause it may change text storage?

rahul-malik:
@nguyenhuy - I noticed that experiment ended about a month before iOS 13 GM was available. Is it worth testing considering the amount of issues in this area suddenly spiked starting mid-late September which lines up pretty well with iOS 13 adoption?

nguyenhuy:
ah, good point on the timing of the experiment and iOS 13 GM. Yeah I guess it's worth to try again.

PR #1457, opened 2019-04-16
Make ASCollectionNode's pan gesture customizable

OskarZhang:
Adding the ability for ASCollectionNode’s subclasses to customize the behavior of its default panGestureRecognizer.

With texture and UIKit, you can't assign a collection's default pan gesture delegate to an instance other than the collectionView itself, so for code like this `self.pagerNode.view.panGestureRecognizer.delegate = self;`, UIKit would raise the following exception:
`Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'UIScrollView's built-in pan gesture recognizer must have its scroll view as its delegate.'`

I don't see any other way to do this since Texture's collectionNode wraps its underlying view and users of this class would not be able to subclass the view.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -1040,6 +1040,28 @@ ASLayoutElementCollectionTableSetTraitCollection(_environmentStateLock)
   return result;
 }
 
+#pragma mark - UIGestureRecognizerDelegate Methods
+
+- (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
+{
+    return YES;
+}
+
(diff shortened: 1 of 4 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d3949444ebdd9ac2c284244bb777b594/raw/9bb1e6a81cbe73aa75db0dec8fe6e75a2066452b/log.txt)

Source/ASCollectionNode.mm
```diff
@@ -1040,6 +1040,38 @@ - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode;
   return result;
 }
 
+#pragma mark - UIGestureRecognizerDelegate Methods
+// The value returned below are default implementation of UIKit's UIGestureRecognizerDelegate
```
garrettmoon: Are we sure these are the default implementations of UICollectionView? Could we do some trickery to get UICollectionView's implementation here instead in case they change in the future?

PR #1433, opened 2019-03-29
Broaden UIGraphicsRenderer Experiment to Everywhere

Adlai-Holler:
This gets rid of the ASExperimentalGraphicsContext experiment that was a dud, and replaces ASGraphicsContexts.h with a wrapper. The wrapper calls into either UIGraphicsBeginImageContext or UIGraphicsImageRenderer to build an image.

These graphics renderers seem pretty awesome. https://gist.github.com/NSProgrammer/38f1b4f9b29c3077af3f39a7e98092b1

Want multiple core-team reviews on this since it touches such core code. It's pretty straightforward though. The one case where it wasn't straightforward (`__didDisplay…`) I just left with UIGraphics.

PR #1458, opened 2019-04-16
Shut down graphics contexts experiment

Adlai-Holler:
This is basically the first half of #1433. We'll replace it with a new experiment that uses UIGraphicsRenderer.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d48c964d1cf495561b39b2e1fb03778c/raw/8ad9efb61482b35752b2e7317f8877e38ea99544/log.txt)

PR #1461, opened 2019-04-19
Refactor signpost code to be safer, more debuggable

Adlai-Holler:
Now that we're all-C++, we can use `static constexpr bool` like this to still verify that something _would_ build if the bool were either way, but still strip code that is unreachable.

This fixes warnings about availability when profiling, and it makes the profiling code more readable & syntax highlighted, and it makes sure that you can't break profiling builds inadvertently.

Confirmed that even in debug mode, unreachable code is totally stripped and the entire `ASSignpost*` function call is not even present at all in the disassembly unless PROFILE=1.

Over time I'd like us to move away from macros when it makes sense.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b4ca9bcbbb15a27fe0182a9da1f05499/raw/7abd2a65a5200194b1994256c7289950b62dfbb3/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/8c7362b47ebc34a653fdd79ebb32b6cf/raw/1a040f9aab60430e17a0ce59e6102a3aba7f8b43/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e6543e8617d67a9ff1bf59f25ace8e29/raw/3e85cceeecf1df1f5d14b95cb7ee66e3d965ca18/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/59b41803c8d16bc55a6e970482aec95a/raw/ec4072f30d2275a4940cb5e827514b414a968785/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/24f41044fa02545ebcb245ed2ba57112/raw/264e3369687a1f1b2f3ca4a0bc6a8eb62f94343d/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/58de7b4a8ccf3be52f06aa7d0d9a42ca/raw/5eea4ddc5ac91175fda652879a25aa3eddef6aac/log.txt)

Adlai-Holler:
@maicki @nguyenhuy @garrettmoon Review plz?

Adlai-Holler:
I'm proud of this but I just learned about the new and incredible os_signpost API, which will _require_ the use of macros, and which has such a different API that everything will need to change.

Closing this in favor of a diff that I hope comes soon that adds support for that, probably keeping the macros.

PR #1463, opened 2019-04-22
Drop support for Xcode 8 and fix some compiler warnings

Adlai-Holler:
Saw these come up when debugging a compiler error on #1461


Kaspik:
@Adlai-Holler How this PR could land if `.travis.yml` for CI is using XCode 8.1? Shouldn't the CI process be updated with this together? (I'm updating it so asking what should be default now and what updated)

PR #1467, opened 2019-04-25
CALayer: Set the default contentsScale to screen scale, update to match contents

Adlai-Holler:
Currently we hack this into ASImageNode but really it applies to any node.

Without this, if you're on a 2.0 screen and you set the content mode of a text to, say, `TopLeft`, the text will be twice its size i.e. rendered at 1.0, and you can only see the top left quarter of the contents.

Basically, we are currently relying on `UIViewContentModeScaleToFill/kCAGravityResize` (the default) to apply our contentsScale. This is wrong and bad.

UIView sets contents scale based on the screen, we should do the same thing on our own.

- Change the default to main screen scale.
- When a contents image is applied to the layer, match its contentsScale to the contents image.
  - Old behavior was to set the node's contentsScale to the layer's contentScale but this is nonsensical, since the node's contentsScale is a bridged property from the layer! It was a no-op.

This is by no means the perfect state (we will still trample on a user-set value of contentsScale, but that's insanely rare) but it's better, and it's similar to how UIKit handles this.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/7c42f3cc917c4cfe4d369b5d7807e18d/raw/52d8be026f87f9f04aeba92b987619fb06d965ec/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4f12e5b4b137612f66547cd5d482eef0/raw/2547335e57b63e2e7b3672735657d8a6556cdc94/log.txt)

nguyenhuy:
@Adlai-Holler Can you rebase with master to unblock the CI?

Adlai-Holler:
Sure, done

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e112301760c33703ae0ff6c1de22eae6/raw/e1c67d4d5a2fe8bc473984bfff1b54be4d2f3253/log.txt)

PR #1468, opened 2019-04-25
Fix open context bug introduced in #1458

Adlai-Holler:
iOS attempts to automatically close these when the thread exits, but some versions (at least iOS 10, checking others) have a bug which causes it to just keep spinning indefinitely.

cc @garrettmoon @nguyenhuy @ay8s Everyone should cherry-pick this into any branches that include #1458.


nguyenhuy:
Will pick this. Thanks!

PR #1470, opened 2019-04-27
Make ASImageNodeRoundBorderModificationBlock more flexible

grangej:
As it stands the current ASImageNodeRoundBorderModificationBlock has two issues:

1. Will produce a border half the desired size
2. Will produce a border that is visibly different size depending on original size of image

This PR seeks to address this by allowing the caller to specify the desired output size (likely matching the final desired node size), the block will then draw in a rect modified to match the desired size. In addition it will double the size of the border to be that of the desired border width. This will more closely match what you might get when using a CALayer border.

### DIFF
--- a/Source/ASImageNode.h
+++ b/Source/ASImageNode.h
@@ -189,6 +189,7 @@ typedef UIImage * _Nullable (^asimagenode_modification_block_t)(UIImage *image);
 /**
  * @abstract Image modification block that rounds (and optionally adds a border to) an image.
  *
+ * @param size The desired final size for the image.
  * @param borderWidth The width of the round border to draw, or zero if no border is desired.
  * @param borderColor What colour border to draw.
  *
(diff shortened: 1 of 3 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/153b87a07ef78bf9dcf5d7b716cc1622/raw/d3813abf89d92974d2c3c28347ecaaea8ea93cea/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/c5d4472c8f1bdb44fe5a13eb81b5a4fe/raw/2547335e57b63e2e7b3672735657d8a6556cdc94/log.txt)

nguyenhuy:
@grangej Can you rebase with master to fix the CI build? Thanks.

PR #1473, opened 2019-04-30
NetworkingImageNode: update URL if we found image in assets

zhongwuzw:
If we find image in bundle assets, we also need to update `URL` to try to load animated image.


zhongwuzw:
@Adlai-Holler Done :)

PR #1475, opened 2019-04-30
Make sure all ASDisplayNode properties have backing ivars for consistency.

bolsinga:
Found this by enabling `#pragma clang diagnostic error "-Wobjc-missing-property-synthesis"` for `ASDisplayNode`. One property is unused, saving 8 bytes of heap space per instance on 64-bit builds. Implement setter/getters for these properties, and add appropriate locking. add the warning as error to the build for this file.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -2635,6 +2635,36 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
   return NO;
 }
 
+- (BOOL)placeholderEnabled
+{
+    return _placeholderEnabled;
+}
+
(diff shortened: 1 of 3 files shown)


Source/ASDisplayNode.mm
```diff
@@ -2635,6 +2635,36 @@ - (BOOL)placeholderShouldPersist
...
+}
+
+- (NSInteger)drawingPriority
+{
+    return _drawingPriority;
+}
+
+- (void)setDrawingPriority:(NSInteger)drawingPriority
+{
+    _drawingPriority = drawingPriority;
```
bolsinga: Discussed with @garrettmoon @nguyenhuy and these properties do need locks. I'll also see if I can always enable this warning as error in this class. This way we can get these warnings enabled, and fixed 1 by 1 until they are able to be be enabled for the entire project.

PR #1476, opened 2019-04-30
Ensure ASControlMode properties lock before accessing their ivars

bolsinga:
Enable `#pragma clang diagnostic error "-Wobjc-missing-property-synthesis"` for this file.


wiseoldduck:
What happened here, this is similar to PR #1479 so refer to the discussion there: The synthesized atomic getters/setters are adequate and superior here. I believe we should revert this.

PR #1479, opened 2019-05-01
Ensure ASCellNode properties lock before accessing their ivars

bolsinga:
Enable `#pragma clang diagnostic error "-Wobjc-missing-property-synthesis"` for this file.

### DIFF
--- a/Source/ASCellNode.h
+++ b/Source/ASCellNode.h
@@ -9,6 +9,9 @@
 
 #import <AsyncDisplayKit/ASDisplayNode.h>
 
+#pragma clang diagnostic push
+#pragma clang diagnostic error "-Wobjc-missing-property-synthesis"
+
 NS_ASSUME_NONNULL_BEGIN
 
 @class ASCellNode, ASTextNode;
@@ -257,3 +260,5 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 @end
 
 NS_ASSUME_NONNULL_END
+
+#pragma clang diagnostic pop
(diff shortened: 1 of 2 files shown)


bolsinga:
Well, AFAIK atomic means the value is fully retrieved or set, not thread safe. https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html

I'm finding these while doing some in-progress work shrinking the instance size of the nodes in the heap. To do this I have various properties (mostly BOOLs and enums) access ivars instead of padded synthesized properties. On average in the Pinterest app, I've reduced the size of various nodes by 8.5%. Since nodes are long lived, this is a good win.

wiseoldduck:
While it's generally true that more than `atomic` getters and setters are required to achieve thread safety, in this case, getters and setters is all there is!

So, here, the synthesized methods should be sufficient (better, really, as Adlai mentioned).

bolsinga:
I don't understand. This object already uses the lock on other properties; why wouldn't it need them on all of them? `atomic` alone will not solve the problem that the lock solves.

Adlai-Holler:
I am 100% behind reducing instance sizes for our classes. Let’s do it by any safe means.

To the atomic vs locked question, if I set a property without any of my own external synchronization, and setting that property has no side effect (a side effect like setNeedsDisplay), then atomic is all I need because I dont stand to gain anything by taking the lock. I have introduced a race condition from the outside – I must be happy with my property-set either being reflected in any concurrent work, or not being reflected in it. Blocking my thread until the in-progress work is completed, or blocking the coincidental work until the property set is completed, benefits nobody.

bolsinga:
OK. Do you have improvements proposed for this diff, or shall I abandon it?

I still don't understand the locking philosophy, as the present pattern disagrees with the documentation.

bolsinga:
Abandoning. Will work to clean this up with guidance from @nguyenhuy

bolsinga:
See #1498

PR #1482, opened 2019-05-03
Don't clear contents when exit display state be called for animated image node

zhongwuzw:
Seems `kCAOnOrderOut` is be called not only when layer removing or hidden, it also can be called when layer is going to be visible. 

Reproducible steps:
1: Set `ASNetworkImageNode`'s url to a GIF image.
2: Run `AsyncDisplayKitOverview` example project.
3: Click into `ASNetworkImageNode` example, we can see layer's `contents` would be cleared after gif played some frames.

### DIFF
--- a/Source/ASImageNode+AnimatedImage.mm
+++ b/Source/ASImageNode+AnimatedImage.mm
@@ -315,18 +315,7 @@
 #if ASAnimatedImageDebug
     NSLog(@"exiting display state: %p", self);
 #endif
-    
-  // Check to see if we're an animated image before calling super in case someone
-  // decides they want to clear out the animatedImage itself on exiting the display
-  // state
-  BOOL isAnimatedImage = self.animatedImage != nil;
   [super didExitDisplayState];
-  
-  // Also clear out the contents we've set to be good citizens, we'll put it back in when we become visible.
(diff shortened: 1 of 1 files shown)


zhongwuzw:
@garrettmoon Emm, yeah, it can retain one frame of gif, but I think the memory usage is the same as static image? we cleared all cached frames and stoped gif decoding.

nguyenhuy:
@garrettmoon knows much more than me regarding this, but in general I think we should reclaim the memory since GIFs take more memory than static images. Closing this PR for now, please feel free to reopen, maybe with profiling and memory graphs before and after this change. Thank you!

zhongwuzw:
@nguyenhuy Yeah, but I don't change the release of GIFs, we still release the cached GIF frames in https://github.com/zhongwuzw/Texture/blob/0849c5847b0fca010878d05b41c29221e5309285/Source/ASImageNode%2BAnimatedImage.mm#L309. And I only not clear the contents, it's the same as static image. I think for static image when using `ASImageNode`, we also not cleared the contents I think, @garrettmoon Am I right? :)

zhongwuzw:
We can see a distinct flash when we load GIF.

PR #1483, opened 2019-05-03
Shrink _ASPendingState from 824 to 808 bytes, reduction of ~2%

bolsinga:

Shrink ASPendingState `BOOL`s. Use `unsigned int` for these bitfields for clarity. Separate from `_flags` as that struct is compared to `0` for testing. The `_boolFlags` struct is just storing other `BOOL`s.

### DIFF
--- a/Source/Private/_ASPendingState.mm
+++ b/Source/Private/_ASPendingState.mm
@@ -17,7 +17,7 @@
 #import <AsyncDisplayKit/ASInternalHelpers.h>
 
 #define __shouldSetNeedsDisplay(layer) (flags.needsDisplay \
-  || (flags.setOpaque && opaque != (layer).opaque)\
+  || (flags.setOpaque && _boolFlags.opaque != (layer).opaque)\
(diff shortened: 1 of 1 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5f48926469d1440c238d5d95768f9ce3/raw/f3d36cb568cd2fd016778d77224688bcc79d86c9/log.txt)

Source/Private/_ASPendingState.mm
```diff
@@ -150,6 +143,24 @@ @implementation _ASPendingState
   NSDictionary<NSString *, id<CAAction>> *actions;
 
   ASPendingStateFlags _flags;
```
garrettmoon: Maybe we could rename this to something else and name _boolFlags _flags? Maybe this should be _stateToApplyFlags?

bolsinga: Good idea.

PR #1485, opened 2019-05-03
Shrink _ASCollectionPendingState from 144 to 128 bytes, a 12.5% reduction

bolsinga:

Group the `BOOL`s into a struct. Shrink the various stored `enum`s to fit the size of their contents. Move the ivars around so that the smaller `enum` are near each other and the bitfield struct.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -35,9 +35,10 @@
 @public
   std::vector<std::vector<ASRangeTuningParameters>> _tuningParameters;
 
+  // Keep these enums by the bitfield struct to save memory.
+  ASLayoutRangeMode _rangeMode;
+  ASCellLayoutMode _cellLayoutMode;
   struct {
-    ASLayoutRangeMode rangeMode;
-    ASCellLayoutMode cellLayoutMode;
     unsigned int allowsSelection:1; // default is YES
     unsigned int allowsMultipleSelection:1; // default is NO
     unsigned int inverted:1; //default is NO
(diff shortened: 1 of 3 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/301b6543dbe223c0b9ec031761de11a4/raw/b778ba1a673fbd1a20280d0331f98635d2a7e847/log.txt)

Source/ASCollectionNode.mm
```diff
@@ -34,6 +34,20 @@
...
+    unsigned int allowsSelection:1; // default is YES
+    unsigned int allowsMultipleSelection:1; // default is NO
+    unsigned int inverted:1; //default is NO
+    unsigned int alwaysBounceVertical:1;
+    unsigned int alwaysBounceHorizontal:1;
+    unsigned int animatesContentOffset:1;
+    unsigned int showsVerticalScrollIndicator:1;
+    unsigned int showsHorizontalScrollIndicator:1;
+    unsigned int pagingEnabled:1;
+  } _fields;
```
garrettmoon: Maybe flags instead of fields?

bolsinga: There are a couple of enums in there too, so I didn't want to use the name `_flags`. I'll see what happens to the size if I move them out of the struct but keep them next to this bitfield struct.

bolsinga: ok. new diff to come.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/82be691367595f28044d965b4e1092b6/raw/fd2914d2d9597f95c06e81e5e489df89af319551/log.txt)

PR #1487, opened 2019-05-03
Shrink ASImageNode by .6% and ASNetworkImageNode by 2.2%

bolsinga:
ASImageNode goes from 1384 to 1376. ASNetworkImageNode goes from 1496 to 1464.

Group the `BOOL`s into a struct. Shrink the various stored `enum`s to fit the size of their contents. Move the ivars around so that the smaller `enum` are near eachother and the bitfield struct.


Source/Private/ASImageNode+AnimatedImagePrivate.h
```diff
@@ -15,14 +15,19 @@
...
   CADisplayLink *_displayLink;
   NSUInteger _lastSuccessfulFrameIndex;
   
   //accessed on main thread only
   CFTimeInterval _playHead;
   NSUInteger _playedLoops;
+
+  struct {
+    unsigned int animatedImagePaused:1;
+    unsigned int cropEnabled:1; // Defaults to YES.
```
garrettmoon: So this is here not because it's related to animated images but then we can group these together to save memory? Do you mind adding a comment to that effect?

Source/ASNetworkImageNode.mm
```diff
@@ -42,41 +42,39 @@ @interface ASNetworkImageNode ()
...
-    unsigned int downloaderImplementsDownloadWithPriority:1;
-  } _downloaderFlags;
 
   // Immutable and set on init only. We don't need to lock in this case.
   __weak id<ASImageCacheProtocol> _cache;
   struct {
-    unsigned int cacheSupportsClearing:1;
-    unsigned int cacheSupportsSynchronousFetch:1;
-  } _cacheFlags;
+      unsigned int delegateWillStartDisplayAsynchronously:1;
```
garrettmoon: Can you add a comment above to indicate that keeping all of these flags in one struct reduces the nodes memory footprint?

Source/Private/ASImageNode+AnimatedImagePrivate.h
bolsinga: Yeah; it's due to this category declaring ivars. I'll add the comment.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/930bca59cc6708109ca49bc8c0071085/raw/bb82f73e4e04fd67e80dbe3c39313023b66ecd71/log.txt)

Source/ASImageNode.mm
```diff
@@ -301,8 +299,8 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
   drawParameters->_contentsScale = _contentsScaleForDisplay;
   drawParameters->_backgroundColor = self.backgroundColor;
   drawParameters->_contentMode = self.contentMode;
-  drawParameters->_cropEnabled = _cropEnabled;
-  drawParameters->_forceUpscaling = _forceUpscaling;
+  drawParameters->_cropEnabled = self.cropEnabled;
+  drawParameters->_forceUpscaling = self.forceUpscaling;
```
maicki: Why not using the `_imageNodeFlags` for both in here? We are already having the lock.

bolsinga: Aha! I see what you mean.

PR #1488, opened 2019-05-03
Support setting `accessibilityElements`

raycsh017:
Sometimes we want to override `accessibilityElements` so that we can
skip over elements unnecessary for interacting with the app using VoiceOver
or limit the interaction to child viewcontrollers presented, etc.

For this purpose, this commit adds support for setting custom
`accessibilityElements`. If not set, `accessibilityElements` should still
return the elements found from the view/layer hierarchy.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -3515,7 +3515,7 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
 - (void)setAccessibilityElementsBlock:(ASDisplayNodeAccessibilityElementsBlock)accessibilityElementsBlock
 {
   MutexLocker l(__instanceLock__);
-  _accessibilityElementsBlock = accessibilityElementsBlock;
+  _accessibilityElementsBlock = [accessibilityElementsBlock copy];
 }
 
 - (ASDisplayNodeAccessibilityElementsBlock)accessibilityElementsBlock
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -3512,6 +3512,18 @@ ASDISPLAYNODE_INLINE BOOL subtreeIsRasterized(ASDisplayNode *node) {
   return _flags.isAccessibilityContainer;
 }
 
+- (void)setAccessibilityElementsBlock:(ASDisplayNodeAccessibilityElementsBlock)accessibilityElementsBlock
+{
+  MutexLocker l(__instanceLock__);
+  _accessibilityElementsBlock = accessibilityElementsBlock;
+}
+
+- (ASDisplayNodeAccessibilityElementsBlock)accessibilityElementsBlock
+{
+  MutexLocker l(__instanceLock__);
+  return _accessibilityElementsBlock;
+}
+
 - (NSString *)defaultAccessibilityLabel
 {
   return nil;
--- a/Source/Details/_ASDisplayViewAccessiblity.mm
+++ b/Source/Details/_ASDisplayViewAccessiblity.mm
@@ -287,6 +287,12 @@ static void CollectAccessibilityElementsForView(UIView *view, NSMutableArray *el
   NSMutableArray *accessibilityElements = [[NSMutableArray alloc] init];
   CollectAccessibilityElementsForView(self.view, accessibilityElements);
   SortAccessibilityElements(accessibilityElements);
+
+  if (_accessibilityElementsBlock) {
+    NSArray *accessibilityElementsCopy = [accessibilityElements copy];
+    accessibilityElements = _accessibilityElementsBlock(accessibilityElementsCopy);
+  }
+
   return accessibilityElements;
 }
 
--- a/Source/ASDisplayNode.h
+++ b/Source/ASDisplayNode.h
@@ -782,6 +782,12 @@ AS_EXTERN NSInteger const ASDefaultDrawingPriority;
 
 @end
 
+/**
+ * Accessibility elements modification block. Used to modify/override the elements returned by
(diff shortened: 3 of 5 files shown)


Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -257,6 +258,7 @@ - (void)setAccessibilityElements:(NSArray *)accessibilityElements
 {
   ASDisplayNodeAssertMainThread();
   _accessibilityElements = nil;
```
nguyenhuy: It's odd that this setter doesn't consider the provided param and always set the ivar to nil. We probably should rename it to `resetAccessibilityElements` or `clearAccessibilityElements`?

raycsh017:
@nguyenhuy 
I like that idea, but in that case, wouldn't each node have to know in what context they are being displayed? For example, if there are multiple instances of a node subclass used in different view controllers and those instances need to have different `accessibilityElements` based on the view controller they are in, they would need to know where and how they are being displayed.

Source/Details/_ASDisplayViewAccessiblity.mm
raycsh017: Yeah that sounds good, I wasn't sure if this was intended or we just haven't had a chance to implement this portion yet :/

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -278,6 +284,11 @@ - (NSArray *)accessibilityElements
 
 @implementation ASDisplayNode (AccessibilityInternal)
 
+- (void)setAccessibilityElements:(NSArray *)accessibilityElements
+{
+  self.view.accessibilityElements = accessibilityElements;
```
nguyenhuy: This method would cause the node to load its view if not yet loaded. Let's do something similar to the `isNodeLoaded` check in `-accessibilityElements` setter.

Source/Details/_ASDisplayViewAccessiblity.mm
nguyenhuy: I *think* setting the provided param to `_accessibilityElements` is the right way to do.

@maicki: thoughts?

nguyenhuy: Also, if the `_accessibilityElements` ivar is updated here, we may not even need a new `_customAccessibilityElements` ivar because setting a non-nil array here will cause the setter to return it (i.e it won't call `[viewNode accessibilityElements]`.

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -268,16 +268,30 @@ - (NSArray *)accessibilityElements
     return @[];
   }
 
+  if (_accessibilityElements != nil) {
+    return _accessibilityElements;
+  }
+
   if (_accessibilityElements == nil || ASActivateExperimentalFeature(ASExperimentalDisableAccessibilityCache)) {
-    _accessibilityElements = [viewNode accessibilityElements];
+    return [viewNode accessibilityElements];
```
raycsh017: This might be the necessary evil (in that we don't cache it and return that next time) if we want to limit the `accessibilityElements` storage to one for either the custom or Texture-defined a11y elements (ie. items returned with current behavior) as opposed to two. It was hard to keep track of whether the ivar contains custom a11y elements or the default especially if you consider the case where you might be dynamically adding/removing subnodes and so you need to update the a11y elements returned for a node. Fwiw, I think another benefit to this method might be the accuracy of the list of items returned from `accessibilityElements`.

raycsh017:
Made changes to allow modification of `accessibilityElements` via `accessibilityElementsBlock` property on display nodes. Updated the test to reflect this as well!

Source/ASDisplayNode.mm
```diff
@@ -3546,6 +3546,18 @@ - (BOOL)isAccessibilityContainer
   return _isAccessibilityContainer;
 }
 
+- (void)setAccessibilityElementsBlock:(ASDisplayNodeAccessibilityElementsBlock)accessibilityElementsBlock
+{
+  MutexLocker l(__instanceLock__);
+  _accessibilityElementsBlock = accessibilityElementsBlock;
```
nguyenhuy: You probbaly want to store a copy of the provided block.

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -287,6 +287,12 @@ - (NSArray *)accessibilityElements
   NSMutableArray *accessibilityElements = [[NSMutableArray alloc] init];
   CollectAccessibilityElementsForView(self.view, accessibilityElements);
   SortAccessibilityElements(accessibilityElements);
+
+  if (_accessibilityElementsBlock) {
+    NSArray *accessibilityElementsCopy = [accessibilityElements copy];
+    accessibilityElements = _accessibilityElementsBlock(accessibilityElementsCopy);
```
nguyenhuy: I'm wondering if it's ok to give the block a mutable array, so that the block may modify it directly and we don't need to copy it beforehand?

I'm not sure if it's a good API design?

raycsh017: That would work too, I just felt this might be safer, sort of forcing people to create their own mutable array and add the object as needed and return a copy of it. Do we have an example of supporting custom blocks like this? it would be nice to follow an existing pattern too if there is one.

nguyenhuy:
My impression is that this diff is blocked/replaced by https://github.com/TextureGroup/Texture/pull/1525 (please correct me if I'm wrong, I've lost the context regarding this area and would need to spend some time to reload). With that, I'm gonna close this PR. Feel free to reopen it if need to. Thanks all.

PR #1489, opened 2019-05-05
[feat] Added download progress for NetworkImageNode

zhongwuzw:
Currently, seems we don't have any property to observe download progress, it's useful for user who can handle progress things(like progress indicator) by observing download progress.


Source/ASNetworkImageNode.h
```diff
@@ -133,6 +133,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (readonly) CGFloat renderedImageQuality;
 
+/**
+ * Loading progress of the current image.
```
nguyenhuy: Let's call it "download progress" to be a bit clearer and consistent with PINRemoteImage.

```suggestion
 * Download progress of the current image.
```

Source/ASNetworkImageNode.h
```diff
@@ -133,6 +133,13 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * When downloading a network image, this value would be updated to track downloading progress. ( number between 0.0 and 1.0 )
```
nguyenhuy:
```suggestion
 * When downloading a network image, this value would be updated to track download progress (value between 0 and 1)
```

Source/ASNetworkImageNode.h
```diff
@@ -133,6 +133,13 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * If the URL is unset, this is 1.0 if defaultImage or image is set to non-nil.
```
nguyenhuy:
```suggestion
 * If the URL is unset, this is 1 if defaultImage or image is set to non-nil.
```

Source/ASNetworkImageNode.mm
```diff
@@ -206,7 +208,9 @@ - (void)setURL:(NSURL *)URL resetToDefault:(BOOL)reset
     _imageWasSetExternally = NO;
     
     [self _locked_cancelImageDownloadWithResumePossibility:NO];
-
+    
+    [self _locked_setLoadingProgress:0.0];
```
nguyenhuy: Should we do something similar to line 221?

Source/ASNetworkImageNode.mm
```diff
@@ -277,6 +281,24 @@ - (void)_setCurrentImageQuality:(CGFloat)imageQuality
...
+  ASLockScopeSelf();
+  _loadingProgress = loadingProgress;
+}
+
+- (CGFloat)loadingProgress
+{
+  return ASLockedSelf(_loadingProgress);
+}
+
+- (void)_locked_setLoadingProgress:(CGFloat)loadingProgress
```
nguyenhuy: Let's add a explanation why we're dispatching to the main queue here instead of locking. Something similar to what `_setCurrentImageQuality` has.

nguyenhuy:
```suggestion
- (void)_setDownloadProgress:(CGFloat)downloadProgress
```

Source/ASNetworkImageNode.h
```diff
@@ -133,6 +133,13 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
+@property (readonly) CGFloat loadingProgress;
```
nguyenhuy:
```suggestion
@property (readonly) CGFloat downloadProgress;
```

Source/ASNetworkImageNode.mm
zhongwuzw: I suggest set `0.0` directly, because we actually don't do download operation?

zhongwuzw:
@nguyenhuy I renamed to `downloadProgress`, please review again. :)

Source/ASNetworkImageNode.mm
```diff
@@ -596,7 +637,14 @@ - (void)_downloadImageWithCompletion:(void (^)(id <ASImageContainerProtocol> ima
...
 
     dispatch_queue_t callbackQueue = [self callbackQueue];
-    ASImageDownloaderProgress downloadProgress = NULL;
+    __weak __typeof__(self) weakSelf = self;
+    ASImageDownloaderProgress downloadProgress = ^(CGFloat progress){
+      __typeof__(self) strongSelf = weakSelf;
+      if (strongSelf == nil) {
+        return;
+      }
+      [strongSelf _updateDownloadedProgress:progress downloadIdentifier:downloadIdentifier];
```
nguyenhuy: I think this is a bit more concise.

```
if (strongSelf != nil) {
  [strongSelf _updateDownloadedProgress:progress downloadIdentifier:downloadIdentifier];
}
```

Source/ASNetworkImageNode.mm
```diff
@@ -768,6 +817,7 @@ - (void)_lazilyLoadImageIfNecessary
           UIImage *newImage;
           if (imageContainer != nil) {
             [strongSelf _setCurrentImageQuality:1.0];
+            [self _setDownloadProgress:1.0];
```
nguyenhuy:
```suggestion
            [strongSelf _setDownloadProgress:1.0];
```

zhongwuzw: 😢 Seems I just do the copied. Thanks! 👍

Source/ASNetworkImageNode.h
```diff
@@ -133,6 +133,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (readonly) CGFloat renderedImageQuality;
 
+/**
+ * Download progress of the current image.
+ * When downloading a network image, this value would be updated to track download progress (value between 0 and 1)
+ * If the URL is unset, this is 1 if defaultImage or image is set to non-nil.
```
nguyenhuy: > If the URL is unset, this is 1 if defaultImage or image is set to non-nil.

This line is not true for download progress because the progress should be 0 (set [here](https://github.com/TextureGroup/Texture/pull/1489/files#diff-652a444aea8e41da45fd4ba883780cccR212)).

Source/ASNetworkImageNode.mm
```diff
@@ -277,6 +281,30 @@ - (void)_setCurrentImageQuality:(CGFloat)imageQuality
...
+  _downloadProgress = downloadProgress;
+}
+
+- (CGFloat)downloadProgress
+{
+  return ASLockedSelf(_downloadProgress);
+}
+
+/**
+ * Always use these methods internally to update the current image quality
```
nguyenhuy:
```suggestion
 * Always use these methods internally to update the current download progress
```

zhongwuzw: Nice catch! Thanks! 👍

zhongwuzw:
@nguyenhuy Done.

PR #1494, opened 2019-05-06
Shrink Button Node

bolsinga:
Button: 1312 to 1288 bytes, running on iPhone SE simulator. 1.9% reduction.

Shrink the enums it stores, and place them next to each other (along with a BOOL) to reduce instance size.

### DIFF
--- a/Source/ASButtonNode+Private.h
+++ b/Source/ASButtonNode+Private.h
@@ -31,14 +31,15 @@
   UIImage *_disabledBackgroundImage;
   
   CGFloat _contentSpacing;
-  BOOL  _laysOutHorizontally;
-  ASVerticalAlignment _contentVerticalAlignment;
-  ASHorizontalAlignment _contentHorizontalAlignment;
   UIEdgeInsets _contentEdgeInsets;
-  ASButtonNodeImageAlignment _imageAlignment;
   ASTextNode *_titleNode;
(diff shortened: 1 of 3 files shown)


Source/ASButtonNode+Private.h
```diff
@@ -31,14 +31,15 @@
...
-  BOOL  _laysOutHorizontally;
-  ASVerticalAlignment _contentVerticalAlignment;
-  ASHorizontalAlignment _contentHorizontalAlignment;
   UIEdgeInsets _contentEdgeInsets;
-  ASButtonNodeImageAlignment _imageAlignment;
   ASTextNode *_titleNode;
   ASImageNode *_imageNode;
   ASImageNode *_backgroundImageNode;
+
+  BOOL  _laysOutHorizontally;
```
Adlai-Holler: Should we invest in a bitfield so that future authors don't put another BOOL?

bolsinga: I'd considered it but decided against it at the moment. It's only 1 right now, and adding another will not be significant (if it is added by this other `BOOL`). The things to do are group small types (BOOLs and other unsigned chars or shorts). Everything else is 64 bits nowadays; the worse thing to do from a memory size perspective is to intersperse BOOLs and pointers.

PR #1495, opened 2019-05-06
Shrink TextNode(2)

bolsinga:
Running in iPhone SE Simulator:
- ASTextNode: 1360 to 1352 bytes .6% reduction
- ASTextNode2: 1360 to 1304 bytes 4.3% reduction

Shrink stored enum sizes. Group BOOLs near these smaller enums. Override -usingExperiment to return constant instead of never set, but allocated ivar.


bolsinga:
> Should we invest in a bitfield for the bools?

Same as before; I didn't want to change things too much. This group of changes gets Pinterest some good wins (and I assume all other clients). It's good enough at the moment.

PR #1496, opened 2019-05-06
Shrink Pager and Scroll Node

bolsinga:
Running on iPhone SE Simulator:
ASPagerNode 1240 to 1232 bytes .7% reduction
ASScrollNode 1104 to 1096 bytes .7% reduction

- make ASScrollNode smaller by shrinking ASScrollDirection, which requires a change to ASRangeController
- shrink ASPagerNode by declaring the BOOL next to the struct

### DIFF
--- a/Source/ASPagerNode.mm
+++ b/Source/ASPagerNode.mm
@@ -29,6 +29,7 @@
     unsigned nodeBlockAtIndex:1;
     unsigned nodeAtIndex:1;
   } _pagerDataSourceFlags;
+    BOOL _allowsAutomaticInsetsAdjustment;
 
   __weak id <ASPagerDelegate> _pagerDelegate;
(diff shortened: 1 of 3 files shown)


Source/ASPagerNode.mm
```diff
@@ -29,6 +29,7 @@ @interface ASPagerNode () <ASCollectionDataSource, ASCollectionDelegate, ASColle
     unsigned nodeBlockAtIndex:1;
     unsigned nodeAtIndex:1;
   } _pagerDataSourceFlags;
+    BOOL _allowsAutomaticInsetsAdjustment;
```
Adlai-Holler: Should we broaden the flags struct and add this into it?

bolsinga: I didn't want to conflate the data source with the object BOOL properties.

PR #1509, opened 2019-05-10
Discard _pendingDisplayNodes after subtree is rendered

Adlai-Holler:
There's a lot more we could do to optimize this, but at least this makes the memory growth curve less steep. @bolsinga 💕 

### Before (note NSHashTables)
Screen Shot 2019-05-10 at 11 28 25 AM

### After

Screen Shot 2019-05-10 at 11 28 23 AM


bolsinga:
Super!

PR #1512, opened 2019-05-13
[Bug] ASButtonNode TintColor Change

GeekTree0101:
ASButtonNode tintColor doesn't work.
In normally, Some developer use an ASImageNodeTintColorModificationBlock
I hope that imageNode applied tintColor automatically from ASButtonNode tintColor property

```objc
- (void)tintColorDidChange
{
  [super tintColorDidChange];
  _imageNode.imageModificationBlock = ASImageNodeTintColorModificationBlock(self.tintColor);
  [_imageNode setNeedsDisplay];
}
```

### DIFF
--- a/Source/ASButtonNode.mm
+++ b/Source/ASButtonNode.mm
@@ -562,4 +562,11 @@
   _titleNode.hidden = (_titleNode.attributedText.length == 0);
 }
 
+- (void)tintColorDidChange
+{
+  [super tintColorDidChange];
+  _imageNode.imageModificationBlock = ASImageNodeTintColorModificationBlock(self.tintColor);
+  [_imageNode setNeedsDisplay];
+}
+
 @end
--- a/Tests/ASButtonNodeTests.mm
+++ b/Tests/ASButtonNodeTests.mm
@@ -51,4 +51,13 @@
                 buttonNode.defaultAccessibilityTraits);
 }
 
+- (void)testTintColor
+{
+  ASButtonNode *buttonNode = nil;
+  buttonNode = [[ASButtonNode alloc] init];
+  XCTAssertFalse(buttonNode.tintColor == [UIColor whiteColor]);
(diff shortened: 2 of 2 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f84cead54980f416b173d0667a50df43/raw/389d863cb7c0e0a4e67ab2ee17c2b3c341121ffb/log.txt)

Source/ASButtonNode.mm
```diff
@@ -562,4 +562,11 @@ - (void)layout
   _titleNode.hidden = (_titleNode.attributedText.length == 0);
 }
 
+- (void)tintColorDidChange
+{
+  [super tintColorDidChange];
+  _imageNode.imageModificationBlock = ASImageNodeTintColorModificationBlock(self.tintColor);
```
nguyenhuy: This will override any modification block set by clients beforehand. 

You may want to implement a new block that first calls the existing block, if any, and then calls a tint color block. Even this won't prevent users from setting their own block after changing the tint color and thus not seeing the color applied. We may want something like a "final tint color modification block" that is guaranteed to be executed last. Perhaps by adding an ASImageNode subclass?

GeekTree0101: Yes, just want to use only imageNode of ASButtonNode.
So, How about null check about imageModificationBlock of ASButtonNode imageNode property?

> Avoid any modification block set by clients beforehand.
```objc
- (void)tintColorDidChange
{
  [super tintColorDidChange];
  if !(_imageNode.imageModificationBlock) {
    _imageNode.imageModificationBlock = ASImageNodeTintColorModificationBlock(self.tintColor);
    [_imageNode setNeedsDisplay];
  }
}
```
Also, if this method doesn't good then i will close this PR :) 

Thank you!

PR #1516, opened 2019-05-23
Assign accessibilityElements to _accessibilityElements

maicki:
We should assign the `accessibilityElements` to the `_accessibilityElements` property.


nguyenhuy:
FYI, @raycsh017.

raycsh017:
👍 thanks for the heads up!

PR #1520, opened 2019-05-28
Add new property alwaysHandleTapTruncationAction to ASTextNode2 and ASTextNode.

dirtmelon:
After this [PR](https://github.com/TextureGroup/Texture/pull/1184), `ASTextNode` and `ASTextNode2` don't handle touches on additional attributed message if `passthroughNonlinkTouches` is YES. But sometimes we maybe want to handle these touches action, even if `passthroughNonlinkTouches` is YES.  

So I add the new property `alwaysHandleTapTruncationAction` to `ASTextNode` and `ASTextNode2`, when `alwaysHandleTapTruncationAction` is YES, `ASTextNode` and `ASTextNode2` always handle touches on additional attributed message.

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -1143,6 +1143,12 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
   return _alwaysHandleTruncationTokenTap;
 }
 
+- (void)setAlwaysHandleTruncationTokenTap:(BOOL)alwaysHandleTruncationTokenTap
+{
+  ASLockScopeSelf();
+  _alwaysHandleTruncationTokenTap = alwaysHandleTruncationTokenTap;
+}
+
 #pragma mark - Shadow Properties
 
 - (CGColorRef)shadowColor
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -209,6 +209,7 @@ static ASTextKitRenderer *rendererForAttributes(ASTextKitAttributes attributes, 
   ASTextNodeHighlightStyle _highlightStyle;
   BOOL _longPressCancelsTouches;
   BOOL _passthroughNonlinkTouches;
+  BOOL _alwaysHandleTapTruncationAction;
 }
 @dynamic placeholderEnabled;
 
@@ -1010,7 +1011,8 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
   NSUInteger lastCharIndex = NSIntegerMax;
   BOOL linkCrossesVisibleRange = (lastCharIndex > range.location) && (lastCharIndex < NSMaxRange(range) - 1);
 
-  if (range.length > 0 && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
+  if ((range.length > 0 && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) ||
+    _alwaysHandleTapTruncationAction) {
     return YES;
   } else {
     return NO;
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -993,7 +993,8 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
 - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event
 {
   ASDisplayNodeAssertMainThread();
-  
+  ASLockScopeSelf(); // Protect usage of _passthroughNonlinkTouches and _alwaysHandleTruncationTokenTap ivars.
+
   if (!_passthroughNonlinkTouches) {
     return [super pointInside:point withEvent:event];
   }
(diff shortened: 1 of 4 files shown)


dirtmelon:
Could you please help to review this pr? Many thanks. @maicki @Adlai-Holler @garrettmoon @appleguy @nguyenhuy @bolsinga

Adlai-Holler:
I can't remember why we decided not to call the delegate method if passthroughNonlinkTouchesEnabled again? I reviewed the original PR but it's not clear why that delegate call is a problem. @maicki What do you think of this diff?

dirtmelon:
I found this [issue](https://github.com/facebookarchive/AsyncDisplayKit/issues/3259), possibly because of this.Sometimes we don't want to handle these touch actions. @Adlai-Holler @maicki

dirtmelon:
@nguyenhuy @Adlai-Holler Please review if possible.

nguyenhuy:
@maicki ping

maicki:
@dirtmelon Thanks for this PR.

I'm good with the change. Maybe let's use a different name that is more aligned with the callback method that is called like: `alwaysHandleTruncationTokenTap` and add as documentation to the property that if this is set to YES, the `-`-[ASTextNodeDelegate textNodeTappedTruncationToken:]` callback will be called.

dirtmelon:
@maicki I have renamed it, please review it again.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4761820175b5b5ee65e229bb5791e702/raw/2547335e57b63e2e7b3672735657d8a6556cdc94/log.txt)

dirtmelon:
@nguyenhuy Could you please review it?

Source/ASTextNode.mm
```diff
@@ -1047,7 +1048,8 @@ - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event
   NSUInteger lastCharIndex = NSIntegerMax;
   BOOL linkCrossesVisibleRange = (lastCharIndex > range.location) && (lastCharIndex < NSMaxRange(range) - 1);
 
-  if (range.length > 0 && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
+  if ((range.length > 0 && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) ||
+    _alwaysHandleTruncationTokenTap) {
```
nguyenhuy: Nit: if we always return `YES` when `_alwaysHandleTruncationTokenTap` is set then I think it's better to return as early as possible. So we can do this [here](https://github.com/TextureGroup/Texture/pull/1520/files#diff-271d6b828178013b300f794e8ec8c740R1037):

```
if (_alwaysHandleTruncationTokenTap) {
  return YES;
}
```

nguyenhuy: Another thing: we don't acquire the lock while accessing/setting `_alwaysHandleTruncationTokenTap`. We should fix that.

Same deal for `_passthroughNonlinkTouches` and `delegate`. It'd be great if you can help to fix them in a follow-up PR.

nguyenhuy:
Also, can you please rebase with latest master to fix the CI issue?

Source/ASTextNode.mm
dirtmelon: @nguyenhuy done, I will add the lock of `_passthroughNonlinkTouches` and `delegate` in a follow-up PR.

Source/ASTextNode.mm
```diff
@@ -1010,7 +1011,8 @@ - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event
   NSUInteger lastCharIndex = NSIntegerMax;
   BOOL linkCrossesVisibleRange = (lastCharIndex > range.location) && (lastCharIndex < NSMaxRange(range) - 1);
 
-  if (range.length > 0 && !linkCrossesVisibleRange && linkAttributeValue != nil && linkAttributeName != nil) {
+  if (_alwaysHandleTruncationTokenTap ||
```
nguyenhuy: Sorry I wasn't very clear. What I meant in my previous comment is that we can check this property early (after the `_passthroughNonlinkTouches` check) and return immediately. We wouldn't bother calling `_linkAttributeValueAtPoint:` because its results wouldn't be considered anyways.

nguyenhuy: Also, you also need to have the lock before accessing the ivar. One simple way for now is using the getter.

Source/ASTextNode.mm
```diff
@@ -1132,6 +1134,12 @@ - (BOOL)_pendingTruncationTap
   return [_highlightedLinkAttributeName isEqualToString:ASTextNodeTruncationTokenAttributeName];
 }
 
+- (BOOL)alwaysHandleTruncationTokenTap
+{
+  ASLockScopeSelf();
+  return _alwaysHandleTruncationTokenTap;
+}
+
```
nguyenhuy: You'd need to implement the setter and acquire the lock there as well.

Source/ASTextNode.mm
dirtmelon: @nguyenhuy Thanks for your suggestion.

PR #1531, opened 2019-05-31
Remove the text rendering experiment. 

garrettmoon:
It decreased rendering speeds at a negligible improvement to OOMs.


garrettmoon:
Not sure how we want to handle merging this as it will affect the existing experiment.

PR #1537, opened 2019-06-04
Fix multiple issues around accessibility handlinig

maicki:
- Fixes calculations for accessibilityFrame of ASAccessibilityCustomAction and ASAccessibilityElement
- Fix crash for accessibility elements within layer backed nodes
- Fix accessibility action label updating after other actions change the accessibility label of the corresponding node

### DIFF
--- a/Source/Details/_ASDisplayViewAccessiblity.mm
+++ b/Source/Details/_ASDisplayViewAccessiblity.mm
@@ -13,14 +13,12 @@
 #import <AsyncDisplayKit/ASAvailability.h>
 #import <AsyncDisplayKit/ASCollectionNode.h>
 #import <AsyncDisplayKit/ASDisplayNodeExtras.h>
+#import <AsyncDisplayKit/ASDisplayNode+FrameworkPrivate.h>
+#import <AsyncDisplayKit/ASDisplayNode+Ancestry.h>
 #import <AsyncDisplayKit/ASDisplayNodeInternal.h>
 #import <AsyncDisplayKit/ASTableNode.h>
 
 #import <queue>
-
-NS_INLINE UIAccessibilityTraits InteractiveAccessibilityTraitsMask() {
-  return UIAccessibilityTraitLink | UIAccessibilityTraitKeyboardKey | UIAccessibilityTraitButton;
-}
 
 #pragma mark - UIAccessibilityElement
 
@@ -36,7 +34,7 @@ typedef NSComparisonResult (^SortAccessibilityElementsComparator)(id<ASAccessibi
 static void SortAccessibilityElements(NSMutableArray *elements)
 {
   ASDisplayNodeCAssertNotNil(elements, @"Should pass in a NSMutableArray");
-  
+
   static SortAccessibilityElementsComparator comparator = nil;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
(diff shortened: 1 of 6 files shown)


Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -55,22 +53,29 @@ static void SortAccessibilityElements(NSMutableArray *elements)
   [elements sortUsingComparator:comparator];
 }
 
+static CGRect ASAccessibilityFrameForNode(ASDisplayNode *node) {
+  ASDisplayNode *supernode = node.firstNonLayerNode;
+  ASDisplayNodeCAssert(!supernode.isLayerBacked, @"No non-layerbacked node found.");
+
+
+  CALayer *layer = node.layer;
+  return [layer convertRect:node.bounds toLayer:ASFindWindowOfLayer(layer).layer];
```
Adlai-Holler: Should we remove `firstNonLayerNode` since it isn't actually used here?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -1100,13 +1101,25 @@ - (NSString *)accessibilityLabel
...
   }
+
+  // We need to update action name when it's changed to reflect the latest state.
+  // Note: Update the custom action itself won't work when a11y is inside a list of custom actions
+  // in which one action results in a name change in the next action. In that case the UIAccessibility
+  // will hold the old action strongly until a11y jumps out of the list of custom actions.
+  // Thus we can only update name in place to have the change take effect.
+  BOOL needsUpdateActionName = self.isNodeLoaded && ![oldAccessibilityLabel isEqualToString:accessibilityLabel] && (_accessibilityTraits & ASInteractiveAccessibilityTraitsMask());
+  if (needsUpdateActionName) {
+    self.accessibilityCustomAction.name = accessibilityLabel;
```
Adlai-Holler: Is this same update performed/needed when the accessibility traits change?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -1100,13 +1101,25 @@ - (NSString *)accessibilityLabel
...
     NSAttributedString *accessibilityAttributedLabel = accessibilityLabel ? [[NSAttributedString alloc] initWithString:accessibilityLabel] : nil;
     _setAccessibilityToViewAndProperty(_accessibilityAttributedLabel, accessibilityAttributedLabel, accessibilityAttributedLabel, accessibilityAttributedLabel);
   }
+
+  // We need to update action name when it's changed to reflect the latest state.
+  // Note: Update the custom action itself won't work when a11y is inside a list of custom actions
+  // in which one action results in a name change in the next action. In that case the UIAccessibility
+  // will hold the old action strongly until a11y jumps out of the list of custom actions.
+  // Thus we can only update name in place to have the change take effect.
+  BOOL needsUpdateActionName = self.isNodeLoaded && ![oldAccessibilityLabel isEqualToString:accessibilityLabel] && (_accessibilityTraits & ASInteractiveAccessibilityTraitsMask());
```
Adlai-Holler: Use `0 != _traits & ASInteractiveTraits()` so that we get `1` instead of some arbitrary non-zero value.

Source/Private/ASDisplayNode+UIViewBridge.mm
maicki: If traits change within a node the right behavior would that to actually propagate that up to the next accessibility container. This is out of this PR right now though.

Source/Private/ASDisplayNode+UIViewBridge.mm
maicki: Done

Source/Details/_ASDisplayViewAccessiblity.mm
maicki: done

PR #1550, opened 2019-06-12
[ASTextNode2] Fix truncation token is wrong after attributed string is changing with different foreground color

maicki:


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/fa2c0f6a0e9826699875e2ec303d5788/raw/2547335e57b63e2e7b3672735657d8a6556cdc94/log.txt)

maicki:
Yes that is still needed. `+layoutWithContainer` will follow in a future PR.

PR #1571, opened 2019-07-05
Cleanup for cgFloatForYogaFloat

maicki:
- Return CGFloat
- Proper check for YGUndefined
- Add YGUndefined default return value

### DIFF
--- a/Source/Layout/ASYogaUtilities.mm
+++ b/Source/Layout/ASYogaUtilities.mm
@@ -117,9 +117,9 @@ float yogaFloatForCGFloat(CGFloat value)
   }
 }
 
-float cgFloatForYogaFloat(float yogaFloat)
+CGFloat cgFloatForYogaFloat(float yogaFloat, CGFloat undefinedDefault)
 {
-  return (yogaFloat == YGUndefined) ? CGFLOAT_MAX : yogaFloat;
+  return YGFloatIsUndefined(yogaFloat) ? undefinedDefault : yogaFloat;
 }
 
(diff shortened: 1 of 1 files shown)


Source/Layout/ASYogaUtilities.mm
```diff
@@ -213,8 +213,8 @@ YGSize ASLayoutElementYogaMeasureFunc(YGNodeRef yogaNode, float width, YGMeasure
   id <ASLayoutElement> layoutElement = (__bridge id <ASLayoutElement>)YGNodeGetContext(yogaNode);
   ASDisplayNodeCAssert([layoutElement conformsToProtocol:@protocol(ASLayoutElement)], @"Yoga context must be <ASLayoutElement>");
 
-  width = cgFloatForYogaFloat(width);
-  height = cgFloatForYogaFloat(height);
+  width = cgFloatForYogaFloat(width, CGFLOAT_MAX);
+  height = cgFloatForYogaFloat(height, CGFLOAT_MAX);
```
rahul-malik: Is there a case where the undefined default is something besides CGFLOAT_MAX?

nguyenhuy: good point. I think it's cleaner to stick to `CGFLOAT_MAX` as the default value (and thus no need for the extra `undefinedDefault` param) until we have cases in which `CGFLOAT_MAX` is not desired.

maicki: There will be a case where `CGFLOAT_MAX` is not desired ...

PR #1573, opened 2019-07-09
Fixes accessibilityLabel bug on ASButtonNode that has no title

SterlingWaves:
We use the `ASButtonNode` as a pure icon button that has no title, and I also give this button an accessibility label. Since this buttton's `titleNode` will be non-nil if the `self.titleNode` getter is ever called, my accessibility label will get reset to empty string every time the button's state changes (`selected`, `enabled`, etc.). However, I'd like my button's accessibility label to remain consistent in this case. This PR addresses this problem, i.e. if the button title and the new title are both either empty or nil, we do not reset the accessibility label. Also added corresponding test.

### DIFF
--- a/Source/ASButtonNode.mm
+++ b/Source/ASButtonNode.mm
@@ -179,9 +179,10 @@
     newTitle = _normalAttributedTitle;
   }
 
-  // Calling self.titleNode is essential here because _titleNode is lazily created by the getter.
-  if (((_titleNode != nil && self.titleNode.attributedText.length > 0) || newTitle.length > 0) && [self.titleNode.attributedText isEqualToAttributedString:newTitle] == NO) {
-    _titleNode.attributedText = newTitle;
+  NSAttributedString *attributedString = _titleNode.attributedText;
+  if ((attributedString.length > 0 || newTitle.length > 0) && [attributedString isEqualToAttributedString:newTitle] == NO) {
+    // Calling self.titleNode is essential here because _titleNode is lazily created by the getter.
+    self.titleNode.attributedText = newTitle;
     [self unlock];
     
     self.accessibilityLabel = self.defaultAccessibilityLabel;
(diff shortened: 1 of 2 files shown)


nguyenhuy:
@maicki Can you take a look at this one when you have time?

maicki:
I think we should change the logic a bit, the if clause is really hard to read. And I also think we don't need to call the setter not before within the if clause:
```
NSAttributedString *attributedString = _titleNode.attributedText;
  if ((attributedString.length > 0 || newTitle.length > 0) && [attributedString isEqualToAttributedString:newTitle] == NO) {
    // Calling self.titleNode is essential here because _titleNode is lazily created by the getter.
    self.titleNode.attributedText = newTitle;
    [self unlock];
    
    self.accessibilityLabel = self.defaultAccessibilityLabel;
    [self updateYogaLayoutIfNeeded];
    [self setNeedsLayout];
    return;
  }
```

Also @SterlingWaves would you mind adding this `ASButtonNode` unit test:
```
- (void)testUpdateTitle
{
  NSAttributedString *title = [[NSAttributedString alloc] initWithString:@"MyTitle"];
  ASButtonNode *buttonNode = [[ASButtonNode alloc] init];
  [buttonNode setAttributedTitle:title forState:UIControlStateNormal];
  XCTAssertTrue([[buttonNode attributedTitleForState:UIControlStateNormal] isEqualToAttributedString:title]);
  XCTAssert([buttonNode.titleNode.attributedText isEqualToAttributedString:title]);
}
```

@nguyenhuy Thoughts?

nguyenhuy:
Yeah, agree with your suggestions, @maicki.

SterlingPinterest:
@maicki @nguyenhuy updated.

nguyenhuy:
Thanks @SterlingPinterest. I assume @maicki would approve this PR so I'm gonna go ahead and merge it.

PR #1575, opened 2019-07-10
Fix the build warning in `ASRangeController.mm`

hanton:
Fix the following warning in the build process:

```
⚠️  /Texture/Source/Details/ASRangeController.mm:402:66: 'scrollableDirections' is deprecated: Use ASCollectionNode property instead. [-Wdeprecated-declarations]
```

### DIFF
--- a/Source/Details/ASRangeController.mm
+++ b/Source/Details/ASRangeController.mm
@@ -11,7 +11,6 @@
 
 #import <AsyncDisplayKit/_ASHierarchyChangeSet.h>
 #import <AsyncDisplayKit/ASAssert.h>
-#import <AsyncDisplayKit/ASCellNode+Internal.h>
 #import <AsyncDisplayKit/ASCollectionElement.h>
 #import <AsyncDisplayKit/ASCollectionView.h>
 #import <AsyncDisplayKit/ASDisplayNodeExtras.h>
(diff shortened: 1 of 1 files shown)


Source/Details/ASRangeController.mm
```diff
@@ -399,7 +400,7 @@ - (void)_updateVisibleNodeIndexPaths
   if (ASDisplayNode.shouldShowRangeDebugOverlay) {
     ASScrollDirection scrollableDirections = ASScrollDirectionUp | ASScrollDirectionDown;
     if ([_dataSource isKindOfClass:NSClassFromString(@"ASCollectionView")]) {
-        scrollableDirections = ((ASCollectionView *)_dataSource).scrollableDirections;
+        scrollableDirections = ((ASCollectionView *)_dataSource).collectionNode.scrollableDirections;
```
nguyenhuy: I agree that we should fix this warning. However, given that this is a legitimate usage of the collection view's property, I don't think this is the best fix. Another way to fix it is simply importing `ASCollectionView+Undeprecated.h`. We have done this elsewhere in the framework.

hanton:
hi @nguyenhuy , thanks for pointing this out, I've updated the commit :P

PR #1579, opened 2019-07-10
Bridge CTFontRef directly to UIFont when converting attributes on NSAttributedString

rahul-malik:
In iOS13, there is an issue with the font returned from "-[UIFontj fontWithName:size:]". When asking for "SFUI-Regular" returns "Times New Roman" font. It seems like a reasonable default would be to return the system font in the case that the requested font was not respected.

### DIFF
--- a/Source/TextKit/ASTextKitCoreTextAdditions.mm
+++ b/Source/TextKit/ASTextKitCoreTextAdditions.mm
@@ -73,8 +73,9 @@ NSDictionary *NSAttributedStringAttributesForCoreTextAttributes(NSDictionary *co
       CGFloat fontSize = CTFontGetSize(coreTextFont);
       UIFont *font = [UIFont fontWithName:fontName size:fontSize];
       ASDisplayNodeCAssertNotNil(font, @"unable to load font %@ with size %f", fontName, fontSize);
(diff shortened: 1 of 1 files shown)


rahul-malik:
@nguyenhuy - Yep! Radar has been filed under FB6592770

mikezucc:
somewhere, somehow, someone consciously put Times New Roman in an catch/else case

rahul-malik:
@maicki @nguyenhuy - I updated the title since the change is more than updating the fallback. It isn't clear why we were trying to create a new font instance when CTFontRef toll-free bridges to UIFont. This appears to pass tests both internally at Pinterest and Texture so I feel like the change is pretty safe and should also solve the iOS 13 regression users are experiencing

nguyenhuy:
Great change, @rahul-malik! Yeah, I think it's safe and we should just get it in.

The original logic is pretty old -- dates back to the original commit. Perhaps @appleguy has some context on why it's done that way.

ex3ndr:
Hey! Any ETA to new release with this fix? This is pretty heavy issue.

rahul-malik:
Hey! We’re actively working on creating a new release now.

PR #1581, opened 2019-07-12
Update .gitignore file [Carthage]

hanton:
Update .gitignore file [Carthage]


.gitignore
```diff
@@ -32,3 +32,11 @@ playground.xcworkspace
 /.buckconfig.local
 /.buckd
 
+# CocoaPods
+*.xcworkspace
+Podfile.lock
```
wiseoldduck: https://guides.cocoapods.org/using/using-cocoapods.html -> "Whether or not you check in the Pods directory, the Podfile and Podfile.lock should always be kept under version control."

hanton: hi @wiseoldduck , thanks for pointing out this, I've updated the commit :)

btw, it seems the repo's example projects don't host `Podfile.lock` at the moment, do you think it is good to add them to be under version control?

nguyenhuy: Yeah examples ideally should have `Podfile.lock`. Not a big deal I'd say.

.gitignore
```diff
@@ -32,3 +32,10 @@ playground.xcworkspace
 /.buckconfig.local
 /.buckd
 
+# CocoaPods
+*.xcworkspace
```
nguyenhuy: Can we not ignore `AsyncDisplayKit.xcworkspace`? We've been using it to sync some project-wide settings across the team.

Also just realized that [this line](https://github.com/TextureGroup/Texture/blame/master/docs/_docs/development/how-to-develop.md#L12) in the dev doc is not correct. We should fix it as well!

.gitignore
hanton: Just noticing the `build.sh`, which will be ran by CI, ignores `Podfile.lock` in the example projects every time it makes a new build

```
...
if [ -f "${example}/Podfile" ]; then
        echo "Using CocoaPods"
        if [ -f "${example}/Podfile.lock" ]; then
            rm "$example/Podfile.lock"
        fi
...
```

So I think it's safe to not commit those `Podfile.lock` at the moment :p

.gitignore
hanton: Sorry for not paying enough attention to it. Let's keep `.xcworkspace` files for this purpose. I've updated the commit to just ignore `Carthage` related files this time.

hanton: Could you clarify why do we need to fix [this line](https://github.com/TextureGroup/Texture/blame/master/docs/_docs/development/how-to-develop.md#L12)? 

```
Run `pod install` in the directory that you cloned to.
```

Because the `Pods` folder is not kept under version control, it seems we need to run `pod install` after cloning the repo.

nguyenhuy: ah, you're right. That line is fine.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d45dfe243f2aa27c5ec69b0eb3f6cb58/raw/500e0b7490b2368e246959627b84ab4b5e5aad9d/log.txt)

PR #1588, opened 2019-07-17
Set background color directly to view if not layer backed

rahul-malik:
UIColor can now be a dynamic color provider to support capabilities like Dark Mode in applications. UIKit is able to re-render views when a user enters dark mode but this dynamic capabilities is not handled at the layer level.

Currently this takes the recommendation in https://github.com/facebookarchive/AsyncDisplayKit/pull/1537#issuecomment-210728772 to set background color on the view directly or layer depending on if the node is layer backed. This change allows UIKit to track these colors and update the UI when the UITraitCollection.userInterfaceStyle changes.


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/fd2e4d2c8d63ba6ce20c821164a18021/raw/3b2631673a41448f74022459038f07eb5c5c486d/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/6891682563dcded3fba3426c39ca31ed/raw/fd2e230b08bf3cdc4c36e98d2b93a4bebc6daaff/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ecdd4a0901e1712d9f5740bd9ae9e51e/raw/8eed9a4e46b54430f848a9203bcb6c53512f5e8d/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5909d2b62134a0247db2e8d3b12a9c15/raw/169040303219d1562017b1f1990d50bf5c39edf7/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/716009346f4bdb344dbba7fc41aae259/raw/b65aa1f27a4a42f365b00fcd417db8c07fb46307/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/29b0a80f0828093ac231c3e9c06cbe8a/raw/e04cedd4568a90a03be65fad69cfd993b7c650c5/log.txt)

linlinyao1:
Dynamic capability is lost during conversion between UIColor and CGColor. Thus a UIColor reference is required in this case

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -727,7 +734,14 @@ - (void)setContentMode:(UIViewContentMode)contentMode
 - (UIColor *)backgroundColor
 {
   _bridge_prologue_read;
-  return [UIColor colorWithCGColor:_getFromLayer(backgroundColor)];
+  if (_loaded(self)) {
+    if (_view != nil) {
```
nguyenhuy: Use `_flags.layerBacked` to be consistent with other methods?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -727,7 +734,14 @@ - (void)setContentMode:(UIViewContentMode)contentMode
...
+      return _view.backgroundColor;
+    } else {
+      return [UIColor colorWithCGColor:_getFromLayer(backgroundColor)];
+    }
+  }
+  return [UIColor colorWithCGColor:ASDisplayNodeGetPendingState(self).backgroundColor];
```
nguyenhuy: This looks fine for now, assuming we'll change `_ASPendingState` to store a UIColor in a follow up diff.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ef8ebac2716e4caff5894681c4e8997a/raw/6a8f7af1a5f8f54e7ce353aa4907ef2327e2359a/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/1a9c5d04911db13ea81725a08b41fc50/raw/e117d5654c450b10902a234dfe43445a22feb8dc/log.txt)

Podfile.lock
```diff
@@ -52,4 +52,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 445046ac151568c694ff286684322273f0b597d6
 
-COCOAPODS: 1.6.0
+COCOAPODS: 1.6.1
```
maicki: We should not update the cocoapods version with this commit imho

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -514,15 +514,17 @@ - (BOOL)isOpaque
...
 - (void)setOpaque:(BOOL)newOpaque
 {
   _bridge_prologue_write;
-  
   BOOL shouldApply = ASDisplayNodeShouldApplyBridgedWriteToView(self);
   
   if (shouldApply) {
-    BOOL oldOpaque = _layer.opaque;
+    BOOL oldOpaque;
+    oldOpaque = _layer.opaque;
```
maicki: Why not directly assigning: `BOOL oldOpaque = _layer.opaque;`

Tests/ASDisplayNodeTests.mm
```diff
@@ -1924,16 +1925,11 @@ - (void)checkBackgroundColorOpaqueRelationshipWithViewLoaded:(BOOL)loaded layerB
   XCTAssertTrue(node.opaque, @"Node should start opaque");
   XCTAssertTrue(node.layer.opaque, @"Node should start opaque");
 
-  node.backgroundColor = [UIColor clearColor];
+  node.backgroundColor = [UIColor blackColor];
```
maicki: Could you elaborate why this test is failing now?

Source/Private/_ASPendingState.mm
```diff
@@ -1095,13 +1095,8 @@ - (void)applyToView:(UIView *)view withSpecialPropertiesHandling:(BOOL)specialPr
...
 
   if (flags.setBackgroundColor) {
-    // We have to make sure certain nodes get the background color call directly set
-    if (specialPropertiesHandling) {
-      view.backgroundColor = [UIColor colorWithCGColor:backgroundColor];
-    } else {
-      // Set the background color to the layer as in the UIView bridge we use this value as background color
-      layer.backgroundColor = backgroundColor;
-    }
+    view.backgroundColor = [UIColor colorWithCGColor:backgroundColor];
```
maicki: If we set the background color to the view the layer should automatically be updated. Could you double cbeck that this is not the case.

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -740,13 +749,13 @@ - (void)setBackgroundColor:(UIColor *)newBackgroundColor
...
     
-    BOOL specialPropertiesHandling = ASDisplayNodeNeedsSpecialPropertiesHandling(checkFlag(Synchronous), _flags.layerBacked);
-    if (specialPropertiesHandling) {
-        _view.backgroundColor = newBackgroundColor;
+    if (_flags.layerBacked) {
+      _layer.backgroundColor = newBackgroundCGColor;
     } else {
-        _layer.backgroundColor = newBackgroundCGColor;
+      _view.backgroundColor = newBackgroundColor;
+      _layer.backgroundColor = _view.backgroundColor.CGColor;
```
maicki: Same as below if you update the view I don't think you need to update the layer.

Source/Private/_ASPendingState.mm
rahul-malik: @maicki - We tested this today and surprisingly, the layer is not updated (at least not immediately)

Source/Private/ASDisplayNode+UIViewBridge.mm
rahul-malik: @nguyenhuy and I were debugging and realized we needed to set both of these. UIView seems to be setting the `_layer.backgroundColor` to `nil` when setting the views background color.

Source/Private/ASDisplayNode+UIViewBridge.mm
rahul-malik: The values of opaque differ depending on if it's a view or layer backed.

Podfile.lock
rahul-malik: I'll undo this

Tests/ASDisplayNodeTests.mm
rahul-malik: This was just for testing to see if `clearColor` was changing the interpretation of `isOpaque` on `CALayer`

Source/Private/ASDisplayNode+UIViewBridge.mm
rahul-malik: We assume these values are bridged but when debugging it seems like the relationship between view and layer is more complex. We're only hitting this now since we are setting backgroundColor directly on the view itself

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -511,17 +511,20 @@ - (void)layoutIfNeeded
...
 
+
 - (void)setOpaque:(BOOL)newOpaque
 {
   _bridge_prologue_write;
-  
   BOOL shouldApply = ASDisplayNodeShouldApplyBridgedWriteToView(self);
   
   if (shouldApply) {
     BOOL oldOpaque = _layer.opaque;
```
nguyenhuy: Can we rely on the layer for the old value? Should we get it from the view if it's view-backed?

Maybe this is ok because in this setter, we always set to the layer. But then clients may actually reach out to the layer and set it directly there. I think the safest approach would be to do similar to the getter (get from the view if it's there).

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -740,13 +750,13 @@ - (void)setBackgroundColor:(UIColor *)newBackgroundColor
...
     
-    BOOL specialPropertiesHandling = ASDisplayNodeNeedsSpecialPropertiesHandling(checkFlag(Synchronous), _flags.layerBacked);
-    if (specialPropertiesHandling) {
-        _view.backgroundColor = newBackgroundColor;
+    if (_flags.layerBacked) {
+      _layer.backgroundColor = newBackgroundCGColor;
     } else {
-        _layer.backgroundColor = newBackgroundCGColor;
+      _view.backgroundColor = newBackgroundColor;
+      _layer.backgroundColor = _view.backgroundColor.CGColor;
```
nguyenhuy: Let's add a comment here on why we're getting the CGColor from the view instead of just using `newBackgroundCGColor`.

Tests/ASDisplayNodeTests.mm
```diff
@@ -1922,18 +1923,23 @@ - (void)checkBackgroundColorOpaqueRelationshipWithViewLoaded:(BOOL)loaded layerB
   }
 
   XCTAssertTrue(node.opaque, @"Node should start opaque");
-  XCTAssertTrue(node.layer.opaque, @"Node should start opaque");
+  if (isLayerBacked) {
+      XCTAssertTrue(node.layer.opaque, @"Set background color should not have made this layer not opaque");
+    } else {
+      XCTAssertTrue(node.view.opaque, @"Set background color should not have made this view not opaque");
+    }
```
nguyenhuy: Is this necessary? At the beginning the view and the layer should agree with each other?

Nit: indentation is off.

Tests/ASDisplayNodeTests.mm
```diff
@@ -1922,18 +1923,23 @@ - (void)checkBackgroundColorOpaqueRelationshipWithViewLoaded:(BOOL)loaded layerB
...
 
   node.backgroundColor = [UIColor clearColor];
+//  [node.view setNeedsDisplay];
```
nguyenhuy: Wondering the reason behind having this line in the first place. Would that be because doing this causes the view to sync its states (opaque and background color) to the layer? Even if that's true, I think we still shouldn't rely on it because of what I said in the preview comment.

Plus, we should just remove this line.

Source/Private/_ASPendingState.mm
nguyenhuy: Confirm what Rahul said: the view doesn't propagate the background color immediately which causes many of our tests to fail because they use the layer to render snapshots. For example [this test](https://github.com/TextureGroup/Texture/blob/master/Tests/ASDisplayNodeSnapshotTests.mm#L19) failed because the parent node has a `nil` background color by the time the test takes a snapshot using [`ASSnapshotVerifyNode`](https://github.com/TextureGroup/Texture/blob/master/Tests/ASSnapshotTestCase.h#L21).

An alternative approach would be updating `ASSnapshotVerifyNode` to use the view if the node is view-backed. But by doing so, we're accepting the risk that there is a time window in which the view and the layer disagree (which we're already doing). And this time window is undocumented by UIKit.

Given that now UIColor and UIView has a new dynamic capability, and that we should rely on UIView more to take advantage of this new capability, we should update the framework to set to both the view and the layer and thus eliminate the time window and its associated risks.

Tests/ASDisplayNodeTests.mm
rahul-malik: i meant to remove this!

Tests/ASDisplayNodeTests.mm
rahul-malik: gah 2 spaces

rahul-malik: so they don't agree with each other in practice. i'll add a comment here

nguyenhuy:
I'm going to merge this PR once CI passes. We want to get this in before our sync this week so we can test it immediately.

@maicki @Adlai-Holler @appleguy Feel free to do post-reviews. @rahul-malik and I will follow up on any of your concerns in new diffs.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/4fcfe4d3eee9c61964489171919ef720/raw/1ae8907b8128d7cdc2d1cc4208c53f9b38e8d15c/log.txt)

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/b194516b9fd51600f1993856a0f90730/raw/47260b0a20936a44a7223867a00a6189bb20878e/log.txt)

PR #1590, opened 2019-07-18
add new networking image delegate method when image does not load from cache

ernestmama:
There is a missing delegate when image cannot load from cache.


Source/ASNetworkImageNode.h
```diff
@@ -172,6 +172,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)imageNodeDidLoadImageFromCache:(ASNetworkImageNode *)imageNode;
 
+/**
+ * Notification that the image node is not able to load image from cache
```
nguyenhuy:
```suggestion
 * Notification that the image node failed to load image from cache
```

PR #1592, opened 2019-07-19
Update ASPendingState backgroundColor type to UIColor

rahul-malik:
Given that UIColor is now a dynamic provider we can no longer rely on the fact that there is a 1:1 mapping between UIColor <-> CGColorRef.

I've updated the UIViewBridge and ASPendingState accordingly to use UIColor and translate to CGColorRef only when interacting with layers. This should preserve any dynamic capabilities we might need in the future when trying to re-render layer backed nodes.

Depends on https://github.com/TextureGroup/Texture/pull/1588

### DIFF
--- a/Source/Private/_ASPendingState.mm
+++ b/Source/Private/_ASPendingState.mm
@@ -16,9 +16,13 @@
 #import <AsyncDisplayKit/ASDisplayNodeInternal.h>
 #import <AsyncDisplayKit/ASInternalHelpers.h>
 
-#define __shouldSetNeedsDisplay(layer) (flags.needsDisplay \
+#define __shouldSetNeedsDisplayForView(view) (flags.needsDisplay \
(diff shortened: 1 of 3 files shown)


Source/Private/_ASPendingState.mm
```diff
@@ -18,7 +18,7 @@
 
 #define __shouldSetNeedsDisplay(layer) (flags.needsDisplay \
   || (flags.setOpaque && _flags.opaque != (layer).opaque)\
-  || (flags.setBackgroundColor && !CGColorEqualToColor(backgroundColor, (layer).backgroundColor)))
+|| (flags.setBackgroundColor && [backgroundColor isEqual:[UIColor colorWithCGColor:(layer).backgroundColor]]))
```
nguyenhuy:
```suggestion
  || (flags.setBackgroundColor && ! [backgroundColor isEqual:[UIColor colorWithCGColor:(layer).backgroundColor]]))
```

Issue #1595, opened 2019-07-23
[_ASDisplayLayer setDelegate:] is not called by UIViews


siucarlos:
UIView should call setDelegate on its backing layer to itself, which happens in iOS 12. But in iOS 13, this does not happen.
Stack in iOS 12:
> #0    0x00000001102e1124 in ::-[_ASDisplayLayer setDelegate:](id) at Texture/Source/Details/_ASDisplayLayer.mm:37
#1    0x0000000125b50923 in _UIViewLayerSetView ()
#2    0x0000000125b1538f in -[UIView _createLayerWithFrame:] ()
#3    0x0000000125b15ec3 in UIViewCommonInitWithFrame ()
#4    0x0000000125b15b2d in -[UIView initWithFrame:] ()
#5    0x0000000125b1522f in -[UIView init] ()
#6    0x0000000110124fa0 in ::-[ASDisplayNode _locked_viewToLoad]() at /Texture/Source/ASDisplayNode.mm:497
#7    0x00000001101259db in ::-[ASDisplayNode _locked_loadViewOrLayer]() at Texture/Source/ASDisplayNode.mm:564
#8    0x00000001101265c6 in ::-[ASDisplayNode view]() at /Texture/Source/ASDisplayNode.mm:634
#9    0x00000001102a5bf2 in ::-[ASViewController loadView]() at r/Texture/Source/ASViewController.mm:128

UICollectionView relies on a delegate call from its layer to trigger a relayout when the frame changes. Since the delegate is not set on _ASDisplayLayer, it never gets the callback.

Example of a stack of the delegate method 

> - (void)layer:(CALayer *)layer didChangeBoundsWithOldValue:(CGRect)oldBounds newValue:(CGRect)newBounds;

being called

> #0	0x00000001101bd02b in ::-[ASPagerNode collectionNode:constrainedSizeForItemAtIndexPath:](ASCollectionNode *, NSIndexPath *) at Texture/Source/ASPagerNode.mm:174
#1	0x00000001100d0560 in ::-[ASCollectionViewFlowLayoutInspector collectionView:constrainedSizeForNodeAtIndexPath:](ASCollectionView *, NSIndexPath *) at Texture/Source/Private/ASCollectionViewFlowLayoutInspector.mm:69
[...]

PR #1609, opened 2019-08-08
Force `-[_ASDisplayLayer setDelegate:]` to evaluate delegateFlags in iOS 13

rahul-malik:
It appears that UIKit might be setting the delegate ivar directly which bypasses establishing the `_delegateFlags` ivar on `ASDisplayLayer`. This is critical for things like ASPagerNode, ASCollectionNode that depend on layer delegateFlags to forward bounds changes.

Radar filed: FB6956185

Resolves: #1595

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -530,7 +530,6 @@ ASSynthesizeLockingMethodsWithMutex(__instanceLock__);
 - (void)_locked_loadViewOrLayer
 {
   DISABLED_ASAssertLocked(__instanceLock__);
-  
   if (_flags.layerBacked) {
     TIME_SCOPED(_debugTimeToCreateView);
     _layer = [self _locked_layerToLoad];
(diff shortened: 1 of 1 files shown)


nguyenhuy:
LGTM, thanks for fixing this, @rahul-malik!

Rahul also raised a question about whether it's still worth it to cache the result of `-respondsToSelector:` (which is what `-[_ASDisplayLayer setDelegate:]` does). IIRC, the implementation of `-respondsToSelector:` is optimized now because it uses a cache itself internally. Perhaps something worth revisiting, @appleguy @Adlai-Holler @maicki?

Adlai-Holler:
It’s a common pattern, I’d say do it if possible, but its not exactly critical except for high frequency code paths. respondsToSelector: is class_respondsTo Selector which is basically just objc_msgSend without actually invoking the function it looked up. 

In this case not critical

nguyenhuy:
Great to know, thanks @Adlai-Holler!

Source/ASDisplayNode.mm
```diff
@@ -550,6 +549,14 @@ - (void)_locked_loadViewOrLayer
     _view = [self _locked_viewToLoad];
     _view.asyncdisplaykit_node = self;
     _layer = _view.layer;
+    // Needed to force `-[_ASDisplayLayer setDelegate:]` to evaluate
+    // It appears that UIKit might be setting the delegate ivar directly which
+    // bypasses establishing the `_delegateFlags` ivar on `ASDisplayLayer`
+    // This is critical for things like ASPagerNode, ASCollectionNode that depend
+    // on layer delegateFlags to forward bounds changes
+    if (_layer.delegate == _view) {
+      [_layer setDelegate:_view];
```
filip-wag: Being that we don't own the layer, it might be a decent idea locking this to ensure that the `_layer`'s delegate didn't change a microsecond after the if-statement 🤷‍♂

PR #1617, opened 2019-08-14
Add support for tinting layer-backed ASDisplayNode

rahul-malik:
Currently we're hitting an assertion on master because `self.tintColor` is read always within `ASImageNode.mm`. We would either ignore this and only allow tinting on view backed nodes but since we're drawing the contents it feels like we can extend this capability to layer backed nodes as well.


Source/ASButtonNode.mm
```diff
@@ -133,7 +137,7 @@ -(void)tintColorDidChange
   [super tintColorDidChange];
   // UIButton documentation states that it tints the image and title of buttons when tintColor is set.
   // | "The tint color to apply to the button title and image."
-  // | From: https://developer.apple.com/documentation/uikit/uibutton/1624025-tintcolor
+  // | From: https://developerx.apple.com/documentation/uikit/uibutton/1624025-tintcolor
```
garrettmoon: intentional change?

rahul-malik: Nope!

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/0688f64040d1629b2a97e65905c93a53/raw/3ae4d290292605f223585b8169148f57c22ddb62/log.txt)

Source/ASButtonNode.mm
```diff
@@ -52,7 +52,11 @@ - (ASTextNode *)titleNode
   ASLockScopeSelf();
   if (!_titleNode) {
     _titleNode = [[ASTextNode alloc] init];
-    // Intentionally not layer-backing the image node since tintColor may be applied
+    #if TARGET_OS_IOS
+          // tvOS needs access to the underlying view
+          // of the button node to add a touch handler.
+        [_titleNode setLayerBacked:YES];
```
nguyenhuy: I know this is from the original impl, but the comment and the code don't add up. If the view is needed under tvOS, then this node should *not* be layer-backed under it (and can be layer-backed otherwise)?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -796,15 +796,44 @@ - (void)setBackgroundColor:(UIColor *)newBackgroundColor
 - (UIColor *)tintColor
 {
   _bridge_prologue_read;
-  ASDisplayNodeAssert(!_flags.layerBacked, @"Danger: this property is undefined on layer-backed nodes.");
-  return _getFromViewOnly(tintColor);
+  if (_loaded(self)) {
+    if (_flags.layerBacked) {
+        // The first nondefault tint color value in the view’s hierarchy, ascending from and starting with the view itself.
+        return _tintColor ?: self.supernode.tintColor;
```
nguyenhuy: Our locking principles don't allow walking up the hierarchy while holding the node's lock. So you'd need to do it after releasing the lock.

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -796,15 +796,44 @@ - (void)setBackgroundColor:(UIColor *)newBackgroundColor
...
 {
   _bridge_prologue_write;
-  ASDisplayNodeAssert(!_flags.layerBacked, @"Danger: this property is undefined on layer-backed nodes.");
-  _setToViewOnly(tintColor, color);
+  if (_loaded(self)) {
+    if (_flags.layerBacked) {
+      if (![_tintColor isEqual:color]) {
+        _tintColor = color;
+        // Tint color has changed. Unlock here before calling subclasses and exit-early
+        AS::MutexLocker unlock(__instanceLock__);
```
nguyenhuy: This might not be safe because the `MutexLocker` (which basically is `std::lock_guard`) will try to unlock again when it's destroyed and yield an exception. Instead I think it'd be easier to just manage the lock manually (i.e remove `_bridge_prologue_write` and this line, and call `lock`/`unlock` directly on `__instanceLock__`.

Source/ASButtonNode.mm
rahul-malik: the comment might be misleading here, I think the intention was to gate this layer backing only for iOS and by doing nothing in tvOS it will automatically be view backed?

rahul-malik: I can make this clearer at least

PR #1618, opened 2019-08-15
Remove _ASDisplayLayer's delegateDidChangeBounds flag

nguyenhuy:
We rely on the delegate setter to be called to set the flag. However, a recent change in iOS 13 (beta) causes the setter to not be called at all and thus we had to put up a workaround (https://github.com/TextureGroup/Texture/pull/1609). While the workaround works, the performance benefit that the flag provides doesn't warranty the extra complexity IMO. So I think we should just remove everything.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -549,14 +549,6 @@ ASSynthesizeLockingMethodsWithMutex(__instanceLock__);
     _view = [self _locked_viewToLoad];
     _view.asyncdisplaykit_node = self;
     _layer = _view.layer;
-    // Needed to force `-[_ASDisplayLayer setDelegate:]` to evaluate
-    // It appears that UIKit might be setting the delegate ivar directly which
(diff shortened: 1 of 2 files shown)


appleguy:
@rahul-malik @nguyenhuy could you guys also escalate this through Apple? We are doing so as well. This is a major / unnecessary iOS behavior change that increases the risk that someday an old iOS version will get broken because of the behavior deltas. This is good to have available, but ideally multiple companies are asking for the root system bug to be fixed.

rahul-malik:
@appleguy - Filed a radar a few weeks ago and will forward to our developer rep. If you have a thread you want us to chime in on let us know

maicki:
@rahul-malik Did you hear anything back yet from Apple regarding the Radar? Thank you!

rahul-malik:
@maicki - Nothing yet from Apple on this. Have you heard anything?

maicki:
No we didn't, but from our investigation it seems it was changed intentionally and as soon as you link after iOS 12 the delegate call will not be called anymore.

PR #1625, opened 2019-08-19
Update FBSnapshotTestCase to iOSSnapshotTestCase (=6.1)

rahul-malik:
Our snapshot dependency was really out-of-date. This updates us to the latest version which has support for XCTAttachments so you can view reference/failed/diff images from within Xcode!


rahul-malik:
@nguyenhuy - I noticed we have two reference directories (References_64, References_10 or something like that). Any reason we have both or should we re-record everything?

bolsinga:
> @nguyenhuy - I noticed we have two reference directories (References_64, References_10 or something like that). Any reason we have both or should we re-record everything?

https://github.com/facebookarchive/AsyncDisplayKit/pull/2295 is something that changes the output file if running against iOS 10.0 and this diff makes all tests run under iOS 10.0.

bolsinga:
> > @nguyenhuy - I noticed we have two reference directories (References_64, References_10 or something like that). Any reason we have both or should we re-record everything?
> 
> [facebookarchive/AsyncDisplayKit#2295](https://github.com/facebookarchive/AsyncDisplayKit/pull/2295) is something that changes the output file if running against iOS 10.0 and this diff makes all tests run under iOS 10.0.

Also the new snapshot lib will replace `-` in reference image filenames with `_`.

bolsinga:
This work was done in #1855.

PR #1626, opened 2019-08-20
Keep using `unowned` macro

hanton:
Keep using `unowned` macro for code consistence, instead of using `__unsafe_unretained` directly.


hanton:
hi @nguyenhuy , could you please add the `CI label` to this PR, so I can trigger the CI again?

nguyenhuy:
Can you please rebase first? I wanted to test if adding the label alone would re-trigger the pipeline.

PR #1629, opened 2019-08-21
Allow full color tinting on grayscale template images.

jessietea:
As an optimization, `imageRendererFormat` decides on a color space based on the image. Since template images are meant to be black and transparent, this means they are assigned a grayscale color space. This is not ideal, because a template image should be able to be tinted to any color, not just grayscale. In order to allow this, we can simply use the default `UIGraphicsImageRendererFormat` if the image is a template image.


Source/ASImageNode.mm
```diff
@@ -594,6 +594,14 @@ - (void)setNeedsDisplayWithCompletion:(void (^ _Nullable)(BOOL canceled))display
   [self setNeedsDisplay];
 }
 
+- (void)tintColorDidChange
+{
+  [super tintColorDidChange];
+  if (self.image.renderingMode == UIImageRenderingModeAlwaysTemplate) {
```
rahul-malik: use `_image` here instead of `self.image`

Source/Details/ASGraphicsContext.mm
```diff
@@ -47,7 +47,15 @@ NS_INLINE void ASConfigureExtendedRange(UIGraphicsImageRendererFormat *format)
 
       UIGraphicsImageRendererFormat *format;
       if (sourceImage) {
-        format = sourceImage.imageRendererFormat;
+        if (sourceImage.renderingMode == UIImageRenderingModeAlwaysTemplate) {
+          // Template images will be black and transparent, so if we use
```
rahul-malik: Nice comment!

rahul-malik:
@jessietea - Can we add a grayscale template image in the test resources directory? https://github.com/TextureGroup/Texture/tree/master/Tests/TestResources

Then we can have a new snapshot test that tints that asset with the `ASExperimentalDrawingGlobal` feature enabled

rahul-malik:
@nguyenhuy - it seems like this is stuck on waiting for Danger to report back. is there any way to get past this?

nguyenhuy:
@rahul-malik Danger stopped working because we removed the BuildKite integration. I’m trying to fix it by running Danger with GitHub Actions (#1635).

nguyenhuy:
@rahul-malik Danger is back up now.
@jessietea Can you please rebase with master to pick up the new GitHub Actions workflow? Thanks.

PR #1630, opened 2019-08-21
support dynamic images in iOS13

vovasty:
iOS13 introduced image variants for dark and light modes. This change provides support for such kind of images.

NB
shall we have a separate cache for dark/light images?


nguyenhuy:
@vovasty Please rebase with latest master to get the right CI config. Thanks.

PR #1635, opened 2019-08-22
Run Danger with GitHub Actions

nguyenhuy:
Our Danger check no longer works after we remove the BuiltKite hook. This PR attempts to run Danger with GitHub Actions.

There are 2 projects: Danger Ruby, which is what we've been using, and Danger JS. Both support GitHub Actions out of the box but the Danger Ruby action is not yet stable and would fail at build step (https://github.com/danger/danger/issues/1148). As a result, I converted our (Ruby) Dangerfile to a new dangerfile.js file and use it with Danger JS.

There are a couple of issues with the Danger JS setup which are documented in the workflow file ([here](https://github.com/TextureGroup/Texture/pull/1635/files#diff-c869aa92fe589f7f9336ffcab747610dR12)). Those are not blocking issues IMO. Danger should be back and running after this PR.

Test Plan: Tested locally (`yarn danger pr`) with multiple PRs: 788, 832, 1492, 1633, 1635, 1528, 1572 and this PR.


nguyenhuy:
All green! 🎉 

Thanks for reviewing!

PR #1637, opened 2019-08-22
Adding experiment to propagate older trait collections when trait

rahul-malik:
collections change

- Re-render nodes when user interface style changes (dark mode)
- Snapshot testing examples for dark mode
- Fixes for applying trait collection overrides in XCTest


Source/ASDisplayNode.mm
```diff
@@ -434,6 +434,18 @@ - (void)onDidLoad:(ASDisplayNodeDidLoadBlock)body
...
 }
 
+
+- (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTraitCollection)previousTraitCollection
+{
+  if (@available(iOS 13.0, *)) {
+    __instanceLock__.lock();
+    if (self.primitiveTraitCollection.userInterfaceStyle != previousTraitCollection.userInterfaceStyle) {
+      __instanceLock__.unlock();
+      [self setNeedsDisplay];
```
bolsinga: This will not unlock if they are the same.

Source/Base/ASAvailability.h
```diff
@@ -50,8 +50,13 @@
   #define __IPHONE_11_0 110000
 #endif
 
+#ifndef __IPHONE_13_0
+  #define __IPHONE_13_0 130000
+#endif
+
 #define AS_AT_LEAST_IOS10  (kCFCoreFoundationVersionNumber >= kCFCoreFoundationVersionNumber_iOS_10_0)
 #define AS_AT_LEAST_IOS11  (kCFCoreFoundationVersionNumber >= kCFCoreFoundationVersionNumber_iOS_11_0)
+#define AS_AT_LEAST_IOS13  (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0)
```
bolsinga: Will this have to change to look like the previous ones at some point?

rahul-malik: we don't think so since it seems like apple hasn't updated kCFCoreFoundationVersionNumber in a long time

Source/ASDisplayNode.mm
bolsinga: oh this will unlock when it goes out of scope. the explicit `unlock` in the `if` statement threw me.

PR #1640, opened 2019-08-23
Store backgroundColor in ASDisplayNode

rahul-malik:
We need this to preserve dynamic color capabilities for layer-backed
nodes. In layer backed nodes we lose the dynamic color logic since we
extract a CGColor from the UIColor applied to ASDisplayNode. With this
change we can respond to changes in UITraitCollection.userInterfaceStyle
for backgroundColor on layer backed nodes


Source/ASDisplayNode.mm
```diff
@@ -442,6 +442,11 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
     if (self.primitiveTraitCollection.userInterfaceStyle != previousTraitCollection.userInterfaceStyle) {
       // When changing between light and dark mode, often the entire node needs to re-render.
       // This change doesn't happen frequently so it's fairly safe to render nodes again
+      if (_loaded(self) && self.isLayerBacked && _backgroundColor) {
+        if (!CGColorEqualToColor(_layer.backgroundColor, _backgroundColor.CGColor)) {
```
bolsinga: can you get the `CGColorRef` once via a local variable?

Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -775,23 +771,24 @@ - (void)setBackgroundColor:(UIColor *)newBackgroundColor
...
   } else {
     // NOTE: If we're in the background, we cannot read the current value of bgcolor (if loaded).
     // When the pending state is applied to the view on main, we will call `setNeedsDisplay` if
     // the new background color doesn't match the one on the layer.
+    _backgroundColor = newBackgroundColor;
     ASDisplayNodeGetPendingState(self).backgroundColor = newBackgroundColor;
   }
 }
 
+
```
bolsinga: unintended whitespace?

PR #1642, opened 2019-08-25
AS::Map function improvement

hanton:
1. Prefer using `reserve function` rather than `fill constructor` in `AS::map()` for **memory efficiency**
  
`fill constructor` **constructs** the vector with **n** elements and **initializes** each element with its default constructor, while `reserve function` just allocates memory, **0** element is need to be initialized. 

For example, after running ```result_type res(iterable.size());```, the `res` vector's size is equal to `iterable.size()`, while after running ```res.reserve(iterable.size());```, the `res` vector's size is 0.

Reference:
- http://www.cplusplus.com/reference/vector/vector/vector
- http://www.cplusplus.com/reference/vector/vector/reserve
- http://www.cplusplus.com/reference/iterator/back_inserter

2. Add tests for `ASLayoutSpecUtilities`


Matthew1234567w:
K

vovasty:
Hi @hanton, could you please resolve merging conflicts, so we can merge you PR?

hanton:
> Hi @hanton, could you please resolve merging conflicts, so we can merge you PR?

hi @vovasty ,`merge conflicts` are solved, but CI fails because of this error:
```
** BUILD FAILED **
The following build commands failed:
	CompileAssetCatalog /Users/runner/ASDKDerivedData/Build/Products/Debug-iphonesimulator/Sample.app /Users/runner/runners/2.159.1/work/Texture/Texture/examples/CustomCollectionView-Swift/Sample/Assets.xcassets
(1 failure)
```

However, in my local build `./build.sh examples-pt1`, it shows `Build Succeeded`.

PR #1644, opened 2019-08-26
tests for `ASImageNode` dynamic tint color

vovasty:
no logic changes, just missing tests.


rahul-malik:
@vovasty - It seems you will either need to add new tests that aren't under iOS 13 or add the define check to overcome this issue
```objc
 /Users/runner/runners/2.157.0/work/Texture/Texture/Tests/ASImageNodeSnapshotTests.mm:14:17: unused function 'makeImageWithColor' [-Werror,-Wunused-function]

static UIImage* makeImageWithColor(UIColor *color, CGSize size) { ❌ 
```

PR #1648, opened 2019-08-28
[Docs] Fix broken link in `Layout-Quickstart`

hanton:
Quick fix for a broken link in `Layout-Quickstart` doc.


vovasty:
the link to `Layout Specs` is correct. Why do you think this is wrong?

hanton:
> the link to `Layout Specs` is correct. Why do you think this is wrong?

hi @vovasty , you're right, thanks for pointing it out. It's the `href` attribute, which is wrong.

PR #1650, opened 2019-08-28
[GitHub CI] Revert back to Ruby Danger as Danger JS is not working properly

Kaspik:
- update `danger.yml` to run as a script
- add ruby danger step to `build.sh` file and run Danger on every PR
- remove Danger JS as it's not working as it should

## Important:
Original Danger Ruby (implemented here) takes about 40sec where the new Danger JS that is right now on master takes about 4 mins.


nguyenhuy:
For reference, https://github.com/TextureGroup/Texture/pull/1635/files introduced the Danger JS file and this PR reverts it.

Kaspik:
Actually, that's a good idea. Updated PR!

.github/workflows/danger.yml
```diff
@@ -3,26 +3,19 @@ name: Danger
 on: [pull_request]
 
 jobs:
-  danger:
-    name: Danger JS
-    runs-on: ubuntu-latest
+  buildsh:
+    strategy:
+      matrix:
+        mode: [danger]
```
nguyenhuy: With only one configuration, using a matrix is a bit overkill since you can just configure the job directly (like [this](https://github.com/TextureGroup/Texture/blob/master/.github/workflows/ci-master-only.yml)). Please feel free to follow up on another PR but not a big deal otherwise.

nguyenhuy:
Btw, thanks again for fixing Danger!

Kaspik:
Thank you for merging it!
Yeah, I copied the main file. Will take a look!

PR #1651, opened 2019-08-30
use lock in asyncTraitCollectionDidChangeWithPreviousTraitCollection

vovasty:
use lock to avoid crash when image might be changed during the trait collection change


Source/ASImageNode.mm
```diff
@@ -767,6 +768,7 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
         // trigger needs display anyway, cause image might be dynamic itself (e.g. dynamic tint color)
```
rahul-malik: Need to remove this or add unlock / return here

PR #1661, opened 2019-09-06
Use lock to access the image renderingMode when tint color changes

bolsinga:
Missing with #1629. Other code accesses this ivar with a lock, so doing it here for consistency sake.

### DIFF
--- a/Source/ASImageNode.mm
+++ b/Source/ASImageNode.mm
@@ -610,7 +610,14 @@ static ASWeakMap<ASImageNodeContentsKey *, UIImage *> *cache = nil;
 - (void)tintColorDidChange
 {
   [super tintColorDidChange];
-  if (_image.renderingMode == UIImageRenderingModeAlwaysTemplate) {
+
+  BOOL isTemplateImage = NO;
+  {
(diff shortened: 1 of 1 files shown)


Source/ASImageNode.mm
```diff
@@ -610,7 +610,14 @@ - (void)setNeedsDisplayWithCompletion:(void (^ _Nullable)(BOOL canceled))display
 - (void)tintColorDidChange
 {
   [super tintColorDidChange];
-  if (_image.renderingMode == UIImageRenderingModeAlwaysTemplate) {
+
+  BOOL isTemplateImage = NO;
+  {
+    AS::MutexLocker l(__instanceLock__);
+    isTemplateImage = (_image.renderingMode == UIImageRenderingModeAlwaysTemplate);
```
rahul-malik: Should we cache this information until `self.image` is assigned a new value? `UIImage` is immutable so this value shouldn't change unless we get a new image

bolsinga: As discussed, nothing else in `ASImageNode` yet caches `UIImage` metadata, so I'd prefer not to have this diff be the first to do so.

PR #1662, opened 2019-09-06
Use MutexLocker when ASImageNode traits change

bolsinga:
Then access the ivars directly so that another lock access is not required. Using this C++ macro ensures that any future early returns or the like will still unlock.

### DIFF
--- a/Source/ASImageNode.mm
+++ b/Source/ASImageNode.mm
@@ -756,17 +756,16 @@ static ASWeakMap<ASImageNodeContentsKey *, UIImage *> *cache = nil;
   [super asyncTraitCollectionDidChangeWithPreviousTraitCollection:previousTraitCollection];
 
   if (AS_AVAILABLE_IOS_TVOS(12, 10)) {
-    __instanceLock__.lock();
+    AS::MutexLocker l(__instanceLock__);
       // update image if userInterfaceStyle was changed (dark mode)
-      if (_image != nil && self.primitiveTraitCollection.userInterfaceStyle != previousTraitCollection.userInterfaceStyle) {
-          UITraitCollection *tc = [UITraitCollection traitCollectionWithUserInterfaceStyle:self.primitiveTraitCollection.userInterfaceStyle];
+      if (_image != nil && _primitiveTraitCollection.userInterfaceStyle != previousTraitCollection.userInterfaceStyle) {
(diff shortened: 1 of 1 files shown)


Source/ASImageNode.mm
```diff
@@ -756,17 +756,16 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
...
-    __instanceLock__.lock();
+    AS::MutexLocker l(__instanceLock__);
       // update image if userInterfaceStyle was changed (dark mode)
-      if (_image != nil && self.primitiveTraitCollection.userInterfaceStyle != previousTraitCollection.userInterfaceStyle) {
-          UITraitCollection *tc = [UITraitCollection traitCollectionWithUserInterfaceStyle:self.primitiveTraitCollection.userInterfaceStyle];
+      if (_image != nil && _primitiveTraitCollection.userInterfaceStyle != previousTraitCollection.userInterfaceStyle) {
+          UITraitCollection *tc = [UITraitCollection traitCollectionWithUserInterfaceStyle:_primitiveTraitCollection.userInterfaceStyle];
           // get an updated image from asset catalog
-          UIImage *updatedImage = [self.image.imageAsset imageWithTraitCollection:tc];
+          UIImage *updatedImage = [_image.imageAsset imageWithTraitCollection:tc];
```
rahul-malik: Should we use the new method in iOS13+? Seems to have more sophisticated logic around loading image variants

https://developer.apple.com/documentation/uikit/uiimage/3294226-imagenamed?language=objc

vovasty: we can't use it in `ASImageNode` cause image is passed directly

vovasty:
I vote for accessing `imageAsset` from main thread.

Source/ASImageNode.mm
bolsinga: I think that may be good for a follow up; This PR is just about using the C++ / macro lock and then accessing ivars directly. Thoughts?

bolsinga:
> I vote for accessing `imageAsset` from main thread.

I also think that is a good follow up PR.

Source/ASImageNode.mm
bolsinga: @vovasty is right; we already have a `UIImage` in this method.

bolsinga:
> > I vote for accessing `imageAsset` from main thread.
> 
> I also think that is a good follow up PR.

And that looks like #1663

PR #1663, opened 2019-09-06
Defer image asset regeneration to draw loop

mikezucc:
There is a potential dealloc race condition that can occur when attempting to perform work with image asset out of the main thread. Moving that work into the image node draw loop if necessary

### DIFF
--- a/Source/ASImageNode.mm
+++ b/Source/ASImageNode.mm
@@ -303,11 +303,12 @@ typedef void (^ASImageNodeDrawParametersBlock)(ASWeakMapEntry *entry);
 
   if (AS_AVAILABLE_IOS_TVOS(13, 10)) {
     if (_imageNodeFlags.regenerateFromImageAsset && _image != nil) {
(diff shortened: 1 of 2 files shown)


Source/ASImageNode.mm
```diff
@@ -300,6 +301,17 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
 {
   ASLockScopeSelf();
 
+  if (AS_AVAILABLE_IOS_TVOS(13, 10)) {
+    if (_imageNodeFlags.regenerateFromImageAsset && _image != nil) {
+      _imageNodeFlags.regenerateFromImageAsset = NO;
+      UITraitCollection *tc = [UITraitCollection traitCollectionWithUserInterfaceStyle:_primitiveTraitCollection.userInterfaceStyle];
+      UIImage *updatedImage = [_image.imageAsset imageWithTraitCollection:tc];
+      if ( updatedImage != nil ) {
+        _image = updatedImage;
```
bolsinga: So we know this is called from the main thread?

mikezucc: Yes the caller of this is main only, with an assertion

bolsinga: thanks!

PR #1666, opened 2019-09-09
Node TintColor Tweaks

bolsinga:
- If the tint color changes, be sure to call -setNeedsDisplay on the view.
- Explictly declare tintColor ivar and getter method.
- If the tintColor is the same as the existing one, early-exit.
- Use the ivar and not the property when setting the tintColor, just as is done for background color.
- Add tests that the tintColor will call -setNeedsDisplay
- Backfill tests to document/verify the tintColor behavior on a node.

### DIFF
--- a/Tests/ASBridgedPropertiesTests.mm
+++ b/Tests/ASBridgedPropertiesTests.mm
@@ -145,7 +145,7 @@ static inline void ASDispatchSyncOnOtherThread(dispatch_block_t block) {
   });
 }
 
-- (void)testThatSettingTintColorSetNeedsDisplay
+- (void)testThatSettingTintColorSetNeedsDisplayOnView
 {
   ASPendingStateController *ctrl = [ASPendingStateController sharedInstance];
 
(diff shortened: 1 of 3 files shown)


Tests/ASBridgedPropertiesTests.mm
```diff
@@ -138,6 +145,24 @@ - (void)testThatReadingABridgedLayerPropertyInBackgroundThrowsAnException
...
+
+  ASDispatchSyncOnOtherThread(^{
+    node.tintColor = UIColor.orangeColor;
+  });
+  XCTAssertNotEqualObjects(view.tintColor, UIColor.orangeColor);
+  XCTAssertEqual(view.setNeedsDisplayCount, 1);
+  [ctrl flush];
+  XCTAssertEqualObjects(view.tintColor, UIColor.orangeColor);
+  XCTAssertEqual(view.setNeedsDisplayCount, 2);
+}
```
rahul-malik: Add  tests here for layer-backed nodes

Tests/ASBridgedPropertiesTests.mm
```diff
@@ -145,17 +145,19 @@ - (void)testThatReadingABridgedLayerPropertyInBackgroundThrowsAnException
...
 
   ASDisplayNode *node = [[ASDisplayNode alloc] initWithViewClass:ASBridgedPropertiesTestView.class];
   ASBridgedPropertiesTestView *view = (ASBridgedPropertiesTestView *)node.view;
   NSUInteger initialSetNeedsDisplayCount = view.setNeedsDisplayCount;
 #if AS_AT_LEAST_IOS13
-  XCTAssertEqual(initialSetNeedsDisplayCount, 2);
-#else
-  XCTAssertEqual(initialSetNeedsDisplayCount, 1);
+  if (@available(iOS 13.0, *)) {
+    XCTAssertEqual(initialSetNeedsDisplayCount, 2);
```
rahul-malik: Maybe add a comment around how this is being called multiple times?

PR #1669, opened 2019-09-12
Fix CFRelease on null variable when using truncationAttributedText

MikeDonahue:
I was hitting a crash in ASTextNode2 (in an ASCellNode) while using `truncationAttributedText` and `additionalTruncationMessage`. It seemed to only happen on a tap event when `maximumNumberOfLines` was 0, and the text was long enough to run off screen requiring the user to scroll. The node rendered fine, but crashed as the user touched down to scroll. A quick null check before calling `CFRelease` was enough to fix it.


rahul-malik:
@MikeDonahue - Thanks for the fix!

PR #764, opened 2018-01-23
Node tint color

ShogunPhyched:
I faced a small problem with tint color when working with TableView cells. If you create cell in a table view, set accessory and tint color, then tint color will not apply to cell.

```
func tableNode(_ tableNode: ASTableNode, nodeBlockForRowAt indexPath: IndexPath) -> ASCellNodeBlock {
     return {
          let cellNode = ASCellNode()
          cellNode.accessoryType = .checkmark
          cellNode.tintColor = .orange
          return cellNode
     }
} 
```
It happens because tintColor wasn't pass inside ASTableView. I fixed it so setting tintColor works fine now.

PR #1670, opened 2019-09-12
Fix main thread assertion for tint color on text nodes, re-render tint-able nodes on hierarchy changes

rahul-malik:
- Avoid accessing `self.tintColor` off-main since this will trigger an
assertion and might be undefined behavior
- Trigger setNeedsDisplay in cases of hierarchy change and
tintColorDidChange in ASButtonNode, ASTextNode{1,2}
- Remove overwriting the tintColor of _ASTableViewCell. This actually breaks the typical inheritance behavior since setting it's tint color to the ASCellNode tintColor will always return a value which will be the `systemBlueColor` if it is unable to be determined


rahul-malik:
@bolsinga - PTAL, this should be good to land now

rahul-malik:
Will definitely update the title / commit message. I did verify the tests fail without these changes

bolsinga:
> Will definitely update the title / commit message. I did verify the tests fail without these changes

Thanks for confirming!

Source/ASTableView.mm
```diff
@@ -113,12 +115,14 @@ - (void)setElement:(ASCollectionElement *)element
     self.selectionStyle = node.selectionStyle; 
     self.focusStyle = node.focusStyle;
     self.accessoryType = node.accessoryType;
-    self.tintColor = node.tintColor;
```
bolsinga: Just for code historian sake, this is removing #764

Issue #1673, opened 2019-09-13
[iOS 13] Clipping Corners are Incompatible with Dynamic Colors


shamanskyh:
The problem:
If I have an ASDisplayNode with a corner radius and corner rounding type set to `ASCornerRoundingTypeClipping`, and my ASDisplayNode has a dynamic background color, the corner layer doesn't seem to update when the trait collection's user interface style changes.

I'm going to investigate more, but I think the solution is going to be an additional call to `_updateClipCornerLayerContentsWithRadius` when the trait collection changes its interface style. Internally, we're tracking this as BUG-99048.

shamanskyh:
Fixed with #1674

PR #1674, opened 2019-09-13
#1673 - Re-render Clipping Corners when User Interface Style Changes

shamanskyh:
This PR addresses an issue I raised with #1673 where the clipping corner layer doesn't get re-rendered when going from a dark to light user interface style. The fix is to add code in the `-asyncTraitCollectionDidChangeWithPreviousTraitCollection:` method that calls through to `-_updateClipCornerLayerContentsWithRadius:backgroundColor:` if the user interface style changes.

I've also added snapshot tests to verify that using clipping corners with a dynamic background color updates when the user interface style changes.


Source/ASDisplayNode.mm
```diff
@@ -442,16 +442,29 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
...
       // This change doesn't happen frequently so it's fairly safe to render nodes again
-      if (_loaded(self) && self.isLayerBacked) {
-        // Background colors do not dynamically update for layer backed nodes since they utilize CGColorRef
-        // instead of UIColor. We utilize the _backgroundColor instance variable to track the full dynamic color
-        // and apply any changes here when trait collection updates occur.
-        CGColorRef cgBackgroundColor = _backgroundColor.CGColor;
-        if (!CGColorEqualToColor(_layer.backgroundColor, cgBackgroundColor)) {
-          _layer.backgroundColor = cgBackgroundColor;
+      BOOL needsClippingCornerUpdate = NO;
+      CGFloat cornerRadius = _cornerRadius;
```
shamanskyh: Capturing these values up here so they're behind the lock, but in scope when we call `_updateClipCornerLayerContentsWithRadius` on line 466

Tests/ASDisplayNodeSnapshotTests.mm
```diff
@@ -61,18 +61,50 @@ - (void)testPrecompositedCornerRounding
 
 - (void)testClippingCornerRounding
 {
+  self.recordMode = YES;
+#if AS_AT_LEAST_IOS13
```
shamanskyh: Looking for feedback on this inline approach. Should I delete the reference images for older Xcode/iOS versions?

bolsinga: I personally think this is fine for now.

Source/ASDisplayNode.mm
```diff
@@ -442,16 +442,29 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
...
+      ASCornerRoundingType cornerRoundingType = _cornerRoundingType;
+      UIColor *backgroundColor = self.backgroundColor;
```
rahul-malik: any reason we're using `self.backgroundColor` vs `_backgroundColor`

Source/ASDisplayNode.mm
```diff
@@ -442,16 +442,29 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
...
+      if (_loaded(self)) {
+        if (self.isLayerBacked) {
+          // Background colors do not dynamically update for layer backed nodes since they utilize CGColorRef
+          // instead of UIColor. We utilize the _backgroundColor instance variable to track the full dynamic color
+          // and apply any changes here when trait collection updates occur.
+          CGColorRef cgBackgroundColor = _backgroundColor.CGColor;
```
rahul-malik: use should use the new local variable here

Tests/ASDisplayNodeSnapshotTests.mm
```diff
@@ -61,18 +61,49 @@ - (void)testPrecompositedCornerRounding
...
+    }
+#else
     node.backgroundColor = UIColor.greenColor;
+#endif
     node.maskedCorners = c;
     node.cornerRadius = 15;
     // A layout pass is required, because that's where we lay out the clip layers.
     [node.layer layoutIfNeeded];
+
+#if AS_AT_LEAST_IOS13
```
rahul-malik: We should probably write a macro for light/dark testing if we're doing this a lot

Source/ASDisplayNode.mm
shamanskyh: `self.backgroundColor` has a getter that pulls from the layer or the view. `_backgroundColor` should only be used if layer backed.

Source/ASDisplayNode.mm
shamanskyh: Done

PR #1677, opened 2019-09-18
Use backgroundColor ivar instead of property access in traitCollectionDidChange

shamanskyh:
`traitCollectionDidChange` isn't necessarily called on Main, but our `self.backgroundColor` property access must be done on main once the view is loaded.

I originally opted to use `self.backgroundColor` rather than `_backgroundColor` in #1674, but I think we should use the ivar here. This _could_ get out-of-sync if someone modified the UIView's background color instead of updating the node, but I think that's unlikely, so we should be safe to use the ivar here.


rahul-malik:
Merged these changes in https://github.com/TextureGroup/Texture/pull/1678

PR #1691, opened 2019-09-30
use iOS 13.0 and Xcode 11 on CI

vovasty:
Xcode changed to 11
Simulator changed to iOS13 and iPhone 8
Fixed snapshot tests
Fixed tests that are depended on a separate `UIWindow`
Disabled `[ASDisplayNodeTests -testThatItIsAllowedToRetrieveDebugDescriptionIncludingVCOffMainThread]`
Weird changes to `[ASViewControllerTests -testThatAutomaticSubnodeManagementScrollViewInsetsAreApplied]` and `[ASDisplayNodeTests -checkValuesMatchDefaults]`

NB
`ASRunLoopQueueTests` became flaky


build.sh
```diff
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-PLATFORM="${TEXTURE_BUILD_PLATFORM:-platform=iOS Simulator,OS=10.2,name=iPhone 7}"
-SDK="${TEXTURE_BUILD_SDK:-iphonesimulator12.2}"
+echo !!!!
```
rahul-malik: Do we need these echo statements still or was this just for debugging?

vovasty: just debugging. There is no documentation how to switch Xcode.

Tests/ASDisplayNodeTests.mm
```diff
@@ -472,7 +472,11 @@ - (void)checkValuesMatchDefaults:(ASDisplayNode *)node isLayerBacked:(BOOL)isLay
   XCTAssertTrue(CGRectEqualToRect(CGRectZero, node.frame), @"default frame broken %@", hasLoadedView);
   XCTAssertTrue(CGPointEqualToPoint(CGPointZero, node.position), @"default position broken %@", hasLoadedView);
   XCTAssertEqual((CGFloat)0.0, node.zPosition, @"default zPosition broken %@", hasLoadedView);
-  XCTAssertEqual(1.0f, node.contentsScale, @"default contentsScale broken %@", hasLoadedView);
+  if (node.isNodeLoaded && !isLayerBacked) {
```
vovasty: is that ok?

Tests/ASViewControllerTests.mm
```diff
@@ -34,8 +34,9 @@ - (void)testThatAutomaticSubnodeManagementScrollViewInsetsAreApplied
   window.rootViewController = [[UINavigationController alloc] initWithRootViewController:vc];
   [window makeKeyAndVisible];
   [window layoutIfNeeded];
+  
   XCTAssertEqualObjects(NSStringFromCGRect(window.bounds), NSStringFromCGRect(node.frame));
-  XCTAssertNotEqual(scrollNode.view.contentInset.top, 0);
+  XCTAssertEqual(scrollNode.view.contentInset.top, 0);
```
vovasty: why it shouldn't be equal to zero?

Tests/ASDisplayNodeTests.mm
filip-wag: Tiny note: it might be a bit clearer to the reader to have this be a ternary operator instead. E.g,

```suggestion
  XCTAssertEqual(node.isNodeLoaded && !isLayerBacked ? 2.0f : 1.0f, node.contentsScale, @"default contentsScale broken %@", hasLoadedView);
```

PR #1693, opened 2019-10-01
fix a node background color in non layer baked mode.

vovasty:
This fixes problem with snapshot testing: when color is dynamic and node is not layer backed, background color remains the same.


Source/ASDisplayNode.mm
```diff
@@ -447,14 +447,20 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
...
           // Background colors do not dynamically update for layer backed nodes since they utilize CGColorRef
           // instead of UIColor. We utilize the _backgroundColor instance variable to track the full dynamic color
           // and apply any changes here when trait collection updates occur.
-          CGColorRef cgBackgroundColor = backgroundColor.CGColor;
           if (!CGColorEqualToColor(_layer.backgroundColor, cgBackgroundColor)) {
             _layer.backgroundColor = cgBackgroundColor;
           }
+        } else {
+          // we set view and layer background color in _ASPendingState, so do the same here
+          if (!CGColorEqualToColor(_layer.backgroundColor, cgBackgroundColor)) {
```
bolsinga: Is it possible to move the color equality test up, so that the code doesn't have to be here twice?

PR #1700, opened 2019-10-06
Allow setting AVPlayer

SpartanRASUL:
Hello. In my previous work, I've run into a problem, where I needed to change standard AVPlayer to AVQueuePlayer, because of it's prefetching feature. So to fix that I usually change the header file with this exact change. I think it may come in handy to other developers, who may want to use AVQueuePlayer instead of AVPlayer.


SpartanRASUL:
@jparise Yeah, really

SpartanRASUL:
i'll update this pull request soon

PR #1707, opened 2019-10-14
Fix element lookup for supplementary attributes.

craighowarth:
The `_measureElementsInRect()` function in `ASCollectionLayout.mm` was performing a `[elementForItemAtIndexPath:]` on the index path for every attribute. Some of those attributes might be supplementary views, but this distinction was being ignored. The fix will check the `representedElementKind` to see if the item is a supplementary view and if it is, `[supplementaryElementOfKind:atIndexPath:]` will be called instead.

### DIFF
--- a/Source/Private/ASCollectionLayout.mm
+++ b/Source/Private/ASCollectionLayout.mm
@@ -338,7 +338,12 @@ static const ASScrollDirection kASStaticScrollDirection = (ASScrollDirectionRigh
   if (NSUInteger count = blockingAttrs.count) {
     ASDispatchApply(count, queue, 0, ^(size_t i) {
       UICollectionViewLayoutAttributes *attrs = blockingAttrs[i];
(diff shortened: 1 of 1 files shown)


craighowarth:
I have accepted the terms for the license. Not sure why the PR hasn't approved that last step. Let me know if there is a step that I missed.

maicki:
@craighowarth Thanks for this PR and sorry that it took a couple of days. LGTM but I would also like to have eyes from @Adlai-Holler and @nguyenhuy on this.

appleguy:
Seeing this 6 months later, but wanted to thank you @craighowarth for your analysis of the code and the resulting contribution. It's sincerely appreciated!

Supplementary items were not widely utilized by the teams that depended on Texture the most as this code was built out, but it's really valuable for the community to isolate any of the faults in those codepaths.

PR #1710, opened 2019-10-22
Attempt to fix deadlock issues on iOS 13

willtr101:


Source/Details/ASDataController.mm
```diff
@@ -660,6 +661,8 @@ - (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet
       }];
     }];
     --_editingTransactionGroupCount;
+    // Leave the group
+    dispatch_group_leave(_editingTransactionGroup);
```
Adlai-Holler: I don't think this works – leaving the group after dispatching_async defeats the purpose of the group.

Adlai-Holler:
Hey @businessengine could you elaborate on the deadlock you're seeing? The current code as-written breaks the semantics of the editing transaction queue.

Source/Details/ASDataController.mm
businessengine: @Adlai-Holler Just double-checked, `dispatch_group_async` is just a convenient function, and it works the same way as `dispatch_group_enter + dispatch_async + dispatch_leave` ref: https://apple.github.io/swift-corelibs-libdispatch/tutorial/. So this change won't make any different comparing to the original. However, the change from `dispatch_group_wait` to `dispatch_group_notify` helped us solving the deadlock issues. Our QA is still in testing and we will investigate more to see why this only does happen on iOS 13, not on the previous iOS versions.

Adlai-Holler: It's not the same – dispatch_group_async is equivalent to:

```objc
group_enter()
dispatch_async(^{
  // work
  group_leave();
});
```

Whereas the code in this PR is different:

```objc
group_enter()
dispatch_async(^{
  // work
});
group_leave();
```

The group is exited before the work is finished.

Adlai-Holler: Oh you're right. So the following two things can't both be true:
- dispatch_group_async is a shorthand for this exact code. 
- Removing dispatch_group_async is necessary to fix deadlocks.

I think the first one is true. So perhaps it's switching from wait to notify that is fixing your deadlocks (seems likely.)

businessengine: Yes, switching from wait to notify fixed our deadlocks. We are confident to release our app with that change as we noticed no crashes/app being frozen issues since that fix. We have to comment an assert due to the `notify` seems work differently with the `wait`.

PR #1718, opened 2019-10-28
Change `layoutSizeDifferentFromBounds` macro to a BOOL

hanton:
Rather than defining `layoutSizeDifferentFromBounds` as a macro, a BOOL can do the job.


ernestmama:
It looks good to me but I am not sure if there is any particular reason to do it that way in the beginning. Would love some eyes from others.

PR #1722, opened 2019-11-05
Fix for images retrieved from the memory cache being reported as disk cache hits.

darrengyles:
Best as I can tell, a source type of `ASNetworkImageSourceAsynchronousCache` was meant to indicate a disk cache hit and `ASNetworkImageSourceSynchronousCache` was meant to indicate a memory cache hit. This change adds reporting of memory cache hits to the async rendering path.

### DIFF
--- a/Source/ASMultiplexImageNode.mm
+++ b/Source/ASMultiplexImageNode.mm
@@ -806,7 +806,7 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   ASDisplayNodeAssertNotNil(completionBlock, @"completionBlock is required");
 
   if (_cache) {
-    [_cache cachedImageWithURL:imageURL callbackQueue:dispatch_get_main_queue() completion:^(id <ASImageContainerProtocol> imageContainer) {
+    [_cache cachedImageWithURL:imageURL callbackQueue:dispatch_get_main_queue() completion:^(id <ASImageContainerProtocol> imageContainer, BOOL isFromMemoryCache) {
       completionBlock([imageContainer asdk_image]);
     }];
   }
--- a/Source/ASMultiplexImageNode.mm
+++ b/Source/ASMultiplexImageNode.mm
@@ -806,7 +806,7 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   ASDisplayNodeAssertNotNil(completionBlock, @"completionBlock is required");
 
   if (_cache) {
-    [_cache cachedImageWithURL:imageURL callbackQueue:dispatch_get_main_queue() completion:^(id <ASImageContainerProtocol> imageContainer, BOOL isFromMemoryCache) {
+    [_cache cachedImageWithURL:imageURL callbackQueue:dispatch_get_main_queue() completion:^(id <ASImageContainerProtocol> imageContainer, __unused ASImageCacheType cacheType ) {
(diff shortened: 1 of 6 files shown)


Source/ASMultiplexImageNode.mm
```diff
@@ -806,7 +806,7 @@ - (void)_fetchImageWithIdentifierFromCache:(id)imageIdentifier URL:(NSURL *)imag
   ASDisplayNodeAssertNotNil(completionBlock, @"completionBlock is required");
 
   if (_cache) {
-    [_cache cachedImageWithURL:imageURL callbackQueue:dispatch_get_main_queue() completion:^(id <ASImageContainerProtocol> imageContainer) {
+    [_cache cachedImageWithURL:imageURL callbackQueue:dispatch_get_main_queue() completion:^(id <ASImageContainerProtocol> imageContainer, BOOL isFromMemoryCache) {
```
jparise:
```suggestion
    [_cache cachedImageWithURL:imageURL callbackQueue:dispatch_get_main_queue() completion:^(id <ASImageContainerProtocol> imageContainer, __unused BOOL isFromMemoryCache) {
```

Source/ASNetworkImageNode.mm
```diff
@@ -888,7 +888,7 @@ - (void)_lazilyLoadImageIfNecessary
               [delegate imageNodeDidLoadImageFromCache:self];
             }
             as_log_verbose(ASImageLoadingLog(), "Decached image for %@ img: %@ url: %@", self, [imageContainer asdk_image], URL);
```
jparise: Would it be helpful to include the "is from memory" bit in this log message?

Source/ASMultiplexImageNode.mm
garrettmoon: This actually wasn't the intention originally. The protocol is in place so any library can provide image downloading and caching either synchronously or asynchronously. It wasn't intended that Texture would know how the cache is backed. I'm guessing some coupling between PINRemoteImage and Texture has crept back in…

I would recommend reporting async / sync?

darrengyles: @jparise Done

@garrettmoon Thanks for the context, have updated to use ansync/sync terminology.

Source/ASNetworkImageNode.mm
darrengyles: @jparise Done thanks.

Source/Details/ASPINRemoteImageDownloader.mm
```diff
@@ -231,15 +231,15 @@ - (void)cachedImageWithURL:(NSURL *)URL
 {
   [[self sharedPINRemoteImageManager] imageFromCacheWithURL:URL processorKey:nil options:PINRemoteImageManagerDownloadOptionsSkipDecode completion:^(PINRemoteImageManagerResult * _Nonnull result) {
     [ASPINRemoteImageDownloader _performWithCallbackQueue:callbackQueue work:^{
-      const BOOL isFromMemoryCache = result.resultType == PINRemoteImageResultTypeMemoryCache;
+      ASImageCacheType source = (result.resultType == PINRemoteImageResultTypeMemoryCache ? ASImageCacheTypeSynchronous : ASImageCacheTypeAsynchronous);
```
jparise: nit: I'd probably call this `cacheType` because that's how we've named it elsewhere.

darrengyles: Fixed, thanks.

Issue #1611, opened 2019-08-12
`ASTextNode2.mm` attributedText wrong point test range.


MapleYin:
Here is the problem I meet
if an attribute text contain a link like this:

<img width="503" alt="屏幕快照 2019-08-12 下午4 43 35" src="https://user-images.githubusercontent.com/3327822/62854035-bbddf700-bd20-11e9-9e46-e098c6cebdd8.png">

If I touch the red area, the `link` text with a Link attribute has been detected.

I notice the code below:
 https://github.com/TextureGroup/Texture/blob/1ba3c4a74f0654ed917e8b74c0ecf7088c98291f/Source/ASTextNode2.mm#L604

why use `closestTextRangeAtPoint:` instead of `textRangeAtPoint`.

alexanderkhitev:
Hey @MapleYin ! did you fix this bug?

PR #1374, opened 2019-03-06
Make ASTextNode2 more forgiving when searching for links

Adlai-Holler:
Instead of only looking at the point you actually touched, this searches in a 44x44 rect around the touch to find any links you might have hit. Similar behavior from ASTextNode1.

PR #1726, opened 2019-11-08
Update logic of searching links in ASTextNode2

dirtmelon:
Reference to #1611 description.
Related pr #1374


bolsinga:
@dirtmelon Thanks! Is it possible to add a unit test (or snapshot test) that fails w/o this fix?

Issue #1731, opened 2019-11-18
[Deadlock] ASImageNode.tintColor


muukii:
Thank you for the great library.

And I've found a deadlock problem on `ASImageNode.tintColor`.
Let me issue about this.

I'm using Texture on  `37e28300de631f38bb5420972ccc520e4e307d90`

![image](https://user-images.githubusercontent.com/1888355/69053268-649ca080-0a4c-11ea-8cd0-5eabc56d5d63.png)

muukii:
Does it relate?
https://github.com/TextureGroup/Texture/pull/1617

muukii:
Investigation Note:

I've tried to back to stable version (6176aa4bc9c3f685ee8358f078ec3d714bb02e79)
Actuallly, I need above commit hash to support iOS13.
And, this commit hash works very well.

However,  the deadlock caused after this PR #1617 merged.

So, maybe the code I wrote has some problems.
I would like to know some ideas or suggestions to avoid deadlocks.

For now, I continue using this hash (6176aa4bc9c3f685ee8358f078ec3d714bb02e79).
If I get time, I try to investigate this again.

Thanks

cc @nguyenhuy

rahul-malik:
@muukii - Can you share what's happening on the other threads? it looks like it might be occurring when trying to ascend the node hierarchy to find a tintColor

muukii:
@rahul-malik 
Thank you for your reply!

Ok! I will share the all of threads tomorrow.
I’ll be in touch!

muukii:
@rahul-malik 

I got the backtrace from lldb when it's breaking on dead-locking.
Let me know if you need other information!

```
(lldb) bt all
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00000001adecad1c libsystem_kernel.dylib`__psynch_mutexwait + 8
    frame #1: 0x00000001addeeb64 libsystem_pthread.dylib`_pthread_mutex_firstfit_lock_wait + 96
    frame #2: 0x00000001addeead0 libsystem_pthread.dylib`_pthread_mutex_firstfit_lock_slow + 264
    frame #3: 0x00000001adf521c4 libc++.1.dylib`std::__1::recursive_mutex::lock() + 16
    frame #4: 0x00000001033399a8 AsyncDisplayKit`AS::Mutex::lock(this=0x000000010dd88010) at ASThread.h:177:22
[...]

PR #1732, opened 2019-11-22
Fix tint color dead lock #1731

lkzhao:
https://github.com/TextureGroup/Texture/issues/#1731
Fix the dead lock issue caused by holding the imageNode's instance lock while walking up the view hierarchy to get the superNode's `tintColor`. If another thread is walking down the tree while holding the supernode's instance lock and trying to acquire the imageNode's lock, then we run into this dead lock.


muukii:
@lkzhao 
Hi, Thank you for fixing this issue!
I'm looking forward to using the latest version!

PR #1741, opened 2019-12-03
Remove unfair lock experiment

bolsinga:
- Experiment removed and never launched due to a crash in iOS10 in 3/2019. At PINS, this was shutdown with D378228.
- Need to import configuration header in more places when before it tagged along with ASThread.h.
- As part of this ASConfigurationInternal.h is no longer a Public header.
- Found this while debugging early use of experiments in the PINS application.


Adlai-Holler:
Interested in details on the crash if they're still available, but before removing this code it might be worth experimenting with the revised lock in #1742, which worked great for us and is now launched.

bolsinga:
> Interested in details on the crash if they're still available, but before removing this code it might be worth experimenting with the revised lock in #1742, which worked great for us and is now launched.

I don't have the crash unfortunately. I just realized this code was reading experiments at start up and nothing was being done with it for a long while. ;)

PR #1764, opened 2020-01-14
feat: make the `panGestureRecognizer` of `ASTableNode` customizable

dklinzh:
Add extensions of  the implementation of `UIGestureRecognizerDelegate` for `ASTableNode`. It would be useful as the same reason that #1457 mentioned.

### DIFF
--- a/Source/ASTableNode.mm
+++ b/Source/ASTableNode.mm
@@ -892,3 +892,39 @@ ASLayoutElementCollectionTableSetTraitCollection(_environmentStateLock)
 }
 
 @end
+
+#pragma mark - UIGestureRecognizerDelegate
+
+@implementation ASTableNode (UIGestureRecognizerDelegate)
+
+- (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
+{
+  return [super gestureRecognizerShouldBegin:gestureRecognizer];
+}
+
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceiveTouch:(UITouch *)touch
+{
+  return YES;
+}
+
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceivePress:(UIPress *)press
+{
+  return YES;
+}
+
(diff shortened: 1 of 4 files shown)


Source/ASTableNode.mm
```diff
@@ -892,3 +892,39 @@ - (void)waitUntilAllUpdatesAreCommitted
...
 
 @end
+
+#pragma mark - UIGestureRecognizerDelegate
+
+@implementation ASTableNode (UIGestureRecognizerDelegate)
+
+- (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
+{
+  return [super gestureRecognizerShouldBegin:gestureRecognizer];
```
garrettmoon, line 902: Looks like this is different than the behavior in [ASCollectionNode](https://github.com/TextureGroup/Texture/pull/1457/files#diff-d505487f1d5096efe0b8a6cd4ae539ceR1047). Should we change the behavior on ASCollectionNode or here?

garrettmoon: Might be worth copying over the comment in ASCollectionNode too?

Source/ASTableView.h
```diff
@@ -37,6 +37,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface ASTableView (UIGestureRecognizerDelegate) <UIGestureRecognizerDelegate>
```
garrettmoon, line 40: Any reason to put this in a category instead of just having conformance above?

dklinzh: I think using category would make the structure of `UIGestureRecognizerDelegate` implementations more explicit and distinct as an extension from class `ASTableView `. It is similar to the Swift syntax [Adding Protocol Conformance with an Extension](https://docs.swift.org/swift-book/LanguageGuide/Protocols.html#ID277) 🤔

Source/ASTableNode.mm
dklinzh: As the subclass of `ASDisplayNode`, it should be supervised by the implementation of `-[gestureRecognizerShouldBegin:]` from its superclass. Is that applicable?

PR #1770, opened 2020-02-04
Prevent crashing during non critical logging at rotation

foxware00:
As mentioned in the [Slack channel](https://asyncdisplaykit.slack.com/archives/C0V63R86T/p1578575412003700). 

I receive a crash during device rotation of certain screens. It happens when logging out new traits inside the `NSStringFromASPrimitiveTraitCollection` method during the  `ASViewController.mm#propagateNewTraitCollection` calls. 

The controller in question is an old UIStoryboard, I'm not sure of the reason for the nil `preferredContentSizeCategory` being nil, but given it's just a log, it shouldn't be crashing the app.

This PR simply adds a null check to the String conversion method, I'm not sure if someone can shed more light on whether or not this is appropriate, or the root cause might be cause for concern? @Adlai-Holler @nguyenhuy

### DIFF
--- a/Source/Details/ASTraitCollection.mm
+++ b/Source/Details/ASTraitCollection.mm
@@ -244,7 +244,9 @@ NSString *NSStringFromASPrimitiveTraitCollection(ASPrimitiveTraitCollection trai
   }
   if (AS_AVAILABLE_IOS(10)) {
     [props addObject:@{ @"layoutDirection": AS_NSStringFromUITraitEnvironmentLayoutDirection(traits.layoutDirection) }];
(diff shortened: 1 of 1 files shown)


nguyenhuy:
To answer your questions, I think the null check is appropriate because the log shouldn't crash the app. And it's weird that you didn't hit assertions around that property (`ASDisplayNodeCAssertPermanent`).

I can't say for sure why the property is nil. Perhaps that's because the old storyboard causes the construction of a trait collection (perhaps on an old iOS version?) that doesn't set a value to that property, or did you construct it yourself? Either way, you should be able to tell by breaking at that line and trace back the origin of that struct.

PR #1777, opened 2020-02-20
resolve background color using system trait collection

vovasty:
We should resolve color using node's trait collection, but Texture changes it from many places, so node may receive the wrong one.
 This is a quick fix. We should revise trait collection propagation mechanism.


nguyenhuy:
CI is broken for some reasons. In previous runs, [everything](https://github.com/TextureGroup/Texture/actions/runs/42652165) [passed](https://github.com/TextureGroup/Texture/actions/runs/42652162) excepts Danger. Gonna land this diff now and figure out the CI issues later.

Issue #1723, opened 2019-11-06
ASCollectionNode row selection/touch not working when Inverted is set to true.


varyP:
Usecase : Using IGListKit + Texture

Issue : 
If ASCollectionNode's .inverted is true, the selection callback for each row didSelectRowAtIndex never gets called! 
For now switching to :branch => '2.8.1' solves the issue.
The issue can be reproduced on `master`.

Easily reproducible across all OS versions.

PS : If it matters, using `UICollectionViewFlowLayout`.

PR #1781, opened 2020-02-27
Fix hit point when ASCollectionNode inverted set to true

### LINKED ISSUE #1763: ASDisplayNode vertical hit point is incorrect inside ASCollectionNode when inverted set to true
I have an inverted ASCollectionNode (`_collectionNode.inverted = YES;`). My collection node items contain some ASDisplayNodes. The tap targets we have setup on these nodes have incorrect (possibly inverted) hit areas now. This issue is not present in 2.7.0, but it is present in 2.8.1 (I did not test 2.8 explicitly but I assume the issue was introduced there as 2.8.1 was a small release).

Example: 
I overrode this method in my ASDisplayNode subclass
```
- (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event {
  NSLog(@"point: x(%f), y(%f)", point.x, point.y);
  return [super hitTest:point withEvent:event];
}
```

When tapping an area of the node in 2.7.0, I see ~(x: 16, y: 13). In 2.8.1 I now see ~(x: 16, y: 83) when tapping the same approximate area of the node. 

---
This might be related to: https://github.com/TextureGroup/Texture/issues/1723

bdolman:
This bug was introduced in commit 86a853e in PR #1268 because it was not accounting for the transform that gets set on the node when the `inverted` flag is set to `true` on the ASCollectionNode.

This fixes #1763 and possibly #1723.


Source/Details/_ASCollectionViewCell.mm
```diff
@@ -106,7 +106,8 @@ - (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event
    * superclass hitTest:withEvent: implementation. If this returns a valid value we can go on with
    * checking the node as it's expected to not be in one of these states.
    */
-  if (![super hitTest:self.bounds.origin withEvent:event]) {
+  CGPoint originPointOnView = [self convertPoint:self.node.view.bounds.origin fromView:self.node.view];
```
nguyenhuy, line 109: Nit: For readability and a tiny tiny bit of perf, please store `self.node` and `self.node.view` in variables and use them throughout the method.

bolsinga:
@bdolman Would you be able to add a test as well as re-trigger a build? CI has been flakey, and we did a little work to get it better. Thanks!

drinkius:
@bdolman Hi Ben! Do you still work on this fix or should I create a new PR for this to fix conflicts?

bdolman:
Apologies for leaving this stale for a while. I'm back to getting this fixed up. My last task is to add a test, so I'll see if I can figure that out today. Any tips would be appreciated though!

bdolman:
@bolsinga I've added a test and verified that it fails before my fix and succeeds after my fix. I'm not sure what's going on with CI though -- it all seems unrelated to my change.

bdolman:
@bolsinga @nguyenhuy This has a test and is now passing all checks. I think this is ready for your final review and a merge.

bolsinga (approved):

bolsinga:
> @bolsinga @nguyenhuy This has a test and is now passing all checks. I think this is ready for your final review and a merge.

Thanks for the ping, @bdolman!

PR #1798, opened 2020-04-03
Fix node using wrong color with ASTraitCollection when using overrideUserInterfaceStyle

IvanChan:
In UIKit, you can set `overrideUserInterfaceStyle` to **UIView/UIViewController** forcing them to display in specific mode

`ASDisplayNode` init with `userInterfaceStyle  = .unspecific`, and only update **userInterfaceStyle** when UIView call `traitCollectionDidChange`, it will miss UIView's default value of `traitCollection.userInterfaceStyle ` 

Another change will be fixing colors using in drawing part of `ASDisplayNode` & `ASImageNode`  in case `overrideUserInterfaceStyle` feature is NOT pass through

Finally, just forwarding iOS13 new api to outsider

### DIFF
--- a/Source/ASCollectionViewProtocols.h
+++ b/Source/ASCollectionViewProtocols.h
@@ -95,6 +95,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (BOOL)collectionView:(UICollectionView *)collectionView canPerformAction:(SEL)action forItemAtIndexPath:(NSIndexPath *)indexPath withSender:(nullable id)sender ASDISPLAYNODE_DEPRECATED_MSG("Implement collectionNode:canPerformAction:forItemAtIndexPath:withSender: instead.");
 - (void)collectionView:(UICollectionView *)collectionView performAction:(SEL)action forItemAtIndexPath:(NSIndexPath *)indexPath withSender:(nullable id)sender ASDISPLAYNODE_DEPRECATED_MSG("Implement collectionNode:performAction:forItemAtIndexPath:withSender: instead.");
 
+- (nullable UIContextMenuConfiguration *)collectionView:(UICollectionView *)collectionView contextMenuConfigurationForItemAtIndexPath:(NSIndexPath *)indexPath point:(CGPoint)point API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
+
+- (nullable UITargetedPreview *)collectionView:(UICollectionView *)collectionView previewForHighlightingContextMenuWithConfiguration:(UIContextMenuConfiguration *)configuration API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
+
+- (nullable UITargetedPreview *)collectionView:(UICollectionView *)collectionView previewForDismissingContextMenuWithConfiguration:(UIContextMenuConfiguration *)configuration API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
+
+- (void)collectionView:(UICollectionView *)collectionView willPerformPreviewActionForMenuWithConfiguration:(UIContextMenuConfiguration *)configuration animator:(id<UIContextMenuInteractionCommitAnimating>)animator API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -453,6 +453,11 @@ ASSynthesizeLockingMethodsWithMutex(__instanceLock__);
           CGFloat cornerRadius = _cornerRadius;
           ASCornerRoundingType cornerRoundingType = _cornerRoundingType;
           UIColor *backgroundColor = _backgroundColor;
+          if (@available(iOS 13.0, *)) {
+             UITraitCollection *tempTraitCollection = [UITraitCollection traitCollectionWithUserInterfaceStyle:_primitiveTraitCollection.userInterfaceStyle];
+             backgroundColor = [backgroundColor resolvedColorWithTraitCollection:tempTraitCollection];
+          }
+         
           __instanceLock__.unlock();
           // TODO: we should resolve color using node's trait collection
           // but Texture changes it from many places, so we may receive the wrong one.
@@ -590,7 +595,10 @@ ASSynthesizeLockingMethodsWithMutex(__instanceLock__);
   } else {
     TIME_SCOPED(_debugTimeToCreateView);
     _view = [self _locked_viewToLoad];
-    _primitiveTraitCollection = ASPrimitiveTraitCollectionFromUITraitCollection(_view.traitCollection);
+    if ([self supernode] == nil) {
+      // Move to supernode wil propagateDown other way
+        _primitiveTraitCollection = ASPrimitiveTraitCollectionFromUITraitCollection(_view.traitCollection);
+    }
     _view.asyncdisplaykit_node = self;
     _layer = _view.layer;
   }
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -590,6 +590,7 @@ ASSynthesizeLockingMethodsWithMutex(__instanceLock__);
   } else {
     TIME_SCOPED(_debugTimeToCreateView);
     _view = [self _locked_viewToLoad];
+    _primitiveTraitCollection = ASPrimitiveTraitCollectionFromUITraitCollection(_view.traitCollection);
     _view.asyncdisplaykit_node = self;
     _layer = _view.layer;
   }
--- a/Source/ASImageNode.mm
+++ b/Source/ASImageNode.mm
@@ -525,9 +525,6 @@ static ASWeakMap<ASImageNodeContentsKey *, UIImage *> *cache = nil;
       UITraitCollection *tempTraitCollection = [UITraitCollection traitCollectionWithUserInterfaceStyle:key.userInterfaceStyle];
       backgroundColor = [backgroundColor resolvedColorWithTraitCollection:tempTraitCollection];
       tintColor = [tintColor resolvedColorWithTraitCollection:tempTraitCollection];
-
-    } else {
-      // Fallback on earlier versions
     }
 
     // if view is opaque, fill the context with background color
--- a/Source/ASImageNode.mm
+++ b/Source/ASImageNode.mm
@@ -517,10 +517,22 @@ static ASWeakMap<ASImageNodeContentsKey *, UIImage *> *cache = nil;
       key.willDisplayNodeContentWithRenderingContext(context, drawParameters);
       contextIsClean = NO;
     }
+    
+    UIColor *backgroundColor = key.backgroundColor;
+    UIColor *tintColor = key.tintColor;
+    
+    if (@available(iOS 13.0, *)) {
+      UITraitCollection *tempTraitCollection = [UITraitCollection traitCollectionWithUserInterfaceStyle:key.userInterfaceStyle];
+      backgroundColor = [backgroundColor resolvedColorWithTraitCollection:tempTraitCollection];
+      tintColor = [tintColor resolvedColorWithTraitCollection:tempTraitCollection];
+
+    } else {
+      // Fallback on earlier versions
+    }
 
     // if view is opaque, fill the context with background color
-    if (key.isOpaque && key.backgroundColor) {
-      [key.backgroundColor setFill];
+    if (key.isOpaque && backgroundColor) {
+      [backgroundColor setFill];
       UIRectFill({ .size = key.backingSize });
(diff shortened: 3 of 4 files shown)


IvanChan:
@jparise @vovasty 

Fix -[ASNetworkImageNode setURL:resetToDefault:] forget to reset animatedImage	

Related to issue:  https://github.com/TextureGroup/Texture/issues/1815

Source/ASDisplayNode.mm
```diff
@@ -453,6 +453,11 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
           CGFloat cornerRadius = _cornerRadius;
           ASCornerRoundingType cornerRoundingType = _cornerRoundingType;
           UIColor *backgroundColor = _backgroundColor;
+          if (@available(iOS 13.0, *)) {
+             UITraitCollection *tempTraitCollection = [UITraitCollection traitCollectionWithUserInterfaceStyle:_primitiveTraitCollection.userInterfaceStyle];
```
vovasty, line 457: in general this is a valid change, but we found that in some cases we have a discrepancy between texture and system `userInterfaceStyle`. E.g. when you background the app, it changes  `userInterfaceStyle` back and forth for snapshotting.

Source/ASImageNode.mm
```diff
@@ -517,10 +517,19 @@ + (UIImage *)createContentsForkey:(ASImageNodeContentsKey *)key drawParameters:(
       key.willDisplayNodeContentWithRenderingContext(context, drawParameters);
       contextIsClean = NO;
     }
+    
+    UIColor *backgroundColor = key.backgroundColor;
+    UIColor *tintColor = key.tintColor;
+    
+    if (@available(iOS 13.0, *)) {
+      UITraitCollection *tempTraitCollection = [UITraitCollection traitCollectionWithUserInterfaceStyle:key.userInterfaceStyle];
```
vovasty, line 525: `ASGraphicsCreateImageWithTraitCollectionAndOptions` runs drawing code inside `[UITraitCollection -performAsCurrentTraitCollection]` thus we don't need to resolve `tintColor` here

Source/ASDisplayNode.mm
```diff
@@ -590,6 +595,10 @@ - (void)_locked_loadViewOrLayer
   } else {
     TIME_SCOPED(_debugTimeToCreateView);
     _view = [self _locked_viewToLoad];
+    if ([self supernode] == nil) {
+      // Move to supernode wil propagateDown other way
+        _primitiveTraitCollection = ASPrimitiveTraitCollectionFromUITraitCollection(_view.traitCollection);
```
vovasty, line 600: this will set `traitCollection` into system default value (the same as unspecified), right?

IvanChan: ASDisplayNode init with userInterfaceStyle = .unspecific, and only update userInterfaceStyle when UIView call traitCollectionDidChange, it will miss UIView's default value of traitCollection.userInterfaceStyle

Source/ASImageNode.mm
IvanChan: If you don't resolve ,  ASDisplayNode & ASImageNode will use wrong color, when overrideUserInterfaceStyle is set

IvanChan:
@vovasty 

1. Put your phone in light-mode & launch the demo
2. You will see UIKit-View & ASDisplayNode render in different color

[Demo.zip](https://github.com/TextureGroup/Texture/files/4602316/Demo.zip)

IvanChan:
Any news? Or better solutions? @vovasty @jparise

vovasty:
Hi @IvanChan,  sorry for the delay. I've tried Demo. If you inherit `TestViewController` from `ASViewController` it  works as expected.
Trying to understand why it like this

vovasty:
problem comes from `[ASDisplayNode _locked_applyPendingViewState]`: it sets layer's background color to a resolved value using system trait collection. For some reason it happens only if `UIViewController ` is used, `ASViewController` is ok.
Workaround is to use `ASViewController`.
Proposed solution looks too hacky for me. @nguyenhuy thoughts?

IvanChan:
Deep in color resolve logic in `UIView`, it should works with `overrideUserInterfaceStyle`

And `ASDisplayNode` has its own color resolve logic and miss  `overrideUserInterfaceStyle` which is NOT synchronized with its *View*

What I did is add the missing logic to node

vovasty:
My concern is that we have strange synchronization logic that cures symptom but not the cause.
I think we need to understand why  ASViewController works, but UIViewController doesn't. Your Demo project gives a good demonstration of this problem.

IvanChan:
> My concern is that we have strange synchronization logic that cures symptom but not the cause.
> I think we need to understand why ASViewController works, but UIViewController doesn't. Your Demo project gives a good demonstration of this problem.

Clearly `ASViewController ` has more opportunity to sync **TraitCollection** from UIKit

Any why  `ASViewController` works just because when `overrideUserInterfaceStyle` is set, `UIViewController` will call `- (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection`,  thus `ASViewController` get a chance to update  **TraitCollection** to right one, that's advantage when using ViewController

But we just miss that chance, when just using **Node** instead of **ViewController**

Source/ASDisplayNode.mm
```diff
@@ -453,6 +453,11 @@ - (void)asyncTraitCollectionDidChangeWithPreviousTraitCollection:(ASPrimitiveTra
           CGFloat cornerRadius = _cornerRadius;
           ASCornerRoundingType cornerRoundingType = _cornerRoundingType;
           UIColor *backgroundColor = _backgroundColor;
+          if (@available(iOS 13.0, *)) {
+             UITraitCollection *tempTraitCollection = [UITraitCollection traitCollectionWithUserInterfaceStyle:_primitiveTraitCollection.userInterfaceStyle];
+             backgroundColor = [backgroundColor resolvedColorWithTraitCollection:tempTraitCollection];
```
nguyenhuy, line 458: Might be a stupid question, but why can't we resolve the color using `_primitiveTraitCollection` here? What makes it different from a new one created from its user interface style and why?

Source/ASCollectionViewProtocols.h
```diff
@@ -95,6 +95,13 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+- (nullable UIContextMenuConfiguration *)collectionView:(UICollectionView *)collectionView contextMenuConfigurationForItemAtIndexPath:(NSIndexPath *)indexPath point:(CGPoint)point API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
+
+- (nullable UITargetedPreview *)collectionView:(UICollectionView *)collectionView previewForHighlightingContextMenuWithConfiguration:(UIContextMenuConfiguration *)configuration API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
+
+- (nullable UITargetedPreview *)collectionView:(UICollectionView *)collectionView previewForDismissingContextMenuWithConfiguration:(UIContextMenuConfiguration *)configuration API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
+
+- (void)collectionView:(UICollectionView *)collectionView willPerformPreviewActionForMenuWithConfiguration:(UIContextMenuConfiguration *)configuration animator:(id<UIContextMenuInteractionCommitAnimating>)animator API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(watchos, tvos);
```
nguyenhuy, line 104: Please move these changes into their own pull request.

Source/ASDisplayNode.mm
nguyenhuy: Please explain why you decided to change the ivar directly instead of calling the setter (i.e `-setPrimitiveTraitCollection:`) which in turn calls `-asyncTraitCollectionDidChangeWithPreviousTraitCollection` if the values are different.

Source/ASDisplayNode.mm
```diff
@@ -590,6 +595,10 @@ - (void)_locked_loadViewOrLayer
   } else {
     TIME_SCOPED(_debugTimeToCreateView);
     _view = [self _locked_viewToLoad];
+    if ([self supernode] == nil) {
+      // Move to supernode wil propagateDown other way
```
nguyenhuy, line 599: I assume this comment tries to explain why we need the super node check above, but it isn't clear to me. Can you elaborate?

Source/ASNetworkImageNode.mm
```diff
@@ -216,6 +216,7 @@ - (void)setURL:(NSURL *)URL resetToDefault:(BOOL)reset
     if (reset || hadURL) {
       [self _setCurrentImageQuality:(hadURL ? 0.0 : 1.0)];
       [self _locked__setImage:_defaultImage];
+      [self _locked_setAnimatedImage:nil];
```
nguyenhuy, line 219: Please move this change into its own pull request.

nguyenhuy:
@IvanChan Thanks for providing the details and sample projects. As I understand it, the main issue this pull request tries to address is this:

> Case 2:
> 
>     init a UIViewController & set UIViewController overrideUserInterfaceStyle
>     add a UIView to UIViewController
>     This UIView won‘t get the same callback, deep inside UIView (I don't know whether is apple's bug), UIView just take it as initial value, no callback of traitCollectionDidChange
> 
> We now are facing Case 2, and got to find a way to sync TraitCollection from its initial value, because there is no callback tell you that TraitCollection is changed.

Essentially, the root view of the UIViewController has a different value rather than `UIUserInterfaceStyleUnspecified` and its node doesn't pick that up. That is a valid issue to me. There are 2 solutions I can think of:
1. Like what you proposed in this pull request: right after the node loads its view, it should ask the view for its interface style and update if it gets a value that is different from what it currently has (most likely `UIUserInterfaceStyleUnspecified`).
2. In your code, set the override interface style to the node's view right after it's loaded (i.e in `-viewDidLoad`).

The second approach doesn't require any changes in the framework, but since I believe this is a framework issue, I'm leaning toward the first one.

As for this pull request, I have a few asks:
- Please break the changes in this pull request into multiple ones. This pull request should only be about the issue described above. That way, it'll be easier for us to review, as well as manage after it's merged.
- Please consider adding a new unit test to ensure the issue won't occur again in the future. The steps you described above are very clear and the test can just follow them.

Source/ASImageNode.mm
nguyenhuy: What happens if `overrideUserInterfaceStyle` is correctly picked up by the node? Will we still need this change? And why?

Source/ASDisplayNode.mm
nguyenhuy: If I understand correctly, what you want to check here is whether this node is the root node within a view controller. If that's the case, you can simply check `node.isViewControllerRoot` is what you want. This flag is set automatically by ASViewController. If you don't use ASViewController, you need to set it yourselves.

Source/ASCollectionViewProtocols.h
nguyenhuy: https://github.com/TextureGroup/Texture/pull/1817/files

nguyenhuy:
Question regarding the below issue:

> Actually ASViewController also has same problem
> 
> If you check out Demo22.zip
> 
> You will find out problem still there even though using ASViewController, when ViewController has child viewController, user interface style is NOT pass through
> 

Can you please confirm that the issue doesn't exist with UIViewController? That is, if the parent and child view controllers are UIViewControllers, the `overrideUserInterfaceStyle` of the parent VC will be passed through to the child VC and its view?

Source/ASDisplayNode.mm
IvanChan: resolvedColorWithTraitCollection need a UITraitCollection class

Source/ASDisplayNode.mm
IvanChan: In some way, I use UIViewController and add subnodes later, these nodes will face the same problem, thus `_flags.isViewControllerRoot` might not be the best choice.

change the ivar directly because I wanted less logic during `_locked_loadViewOrLayer` which might have less side effect for new code. And I debug it find out here is quite early like init, and this value updated will propagate down later.
  
 `setPrimitiveTraitCollection` also works and it's little bit heavy I think.

IvanChan:
At first, sorry for combine different changed here :(
I will pull a new request and link here later.

UIView works well at the same case in demo.

>In your code, set the override interface style to the node's view right after it's loaded (i.e in -viewDidLoad).

I think it's unpractical, If I added 10 sub nodes to my UIViewController's View, I have to find them all out and set override interface style to them.

IvanChan:
moved to https://github.com/TextureGroup/Texture/pull/1862
@nguyenhuy @vovasty @jparise

jparise:
@IvanChan should we close this issue then?

IvanChan:
> @IvanChan should we close this issue then?

OK I will close this one.

@nguyenhuy should review the new one then

PR #1807, opened 2020-04-22
[ASDisplayNode] Allow explicit setting of accessibilityElements

rcancro:
Since `NSObject` conforms to the informal accessibility protocol, `ASDisplayNode` has an inherited property for `accessibilityElements`. However, setting this property has no effect since `ASDisplayNode` overrides the `accessibilityElements` getter. Added a small change to the getter to check if the `accessibilityElements` property on `ASDisplayNode` has been explicitly set, and if so return that.

I also added a comment around  `_ASDisplayView`’s `setAccessibilityElements:` method to clear up some (of my own) confusion.

### DIFF
--- a/Source/Details/_ASDisplayViewAccessiblity.mm
+++ b/Source/Details/_ASDisplayViewAccessiblity.mm
@@ -259,6 +259,10 @@ static void CollectAccessibilityElements(ASDisplayNode *node, NSMutableArray *el
 - (void)setAccessibilityElements:(NSArray *)accessibilityElements
 {
   ASDisplayNodeAssertMainThread();
+  // While it looks very strange to ignore the accessibilyElements param and set _accessibilityElements to nil, it is actually on purpose.
+  // _ASDisplayView's accessibilityElements method will always defer to the node for accessibilityElements when _accessibilityElements is
+  // nil. Calling setAccessibilityElements on _ASDisplayView is basically us clearing the cache and forcing _ASDisplayView to ask the node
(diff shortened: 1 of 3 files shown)


Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -259,6 +259,10 @@ @implementation _ASDisplayView (UIAccessibilityContainer)
 - (void)setAccessibilityElements:(NSArray *)accessibilityElements
 {
   ASDisplayNodeAssertMainThread();
+  // While it looks very strange to ignore the accessibilyElements param and set _accessibilityElements to nil, it is actually on purpose.
+  // _ASDisplayView's accessibilityElements method will always defer to the node for accessibilityElements when _accessibilityElements is
+  // nil. Calling setAccessibilityElements on _ASDisplayView is basically us clearing the cache and forcing _ASDisplayView to ask the node
```
vovasty, line 264: nit: extra us?

nguyenhuy:
Our CI pipeline has been a bit flaky lately. `testThatViewControllerFrameIsRightAfterCustomTransitionWithNonextendedEdges` fails about 20% of the time, GH's "re-run jobs" button would fail 100% of the time, and some jobs (like examples-pt4 in this PR) would fail without any details. @rcancro ran examples-pt4 locally and didn't hit any issues. I'm going to land this diff as is. Those flakiness issues should be looked at later on.

PR #1811, opened 2020-04-27
Nil contents on zero bounds

wsdwsd0829:
e.g. on setting nil to text node.
Following would stop text node from clearing its content.
https://github.com/TextureGroup/Texture/blob/master/Source/Private/ASDisplayNode%2BAsyncDisplay.mm#L192


vovasty:
@wsdwsd0829 could you please squash commits into one?

bolsinga:
@wsdwsd0829 Is it possible to add a unit tests that fails w/o this fix, and passes w/ it?

PR #1812, opened 2020-05-05
[ASDisplayViewAccessibility] A few accessibility improvements

rcancro:
This diff includes a few improvements for accessibility in Texture.

* When determining a node’s accessibility elements, ignore any elements that are hidden, transparent, or out of the window. This matches UIKit’s behavior.
* When sorting the accessible elements and their origins are equal, give precedence to elements with shorter accessibility frames, followed by elements with narrower widths.
* Allow the ability to customize the comparator block that sorts the accessibility elements
* Create an experiment to stop caching accessibilityElements in `_ASDisplayView`. If we cache elements, we will require users to clear the cache (by calling `setAccessibilityElements` to get the side effect of clearing the cache) when nodes change their hidden state or opacity. This seems like a lot to request of the user. We will put this in an experiment so we can see the perf implication is both when using voice over and when not using voice over.

A few other notes:
* I got rid of the `ASAccessibilityElementPositioning` protocol in favor of passing `NSObjects` to the sort comparator. `NSObject` implements the informal `UIAccessibility` protocol and therefore has an `accessibilityFrame` property.
* I removed `static` from the `SortAccessibilityElements()` method definition. This allows me to declare it as `extern` in test it via unit tests.

### DIFF
--- a/Source/Details/_ASDisplayViewAccessiblity.h
+++ b/Source/Details/_ASDisplayViewAccessiblity.h
@@ -23,7 +23,7 @@
 //
 // In general this seems to work fairly well. However, if you want to provide a custom sort you can do so via
 // setUserDefinedComparator(). The two elements you are comparing are NSObjects, which conforms to the informal
-// UIAccessibilityProtocol, so you can safely compare properties like accessibilityFrame.
+// UIAccessibility protocol, so you can safely compare properties like accessibilityFrame.
 typedef NSComparisonResult (^ASSortAccessibilityElementsComparator)(NSObject *, NSObject *);
 
 // Use this method to supply your own custom sort comparator used to determine the order of the accessibility elements
--- a/Source/Details/_ASDisplayViewAccessiblity.h
+++ b/Source/Details/_ASDisplayViewAccessiblity.h
@@ -14,3 +14,17 @@
 // should still work as long as accessibility is enabled, this framework provides no guarantees on
 // their correctness. For details, see
 // https://developer.apple.com/documentation/objectivec/nsobject/1615147-accessibilityelements
+
+// After recusively collecting all of the accessibility elements of a node, they get sorted. This sort determines
+// the order that a screen reader will traverse the elements. By default, we sort these elements based on their
+// origin: lower y origin comes first, then lower x origin. If 2 nodes have an equal origin, the node with the smaller
+// height is placed before the node with the smaller width. If two nodes have the exact same rect, we throw up our hands
+// and return NSOrderedSame.
+//
+// In general this seems to work fairly well. However, if you want to provide a custom sort you can do so via
+// setUserDefinedComparator(). The two elements you are comparing are NSObjects, which conforms to the informal
+// UIAccessibilityProtocol, so you can safely compare properties like accessibilityFrame.
+typedef NSComparisonResult (^ASSortAccessibilityElementsComparator)(NSObject *, NSObject *);
+
+// Use this method to supply your own custom sort comparator used to determine the order of the accessibility elements
+void setUserDefinedAccessibilitySortComparator(ASSortAccessibilityElementsComparator userDefinedComparator);
--- a/Source/Details/_ASDisplayViewAccessiblity.h
+++ b/Source/Details/_ASDisplayViewAccessiblity.h
@@ -22,8 +22,8 @@
 // and return NSOrderedSame.
 //
 // In general this seems to work fairly well. However, if you want to provide a custom sort you can do so via
-// setUserDefinedComparator(). The two elements you are comparing are NSObjects, which conforms to the informal
-// UIAccessibility protocol, so you can safely compare properties like accessibilityFrame.
+// setUserDefinedAccessibilitySortComparator(). The two elements you are comparing are NSObjects, which conforms to the
+// informal UIAccessibility protocol, so you can safely compare properties like accessibilityFrame.
 typedef NSComparisonResult (^ASSortAccessibilityElementsComparator)(NSObject *, NSObject *);
 
 // Use this method to supply your own custom sort comparator used to determine the order of the accessibility elements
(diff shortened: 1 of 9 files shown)


Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -21,43 +22,54 @@
...
+      return (originA.y < originB.y) ? NSOrderedAscending : NSOrderedDescending;
+    };
   });
-  [elements sortUsingComparator:comparator];
+  
+  if (_userDefinedComparator) {
+    [elements sortUsingComparator:_userDefinedComparator];
+  } else {
+    [elements sortUsingComparator:defaultComparator];
+  }
```
jparise, line 64: There isn't a way to switch back to the default comparator. You could avoid this conditional by going back to the single `static` variable (`comparator`) that is initially set to the default implementation (using the `dispatch_once` block) but that can be reassigned (customized) via `setAccessibilitySortComparator:`.

rcancro: Couldn't you call `setUserDefinedAccessibilitySortComparator(nil)`? Or do you mean there is no good reason to need to switch? Which I may agree with.

jparise: `setUserDefinedAccessibilitySortComparator(nil)` could move the single global pointer back to the default implementation.

The core thing I was trying to express is that we could still have a single global `static` pointer pointing at the active comparator rather than conditionally selecting between two pointers.

rcancro: Hmmm... So you are saying we can make `setUserDefinedAccessibilitySortComparator` do something like: 

```
void setUserDefinedAccessibilitySortComparator(ASSortAccessibilityElementsComparator userDefinedComparator) {
  if (userDefinedComparator) {
    comparator = userDefinedComparator;
  } else {
    comparator = ^NSComparisonResult(NSObject *a, NSObject *b) { ... };
  }
)
```

Then in `SortAccessibilityElements` check in the `dispatch_once` that `comparator` isn't nil (in case no one ever called `setUserDefinedAccessibilitySortComparator`) and set it up if it is (I guess by calling `setUserDefinedAccessibilitySortComparator(nil)` which is pretty gross)? 

On the other hand, if we have 1 pointer and `setUserDefinedAccessibilitySortComparator` is called before the first `SortAccessibilityElements` we could overwrite the user's comparator  in the dispatch_once unless we check for a non-nil `comparator`.

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -226,6 +238,16 @@ static void CollectAccessibilityElements(ASDisplayNode *node, NSMutableArray *el
   }
 
   for (ASDisplayNode *subnode in node.subnodes) {
+    // If a node is hidden or has an alpha of 0.0 we should not include it
+    if (subnode.hidden || subnode.alpha == 0.0) {
+        continue;
+    }
+    // If a subnode is outside of the view's window, exclude it
+    CGRect nodeInWindowCoords = [node convertRect:subnode.frame toNode:nil];
+    if (view.window && !CGRectIntersectsRect(view.window.frame, nodeInWindowCoords)) {
```
rcancro, line 247: Does it make sense to exit early if the view has no window? By requiring a window before checking to see if the view is in its bounds I'm keeping the behavior the same as before, but not sure if it is correct.

Source/Details/_ASDisplayViewAccessiblity.mm
jparise: I'm saying have one `static` pointer named `defaultComparator` (populated by the `dispatch_once` block) and another named `currentComparator`.

```objc
void setAccessibilitySortComparator(ASSortAccessibilityElementsComparator comparator) {
  currentComparator = comparator ?: defaultComparator;
}
```

rcancro: Oh, ok. 
> The core thing I was trying to express is that we could still have a single global static pointer

This was confusing me because it sounded like you had a way to do this with 1 pointer.

jparise: Oh, I think you're trying to avoid having the default comparator visible outside of the function's scope?

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -22,17 +22,18 @@
 
 #pragma mark - UIAccessibilityElement
 
-static ASSortAccessibilityElementsComparator _userDefinedComparator = nil;
+static ASSortAccessibilityElementsComparator currentAccessibilityComparator = nil;
+static ASSortAccessibilityElementsComparator defaultComparator = nil;
```
jparise, line 26: Name these consistently? Either both with, or both without, "Accessibility".

Source/Details/_ASDisplayViewAccessiblity.h
```diff
@@ -14,3 +14,17 @@
...
 // https://developer.apple.com/documentation/objectivec/nsobject/1615147-accessibilityelements
+
+// After recusively collecting all of the accessibility elements of a node, they get sorted. This sort determines
+// the order that a screen reader will traverse the elements. By default, we sort these elements based on their
+// origin: lower y origin comes first, then lower x origin. If 2 nodes have an equal origin, the node with the smaller
+// height is placed before the node with the smaller width. If two nodes have the exact same rect, we throw up our hands
+// and return NSOrderedSame.
+//
+// In general this seems to work fairly well. However, if you want to provide a custom sort you can do so via
+// setUserDefinedComparator(). The two elements you are comparing are NSObjects, which conforms to the informal
```
nguyenhuy, line 25:
```suggestion
// setUserDefinedAccessibilitySortComparator(). The two elements you are comparing are NSObjects, which conforms to the informal
```

Source/Details/_ASDisplayViewAccessiblity.mm
nguyenhuy: I think it makes sense to exit early.

rcancro: Updated! I had to update a few accessibility tests to make sure the node we are testing was in a window.

PR #1820, opened 2020-05-19
Removes unnecessary `-tearDown` code.

dmaclach:
All of the work being done in this `tearDown` is also being done in -[ASTest tearDown].
So we might as well get rid of the redundant code.

### DIFF
--- a/Tests/Common/ASTestCase.mm
+++ b/Tests/Common/ASTestCase.mm
@@ -64,12 +64,12 @@ static __weak ASTestCase *currentTestCase;
 
   for (OCMockObject *mockObject in registeredMockObjects) {
     OCMVerifyAll(mockObject);
-    [mockObject stopMocking];
+   // [mockObject stopMocking];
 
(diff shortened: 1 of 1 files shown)


Tests/Common/ASTestCase.mm
```diff
@@ -64,12 +64,12 @@ - (void)tearDown
 
   for (OCMockObject *mockObject in registeredMockObjects) {
     OCMVerifyAll(mockObject);
-    [mockObject stopMocking];
+   // [mockObject stopMocking];
```
bolsinga, line 67: Hi @dmaclach Where is it being done elsewhere? It's not clear to me from your comment. Thanks (and hello)!

dmaclach: Hey Greg!

My fault.. fat fingered an auto complete. I'm going to drop this PR and replace it with #1822 1822

dmaclach:
Apologies for wasting your time :(

PR #1832, opened 2020-05-22
Use ASReplaceMethodWithBlock instead of swizzling with a library

bolsinga:
- I tried to make these tests work with OCMock, but that only ever saw 1 call.
- All these do is count `-reloadData` calls. I'm not sure why so many other methods were recorded.
- Why we are doing this is the external library targets iOS 5 and is flagged as a warning in builds.


bolsinga:
> Just to confirm, you weren't successful in using OCMock to observer and count the number of expected `-reloadData` calls?

Yes; I tried most of the morning! I stubbed with counts, used the new `times` stuff, and sat in the debugger. I was unable to get this to work. This was the smallest fix in the end to remove the out of date dependency,

PR #1837, opened 2020-05-24
Update Swift collection view interop example

drinkius:
There is an issue with interop in Swift - you can't return nil instead of node or node block. This would help people reading the documentation overcome the issue.

Confusion example:
https://stackoverflow.com/questions/47676341/texture-ascollectiondatasourceinterop-swift/61993513#61993513


docs/_docs/uicollectionviewinterop.md
```diff
@@ -71,8 +71,12 @@ For this example, the data source method `collectionNode:nodeBlockForItemAtIndex
   </pre>
 
   <pre lang="swift" class = "swiftCode hidden">
-  // Click the "Edit on GitHub" button at the bottom of this 
-  // page to contribute the swift code for this section. Thanks!
```
bolsinga, line 75: please do not remove the existing text; please add your example before this block. Thanks.

PR #1858, opened 2020-05-28
[ASDisplayNode] Implement accessibilityViewIsModal

rcancro:
A PR to add support for `accessibilityViewIsModal` in `CollectAccessibilityElements`.

If in a list of subnodes more than 1 subnode has `accessibilityViewIsModal` marked as `YES`, then the node with the highest index in `subnodes` will be the one that is considered modal. This behavior matches UIKit.

If the value of `accessibilityViewIsModal` changes, we need to clear all the cached `accessibilityElements` from that view up. I added this in ASDisplayNode’s `setAccessibilityViewIsModal` method. Note that if we ship `ASExperimentalDoNotCacheAccessibilityElements` then we can remove the invalidation step.

Finally, I changed all the tests to ask the view for accessibilityElements, not the node. This is a better representation of what will really happen when UIKit asks a node’s view for its accessibility elements. It also allowed me to test that clearing the accessibilityElements was working.

### DIFF
--- a/Source/Details/_ASDisplayViewAccessiblity.mm
+++ b/Source/Details/_ASDisplayViewAccessiblity.mm
@@ -254,7 +254,21 @@ static void CollectAccessibilityElements(ASDisplayNode *node, NSMutableArray *el
     return;
   }
   
-  for (ASDisplayNode *subnode in node.subnodes) {
+  // see if one of the subnodes is modal. If it is, then we only need to collect accessibilityElements from that
+  // node. If more than one subnode is modal, UIKit uses the last view in subviews as the modal view (it appears to
+  // be based on the index in the subviews array, not the location on screen). Let's do the same.
+  ASDisplayNode *modalSubnode = nil;
+  for (ASDisplayNode *subnode in node.subnodes.reverseObjectEnumerator) {
+    if (subnode.accessibilityViewIsModal) {
+      modalSubnode = subnode;
+      break;
+    }
+  }
+  
+  // If we have a modal subnode, just use that. Otherwise, use all subnodes
+  NSArray *subnodes = modalSubnode ? @[ modalSubnode ] : node.subnodes;
+  
+  for (ASDisplayNode *subnode in subnodes) {
     // If a node is hidden or has an alpha of 0.0 we should not include it
     if (subnode.hidden || subnode.alpha == 0.0) {
       continue;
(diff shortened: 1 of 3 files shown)


Source/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -1137,6 +1137,18 @@ - (BOOL)_locked_insetsLayoutMarginsFromSafeArea
 
 @implementation ASDisplayNode (UIViewBridgeAccessibility)
 
+// Walks up the view tree to nil out all the cached accsesibilityElements. This is required when changing
+// accessibility properties like accessibilityViewIsModal.
+// NOTE: If we ship ASExperimentalDoNotCacheAccessibilityElements this is not required.
+- (void)invalidateAccessibilityElements
+{
+  BOOL loaded = _loaded(self);
+  if (loaded && ASDisplayNodeThreadIsMain()) {
```
jparise, line 1146: Which is the cheaper / faster check?

- `if (_loaded(self) && ASDisplayNodeThreadIsMain())`
- `if (ASDisplayNodeThreadIsMain() && _loaded(self))`

rcancro: I changed the order because `_loaded` should only be called on the main thread!

Source/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -280,7 +294,7 @@ static void CollectAccessibilityElements(ASDisplayNode *node, NSMutableArray *el
     } else if (subnode.isLayerBacked) {
       // Go down the hierarchy of the layer backed subnode and collect all of the UIAccessibilityElement
       CollectUIAccessibilityElementsForNode(subnode, node, view, elements);
-    } else if (subnode.accessibilityElementCount > 0) {
+    } else if (subnode.accessibilityElements.count > 0) {
```
rcancro, line 297: Does anyone have any thoughts around this change? When I made an example app with `accessibilityViewIsModal` set to `YES` `accessibilityElementCount` was 1 when I turned on voice over. However, in the tests it was always 0. 

I'm not really sure where/how `accessibilityElementCount` is being changed in my test app. It isn't a property on `NSObject`, but a method that should be overwritten (which we aren't doing). This comment also exists in the header of `_ASDisplayViewAccessiblity.h`, so I'm not sure why we are favoring `accessibilityElementCount` over `accessibilityElements`:

```
// WARNING: When dealing with accessibility elements, please use the `accessibilityElements`
// property instead of the older methods e.g. `accessibilityElementCount()`. While the older methods
// should still work as long as accessibility is enabled, this framework provides no guarantees on
// their correctness.
```

vovasty: as far as I understand we should rely on `accessibilityElementCount`, but
according to documentation it returns `0` by default. May be default implementation? 
```
- (NSInteger)accessibilityElementCount {
  return [[self accessibleElements] count];
}
```

rcancro: I just played with this some more, and it doesn't work. I'll focus on a way to make the tests pass and revert this change.

PR #795, opened 2018-02-07
[ASDisplayNode] Ensure accessibilityElementsHidden is respected before returning accessibilityElements to UIKit.

fruitcoder:
Check whether node has `accessibilityElementsHidden` set before returning `accessibilityElements` to UIKit

PR #1859, opened 2020-06-01
[ASDisplayNode] Implement accessibilityElementsHidden

rcancro:
Most of this code comes from an old PR that @fruitcoder put up https://github.com/TextureGroup/Texture/pull/795 2 years ago.

When creating our array of accessibilityElements, we need to respect the value of `accessibilityElementsHidden`. If the value of this property changes, we need to invalidate the cached accessibility elements (unless we are in the experiment that doesn’t cache `accessibilityElements`).

I created a simple test app and made sure this matched UIKit’s implementation. I also added a test case that changes the value of `accessibilityElementsHidden` and makes sure the proper accessibilityElements are returned.


fruitcoder:
@rcancro Wow, that looks great! Thanks for implementing this 👍

Issue #1815, opened 2020-05-08
-[ASNetworkImageNode setURL:resetToDefault:] forget to reset animatedImage


IvanChan:
If you have a `ASNetworkImageNode ` with animatedImage

PR #1861, opened 2020-06-02
-[ASNetworkImageNode setURL:resetToDefault:] forget to reset animatedImage

IvanChan:
If you have a ASNetworkImageNode with animatedImage

Then you call [ImageNode setURL:someLink resetToDefault:true]
Node will still always display Old animatedImage, the new static image will never be display

Related issue https://github.com/TextureGroup/Texture/issues/1815


IvanChan:
@bolsinga unit tests added

bolsinga:
> @bolsinga unit tests added

Thank you!

> I wonder, should also clear animatedImage when a static image is download finish?

I don't know enough about that. However, I do think it would be best in a separate PR!

PR #1872, opened 2020-06-25
Fix all the warnings and re-enable on CI

garrettmoon:


Texture.podspec
```diff
@@ -16,7 +16,7 @@ Pod::Spec.new do |spec|
 
   # Subspecs
   spec.subspec 'Core' do |core|
-    core.compiler_flags = '-fno-exceptions -Wno-implicit-retain-self'
```
garrettmoon, line 19: I'm not sure why this was in here 😬

Issue #1871, opened 2020-06-25
Xcode 12 build error when imported AuthenticationServices


hovox:
When building with Xcode 12 beta1 we are getting this error `Redefinition of 'ASViewController' as different kind of symbol`
Seems cause is this in `ASFoundation.h`:
```
#if __has_include(<UIKit/UIViewController.h>)
typedef UIViewController ASViewController;
#endif
```
I suppose we need to rename `ASViewController` to something else ? E.g. `AS` => `TK`  as a texture kit ?

ay8s:
Wonder if you have any preference on this one @garrettmoon? Wonder if it might be a good thing to slot into 3.0.

hovox:
I have renamed `ASViewController` into `ASDKViewController` and `AS_EXTERN` into `ASDK_EXTERN` and seems that is enough. 


Not related to this issue but related to Xcode 12, there are some new warnings: 

```
'_Atomic' is a C11 extension
```

garrettmoon:
Of course this would happen after we put out an RC.

garrettmoon:
@hovox can you confirm that this is fixed on master? I want to get it into the 3.0 release.

ay8s:
Working for us @garrettmoon 👍

garrettmoon:
Gonna close this out, please lemme know if I should reopen!

mycroftcanner:
https://github.com/TextureGroup/Texture/issues/2013

PR #1876, opened 2020-06-26
Renames AS_EXTERN and ASViewController

garrettmoon:
To ASDK_EXTERN and ASDKViewController.
This is to avoid conflicting with AuthenticationServices in
Xcode 12


garrettmoon:
Resolves https://github.com/TextureGroup/Texture/issues/1871

PR #1878, opened 2020-06-27
Improve ThreeMigrationGuide.md

rogerluan:
## Description

This PR adds `ASImageCacherCompletion` breaking API change to the Texture 3.0 Migration Guide.

This is the only point of contact that I had modify in order to make my project compile. It's not clear whether there're more changes or not, I can't easily tell without looking at the tens of PRs that got merged since 2.8.0 🙈 

Texture 3.0 is now under QA process at Tellus and I'll keep this Migration Guide updated with any other internal details we find.

Thanks once again for pushing this release to the finish line :) it's been a long time! 😄 

Cheers!

### DIFF
--- a/ThreeMigrationGuide.md
+++ b/ThreeMigrationGuide.md
@@ -1,4 +1,4 @@
-## Texture 3.0 Migration Guide ###
+## Texture 3.0 Migration Guide
 
 Got a tip for upgrading? Please open a PR to this document!
 
@@ -21,4 +21,3 @@ ASPINRemoteImageDownloader.shared().cachedImage(with: url, callbackQueue: .main)
     …
 }
 ```
-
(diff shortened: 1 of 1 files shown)


ThreeMigrationGuide.md
```diff
@@ -1,3 +1,24 @@
-### Texture 3.0 Migration Guide ###
+## Texture 3.0 Migration Guide ###
```
jparise, line 1: The two sides are now imbalanced.

rogerluan: Why is there trailing pounds? The trailing is invisible, only the leading is required for the header formatting

rogerluan: I'm more inclined to remove the trailing ones than making it match the leading ones

jparise: I agree. Let's remove the trailing characters.

rogerluan:
```suggestion
## Texture 3.0 Migration Guide
```

rogerluan:
@jparise I don't understand how this PR is failing unit tests 🤔 are the tests broken in master as well? I think I don't have visibility into that

PR #1898, opened 2020-08-20
Propagate leadingScreensForBatching down to the view

shivbaijal:
Collection nodes/table nodes' leadingScreensForBatching property wasn't getting updated once the view was loaded. This fixes that.


shivbaijal:
will update shortly

PR #1903, opened 2020-09-01
Fix hit point test for inverted ASCollectionNode

drinkius:
There are many chat applications that rely on inverted CollectionView. Cell taps were broken in the 3.0.0 release, this change fixes taps for such cases. It contains the same code as #1781 but that branch is stale and not fixes the issues. I intend on fixing those to get the changes merged


bolsinga:
Thanks for the diff! If you would also work on tests so this does not regress, as also asked in #1781, it would be terrific. Thanks again.

drinkius:
Sure, will deliver the fixes shortly

PR #1908, opened 2020-09-17
Update cartfile to use PINRemoteImage master branch to support Xcode 12

udayasri:


Cartfile
```diff
@@ -1,2 +1,2 @@
-github "pinterest/PINRemoteImage" "3.0.0-beta.14"
-github "pinterest/PINCache" "3.0.1-beta.7"
+github "pinterest/PINRemoteImage" "master"
+github "pinterest/PINCache" "master"
```
bolsinga, line 2: My gut feeling tells me that these should not be pinned to master. The Podspec has 3.00 for PINRemoteImage, and 3.0.1-beta.8 for PINCache.

MeGaPk:
set "master" like a version, it's very bad idea....

PR #1911, opened 2020-09-17
Add an experiment that makes ASDataController to do everything on main thread

nguyenhuy:
- Under this experiment, ASDataController will allocate and layout all nodes on the main thread. This helps to avoid deadlocks that would otherwise occur if some of the node allocations or layouts caused ASDataController's background queue to block on main thread. As a bonus, this experiment also helps to measure how much performance wins we get from doing the work off main.

### DIFF
--- a/Source/Details/ASDataController.mm
+++ b/Source/Details/ASDataController.mm
@@ -33,8 +33,6 @@
 //#define LOG(...) NSLog(__VA_ARGS__)
 #define LOG(...)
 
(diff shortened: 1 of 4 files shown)


rcancro (approved):

garrettmoon (approved):
Looks good to me. If this works we may want to rethink how we solve this problem longer term.

woshimaliang (approved):
thanks @nguyenhuy !

Source/Details/ASDataController.mm
```diff
@@ -133,8 +131,6 @@ - (void)setLayoutDelegate:(id<ASDataControllerLayoutDelegate>)layoutDelegate
 
 - (void)_allocateNodesFromElements:(NSArray<ASCollectionElement *> *)elements
 {
-  ASSERT_ON_EDITING_QUEUE;
```
nguyenhuy, line 136: Removing this assertion since it's no longer true under the new treatment group.

PR #1920, opened 2020-09-24
Update podspec to latest stable PINRemoteImage and PINCache dependencies

DShamaev:
To address issues with using Texture pod in Xcode 12

Important note: Tests have been passed without issues using Xcode 11 locally for those changes, but fails in Xcode 12 (c++ language related issue in mock logic)

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,6 +1,6 @@
 PODS:
   - iOSSnapshotTestCase/Core (6.2.0)
-  - OCMock (3.6)
+  - OCMock (3.7.1)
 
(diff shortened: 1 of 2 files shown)


garrettmoon:
I'm not sure these changes are necessary. [The spec already allows for this update by using `~`](https://stackoverflow.com/questions/20213751/what-is-the-usage-of-in-cocoapods) Would running `pod update` not suffice?

DShamaev:
Oh, i forgot about that, yes, you are right, so i will close this PR

PR #1941, opened 2020-12-16
All ASCellNode nodes to be non accessible if needed

decim92:
Currently there's no way to make a ASCellNode node non accessible, this change solves the problem for cases when you only want the cell node to be a decorative view like a custom separator between cells

### DIFF
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -116,7 +116,8 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
     self.selectionStyle = node.selectionStyle; 
     self.focusStyle = node.focusStyle;
     self.accessoryType = node.accessoryType;
-    
+    self.isAccessibilityElement = node.isAccessibilityElement;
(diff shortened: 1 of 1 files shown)


bolsinga (approved):

bdolman:
@decim92 This change has caused some issues for us and others (see https://github.com/TextureGroup/Texture/issues/1997). If I set my `myNode.isAccessibilityElement = true`, then what ends up happening is that the table cell wrapper view also has `isAccessibilityElement = true`. But then when iOS goes through the hierarchy it identifies the table cell _wrapper_ view as the target and attempts to use that `accessibilityLabel`, which doesn't exist (because it's actually on the nested node view). So our node's label is never used.

What I'm wondering is more about why this change was necessary for you, since `isAccessibilityElement`'s default value is already false, so there doesn't seem to be a need to forcefully set it to false. Would love to hear more about your use case so that we could potentially find a way to handle your use case but restore accessibility functionality for everyone else. Thanks!

paulgee31 (commented):
I can confirm this fix restores the correct behaviour

PR #1948, opened 2021-01-22
Exposes a new option in ASImageDownloaderProtocol to retry image downloads

chggr:
At the moment the ASBasicImageDownloader does not automatically retry image downloads if the remote
host is unreachable. On the contrary the ASPINRemoteImageDownloader automatically retries. Retrying is
something that ultimately clients need to be able to control, for example to fail fast to an alternative image
rather than keep retrying for more than one minute while not displaying any image. This change exposes
a new option in the ASImageDownloaderProtocol to retry image downloads. It also uses this new option
in both ASNetworkImageNode and also ASMultiplexImageNode, setting it to YES to preserve the current
behaviour.

### DIFF
--- a/Source/ASMultiplexImageNode.h
+++ b/Source/ASMultiplexImageNode.h
@@ -131,6 +131,12 @@ typedef NS_ENUM(NSUInteger, ASMultiplexImageNodeErrorCode) {
 @property (nonatomic) BOOL shouldRenderProgressImages;
 
 /**
+ * Specifies whether the underlying image downloader should attempt to retry downloading the image if the remote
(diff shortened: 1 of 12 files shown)


chggr:
Hi team, just wanted to check whether you had any feedback on this PR. Thanks!

chggr:
Hi team, just a friendly ping, who would be the best person to review and provide feedback / merge this PR?

bolsinga (approved):
The only nit I have is to back fill tests that the `PINRemoteImageManagerDownloadsOptions` are populated correctly. Thanks!

PR #1949, opened 2021-01-24
fix format specifier

joprice:
This applies the suggested fix that shows up when `ASEnableVerboseLogging` is enabled:

```
external/Texture/Source/ASNetworkImageNode.mm:891:144: error: enum values with underlying type 'NSInteger' should not be used as format arguments; add an explicit cast to 'long' instead [-Werror,-Wformat]
            as_log_verbose(ASImageLoadingLog(), "Decached image for %@ img: %@ url: %@ cacheType: %@", self, [imageContainer asdk_image], URL, cacheType);```

### DIFF
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -888,7 +888,7 @@ static std::atomic_bool _useMainThreadDelegateCallbacks(true);
             if (delegateDidLoadImageFromCache) {
               [delegate imageNodeDidLoadImageFromCache:self];
             }
-            as_log_verbose(ASImageLoadingLog(), "Decached image for %@ img: %@ url: %@ cacheType: %@", self, [imageContainer asdk_image], URL, cacheType);
+            as_log_verbose(ASImageLoadingLog(), "Decached image for %@ img: %@ url: %@ cacheType: %ld", self, [imageContainer asdk_image], URL, (long) cacheType);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Source/ASNetworkImageNode.mm [hunk 1]
```diff
@@ -888,7 +888,7 @@ - (void)_lazilyLoadImageIfNecessary
             if (delegateDidLoadImageFromCache) {
               [delegate imageNodeDidLoadImageFromCache:self];
             }
-            as_log_verbose(ASImageLoadingLog(), "Decached image for %@ img: %@ url: %@ cacheType: %@", self, [imageContainer asdk_image], URL, cacheType);
+            as_log_verbose(ASImageLoadingLog(), "Decached image for %@ img: %@ url: %@ cacheType: %ld", self, [imageContainer asdk_image], URL, (long) cacheType);
```
bolsinga, line 891: `sizeof(NSInteger) > sizeof(long)`

Another way to do this is to keep the formatter the same but box the scalar ie `@(cacheType)`.

joprice:
I'm trying to verify this change by using `build.sh`, but I have a newer sdk. When I provide TEXTURE_BUILD_PLATFORM and TEXTURE_BUILD_SDK with my current versions, I get errors related to missing extern modifiers, etc. It seems that `-Werror`, and I'm not sure if that's intentional and the sdk version caused this, or if these warnings are expected and I'm supposed to ignore them somehow. 

Up until now, I've been testing this by building it via bazel using PodToBuild, but I'd like to be able to test it directly and run the tests (or perhaps bazel support could be added directly?).

PR #1952, opened 2021-01-24
fix missing hidden class

joprice:
This swift code example is missing the "hidden" class, which all others have, which causes the objc and swift code to show up simultaneously on initial page load.


garrettmoon (approved):
Indeed! Good fix, thanks!

bolsinga:
@joprice please sign the license

joprice:
Signed

PR #1955, opened 2021-01-25
Fix WKWebView Accessibility

### LINKED ISSUE #1954: WKWebView in ASDisplayNode doesn't work with VoiceOver
If a WKWebView is loaded into an ASDisplayNode, the web view's HTML contents are not accessible to VoiceOver. I will upload a sample project later, or if I can get the test target building on Xcode 12, I'll write a unit test that demonstrates the problem.

A potential fix that seems to work: modify _ASDisplayViewAccessibility.mm's implementation of -accessibilityElements to return `nil` instead of an empty array if it does not find any elements. I believe this is because WKWebView's accessibility elements are handled out-of-process. By returning an empty array, I think we were telling the system that there were no accessibility elements.

Reproduced on iOS 14.3.

ZevEisenberg:
Fixes #1954.

TODO:
- [x] ~do we need to worry about whether `CollectAccessibilityElements` encountered empty arrays vs nils?~ no, because either way we're returning nil on empty
- [x] write unit or UI test


bolsinga:
@ZevEisenberg Please sign the license

ZevEisenberg:
We can certainly test using an array, which would be a unit test. What I've added is more of an integration test, because it tests what the actual bug was: that WKWebView was not accessible. I've added a similar test to our own code base, so if you think it's not appropriate here, or not worth adding the UI test target, I'm happy to replace it with a unit test.

Also, it looks like the license bot is broken, per a message on another PR.

ZevEisenberg:
Update: license should be all good to go. @bolsinga or @garrettmoon what do you think about unit vs UI test? I'll do whichever you prefer.

garrettmoon:
I would also prefer a unit test if it will catch errors in the future!

ZevEisenberg:
It will catch errors with returning nil vs `[]`, but it won't catch errors if we or Apple somehow break nested WKWebView accessibility in the future. That seems less likely, just making sure the tradeoffs are understood. I'll work on a unit test tomorrow.

garrettmoon:
Thanks for being clear! I think that tradeoff is worth it in this case. Thanks again!

ZevEisenberg:
Removed the UI test, added a unit test.

garrettmoon (approved):
LGTM, thank you for the fix!

bolsinga (approved):
thanks for the update!

ZevEisenberg:
The build failed for seemingly flaky reasons. I don't have permission to re-run builds, so I pushed and then deleted a commit to kick it.

PR #1958, opened 2021-01-27
Fix some static analysis warnings

ZevEisenberg:
Didn't fix all of them, just a couple that were easy.

### DIFF
--- a/Source/ASDisplayNodeExtras.mm
+++ b/Source/ASDisplayNodeExtras.mm
@@ -65,7 +65,7 @@ ASInterfaceState ASInterfaceStateForDisplayNode(ASDisplayNode *displayNode, UIWi
         // if not already, about to be set to invisible as it is not possible for an element to be visible
         // while outside of a window.
         ASInterfaceState interfaceState = displayNode.pendingInterfaceState;
-        return (window == nil ? (interfaceState &= (~ASInterfaceStateVisible)) : interfaceState);
+        return (window == nil ? (interfaceState & (~ASInterfaceStateVisible)) : interfaceState);
     } else {
         // For not range managed nodes we might be on our own to try to guess if we're visible.
         return (window == nil ? ASInterfaceStateNone : (ASInterfaceStateVisible | ASInterfaceStateDisplay));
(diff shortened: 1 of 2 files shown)


Source/ASDisplayNodeExtras.mm
```diff
@@ -65,7 +65,7 @@ ASInterfaceState ASInterfaceStateForDisplayNode(ASDisplayNode *displayNode, UIWi
         // if not already, about to be set to invisible as it is not possible for an element to be visible
         // while outside of a window.
         ASInterfaceState interfaceState = displayNode.pendingInterfaceState;
-        return (window == nil ? (interfaceState &= (~ASInterfaceStateVisible)) : interfaceState);
+        return (window == nil ? (interfaceState & (~ASInterfaceStateVisible)) : interfaceState);
```
bolsinga, line 68: while Xcode may flag this, I think the comment above indicates its intentional.

Source/TextExperiment/Component/ASTextLayout.mm
```diff
@@ -2385,7 +2385,7 @@ static void ASTextDrawRun(ASTextLine *line, CTRunRef run, CGContextRef context,
               if (mode) { // CJK glyph, need rotated
                 CGFloat ofs = (ascent - descent) * 0.5;
                 CGFloat w = glyphAdvances[g].width * 0.5;
-                CGFloat x = x = line.position.x + verticalOffset + glyphPositions[g].y + (ofs - w);
```
bolsinga, line 2388: This seems good. small diffs are terrific. Would you mind splitting this up? Thanks.

bolsinga (changes requested):

Source/ASDisplayNodeExtras.mm
bolsinga: regardless, since I have no idea, but that seems more in-depth thinking is required than the other one. I think splitting this PR up is best. 

For this change, I think I'd want unit tests that ensure the existing behavior persists. I've no idea how much work that would be w/o an expert weighing in on this one.

ZevEisenberg: Opened a [new PR](https://github.com/TextureGroup/Texture/pull/1961) for this. If you still want a unit test after reading my description, I'm happy to add one :)

ZevEisenberg:
Closing in favor of #1960 and #1961

PR #1962, opened 2021-02-08
Update asdkGram swift sample to swift version 5.3

MussaCharles:
This fixes: Fix: `SWIFT_VERSION '3.0' is unsupported, supported versions are: 4.0, 4.2, 5.0`. when running on Xcode 12+. 

Minor changes have been added as follows: - 
1. Changed the version number on ASDKGram sample Build Settings from Swift 3 to Swift 5.3.
2. Renamed a couple of method calls on ASDKGram sample to satisfy new Swift 5.3 API.


garrettmoon (approved):
Thank you for the improvements!

PR #1963, opened 2021-02-09
Fix order-dependent ASTextNodeTests

tjaneczko:
The ASTextNodeTests currently depend on executing in a specific order as some tests enable the `ASExperimentalTextNode` experimental feature. This is set in a global config, so running any tests which require this experimental feature to not be enabled after a test which enables it will cause that test to fail. This PR resets the experimental features to `kNilOptions` in the per-test setup, removing this ordering dependency.

### DIFF
--- a/Tests/ASTextNodeTests.mm
+++ b/Tests/ASTextNodeTests.mm
@@ -61,6 +61,12 @@
 - (void)setUp
 {
   [super setUp];
+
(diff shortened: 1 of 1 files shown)


appleguy (approved):
Looks good, makes sense that we'd have to ensure ASTextNode is enabled for all builders of Texture running these tests.

PR #1965, opened 2021-02-19
[ASTextNode2] If a textnode's bounds match either the text container's size OR the layout's size, consider it a cache hit

rcancro:
While working in `ASTextNode2` I noticed we don’t seem to use the cached `ASTextLayout` when drawing the text. Here is what I’m seeing:

When `ASTextNode2` calculates its size it calls `layoutWithContainer:text:range:`. The container passed into `layoutWithContainer:text:range:` holds the input information for attributes like size, insets, exclusion paths, etc that will help create the layout. When the layout is computed it is passed back to `ASTextNode2` via an `ASTextLayout` object. This layout is placed into a cache using the container's size for the key.

When the text is actually drawn via `drawRect`, `drawParametersForAsyncLayer` provides us with a container that we use to look for the cached layout. However, after copying `ASTextNode2's`  `_textContainer`, it changes its size to `self.bounds.size` :

```
    copiedContainer = [_textContainer copy];
    copiedContainer.size = self.bounds.size;
    [copiedContainer makeImmutable];
```

We then take this container and look in the cache for a related `ASTextLayout`. Of course, we don’t find the layout we just placed into the cache because we are using a different `size` as a key into the cache. It seems like this behavior guarantees that we will always have a cache miss unless our original constrained size and the final size of the layout happen to be the same.

Am I missing something? Do we really want to change the container’s bounds here?


wiseoldduck:
I think the original layout pass (`calculateSizeThatFits:`) is usually unconstrained, at least when working with Yoga, and when we draw we know our exact size which might be quite different depending on what happened during layout. And we can't reuse the layout if the dimensions (in particular the width) are substantially different.

rcancro:
Even if we lay out originally with an unconstrained size, the `ASTextLayout` that we compute will have the proper text bounds.   That layout will go into the cache with the key of size (inf, inf) for our given text string.

If we then change the container's size before drawing, we'll never be able to use this already computed layout. `ASTextNode2's` `calulateSizeThatFits:` returns the computed `ASTextLayout's` `textBoundingSize` as the calculated size. So when we set `copiedContainer.size = self.bounds.size;` it seems like all we are doing is ensuring a cache miss. If we left the size alone, we'd pull out the `ASTextLayout` from the cache with the proper `textBoundingSize`. Instead, we are running the layout again with our container's size set to what we already determined to be the text's size.

Would this be different somehow with Yoga?

wiseoldduck:
I guess it is -- I'm actually not familiar with what Texture layout does _without_ yoga, but with it, the value we return from `calculateSizeThatFits:` is a preferred size, and not guaranteed to be the final size at all.

wiseoldduck:
It does seem like you're onto something though. It would be great to avoid making another `ASTextLayout` in the large number of cases where the size we return is the size we wind up. Maybe we could insert it into the cache both under the unconstrained size and the size we are returning?

Adlai-Holler:
So, in the past the cache key checking included some "intelligence" in the ASTextContainer size, to handle this. Like, if the container in the cache-key was (100, inf) and we it had a text layout of size (100, 500), then the cache-key-equal method would say "oh! This works! The size is in the range requested." But during a time when I was out of the loop, the size-checking-fuzziness caused false positives and was removed and it had to match exactly. I don't remember what the explanation for it was. Obviously there was some condition under which it was _too lenient_. But yes!! It may be worth launching an experiment to reintroduce the aforementioned "intelligence" into the cache checking because laying out text twice every time is awful.

It might be good for a small population, just to see what performance is left on the table. Maybe they'd see bad text 0.1% of the time – not sure.

I think the bounds-set is necessary because we do have to honor the fact that, whatever our bounds are when it comes time to draw, those are our bounds so we need a text layout that fits them. In terms of our layout engine state we may have a calculatedSize or a constraint or whatever of whatever, but our bounds are our bounds so we should search the cache for that, hence the copy.

rcancro:
Thank you both for taking a look and thinking about this! And @Adlai-Holler, nice to hear from you!!!!

@wiseoldduck I was also thinking that maybe we could use both the container size and the resulting layout size as keys into the cache. Let me give that a shot on Monday and see what it looks like. I'll update the diff once I'm happy with it!

rcancro:
I took a second pass at this. I opted for not putting the layout in the cache twice, as it appears we limit the number of layouts to 3: 
```
    cacheValue->_layouts.push_front(std::make_tuple(container.size, layout));
    if (cacheValue->_layouts.size() > 3) {
      cacheValue->_layouts.pop_back();
    }
```

Instead, in `ASTextNodeCompatibleLayoutWithContainerAndText` I consider a cached layout useable if the container's size is equal to the key in the cache OR if the container's size is equal to the layout's `textBoundingSize`*

 `*` Actually the textBoundingSize is not always equal to the text node's size when there is truncation involved. So I moved some logic in `ASTextNode2's` `calculateThatFits` into a method to return the `textBoundingSize` taking the truncated line into consideration. If this size is equal to the container's size, then I determine we can use that layout.

rcancro:
I looked into the snapshot test (`testTextContainerInsetIsIncludedWithSmallerConstrainedSize_ASTextNode2`) that was failing and I think it is actually something that this change fixes. 

In the test, we render `judar judar judar judar judar judar` in a textnode with a constrained size of `(100, 80)`. The resulting layout has a size of `(92, 57)` and renders as `juda…` . Before drawing the text on screen, we set the container's size to `(92, 57)` per Adlai's comments above. This causes a cache miss, so we calculate layout again with our new constrained size of `(92, 57)`.

In the layout, we use the width of our size (minus any insets) as the `truncatedWidth`. This width is used to create `truncatedLine` by calling `CTLineCreateTruncatedLine`. The first time through our `truncatedWidth` is `80` (`100` minus the left and right insets of `10`). The second time it is `72`(`92` minus the insets). In both cases we end up with a `textBoundingSize` of `(92, 57)`, but in the second pass we truncate an extra letter since we have a smaller width:

Screen Shot 2021-02-23 at 1 58 11 PM
First pass with a container size of `(100, 80)`

Screen Shot 2021-02-23 at 1 57 21 PM
Second pass with a container size of `(92, 57)`

Note the text nodes are the same size, but the second one clips more text. 

The new code in this PR allows us to check if our textNode's bounds are the same size as the layout's `textBoundingSize`. If they are, we use the cached layout. The cached layout had the `truncatedWidth` of 80 and fit the extra letter. 

As a side note, if we try to render the same thing in `ASTextNode` (v1) it renders `juda...`.

wiseoldduck:
This all makes more sense than throwing out the size change did, however my concern now is what Scott said in the Slack channel, just from a practical perspective as this becomes more and more not a one-liner it is going to complicate our efforts to merge everything and push up the rest of our changes.

I don't think much has changed in this area, but have you compared with the version of [ASTextNode2](https://github.com/wiseoldduck/Texture/tree/astextnode-2) we pushed up? There is also an entirely new integration with Yoga (creatively dubbed Yoga2) that has not been shared yet (or if it has it is a very early version).

Anyway, if it's practical, it would make our lives easier to wait on merging this for a bit...

I suspect we would also want to gate the snapshot-changing part behind an experiment, since it is absurdly expensive to push out a text rendering change like that for us even if it is technically correct :-/ (which trust me I hate)

wiseoldduck:
To be sure, this is a very exciting change! If we can throw out even a small percentage of text layouts (and this promises to be more than a small percentage) it will be a substantial performance gain for us as we render lots and lots and lots of text. I'm eager to get caught up in particular so we can potentially take advantage of this.

rcancro:
I'm happy to put this change behind an experiment -- it definitely seems like a safer choice! Plus it'll make it easier to see if this change actually moves perf metrics!

rcancro:
@wiseoldduck Regarding waiting to merge this, do you have an ETA for your `ASTextNode2` changes? I've got some more work planned for the class that I need to do so we can move forward with RTL. If the PR isn't imminent we may have to get creative with ideas around how to unblock each other.

wiseoldduck:
Let's move that conversation to the teams thread, since it is more about the logistics of our teams' merging and scheduling than specific to the content of this PR.

(You in fact already did, this comment is really for posterity) 😃

PR #1973, opened 2021-03-11
Remove trailing semicolons between method parameters and body

sdefresne:
Having a semi-colon between a method parameters list and a method
body is not not correct and is usually caused by a copy and paste
error while creating the method definition from its declaration.

Fixes the following compilation warnings when building with
-Wsemicolon-before-method-body (which is part of -Wextra):

  ASPINRemoteImageDownloader.mm:230:85: error: semicolon before method body is ignored [-Werror,-Wsemicolon-before-method-body]
  - (id <ASImageContainerProtocol>)synchronouslyFetchedCachedImageWithURL:(NSURL *)URL;
                                                                                      ^
  ASPINRemoteImageDownloader.mm:275:76: error: semicolon before method body is ignored [-Werror,-Wsemicolon-before-method-body]
                           completion:(ASImageDownloaderCompletion)completion;
                                                                             ^
  2 errors generated.

Fixes applied to both code, examples and samples in documentation.

### DIFF
--- a/Source/ASCellNode.mm
+++ b/Source/ASCellNode.mm
@@ -159,7 +159,7 @@
   }
 }
 
-- (void)__setSelectedFromUIKit:(BOOL)selected;
+- (void)__setSelectedFromUIKit:(BOOL)selected
 {
   // Note: Race condition could mean redundant sets. Risk is low.
   if (ASLockedSelf(_selected != selected)) {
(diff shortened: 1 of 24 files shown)


garrettmoon (approved):
LGTM! Thanks for the fixes!

PR #1974, opened 2021-03-12
[ASTextNode2] Scale the truncation token via pointScaleFactors if needed

rcancro:
If we need to scale text in order to fit in our size, don’t forget to also scale the truncation token!

(disabled carthage so the builds will pass)


bolsinga (approved):
thanks for adding the test

PR #1976, opened 2021-03-24
Add experiment to ensure ASCollectionView's range controller updates …

rqueue:
…on changeset updates

This experiment makes sure a ASCollectionView's `rangeController` updates when
a changeset WITH updates is applied. Currently it is possible for nodes
inserted into the preload range to not get preloaded when performing a batch
update.

  For example, suppose a collection node has:
  - Tuning parameters with a preload range of 1 screenful for the given range
    mode.
  - Nodes A and B where A is visible and B is off screen.
  
  Currently if node B is deleted and a new node C is inserted in its place, node
  C will not get preloaded until the collection node is scrolled. This is because
  the preloading mechanism relies on a `setNeedsUpdate` call on the range
  controller as part of the `-collectionView:willDisplayCell:forItemAtIndexPath:`
  delegate method when the batch update is submitted. However, in the example
  outlined above, this sometimes doesn't happen automtically, causing the range
  update to be delayed until the next the view scrolls.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -2311,6 +2311,9 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
       
       // Flush any range changes that happened as part of submitting the update.
       as_activity_scope(changeSet.rootActivity);
+      if (numberOfUpdates > 0 && ASActivateExperimentalFeature(ASExperimentalRangeUpdateOnChangesetUpdate)) {
+        [self->_rangeController setNeedsUpdate];
+      }
       [self->_rangeController updateIfNeeded];
     }
   });
--- a/Source/ASExperimentalFeatures.h
+++ b/Source/ASExperimentalFeatures.h
@@ -30,6 +30,7 @@ typedef NS_OPTIONS(NSUInteger, ASExperimentalFeatures) {
   ASExperimentalOptimizeDataControllerPipeline = 1 << 9,                    // exp_optimize_data_controller_pipeline
   ASExperimentalDisableGlobalTextkitLock = 1 << 10,                         // exp_disable_global_textkit_lock
   ASExperimentalMainThreadOnlyDataController = 1 << 11,                     // exp_main_thread_only_data_controller
+  ASExperimentalRangeUpdateOnChangesetUpdate = 1 << 12,                     // exp_range_update_on_changeset_update
   ASExperimentalFeatureAll = 0xFFFFFFFF
 };
 
--- a/Source/ASExperimentalFeatures.mm
+++ b/Source/ASExperimentalFeatures.mm
@@ -23,7 +23,8 @@ NSArray<NSString *> *ASExperimentalFeaturesGetNames(ASExperimentalFeatures flags
                                       @"exp_drawing_global",
                                       @"exp_optimize_data_controller_pipeline",
                                       @"exp_disable_global_textkit_lock",
-                                      @"exp_main_thread_only_data_controller"]));
+                                      @"exp_main_thread_only_data_controller",
+                                      @"exp_range_update_on_changeset_update"]));
   if (flags == ASExperimentalFeatureAll) {
     return allNames;
   }
--- a/Tests/ASCollectionViewTests.mm
+++ b/Tests/ASCollectionViewTests.mm
@@ -24,6 +24,7 @@
 
 @property (nonatomic) NSUInteger setSelectedCounter;
 @property (nonatomic) NSUInteger applyLayoutAttributesCount;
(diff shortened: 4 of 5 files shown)


Source/ASExperimentalFeatures.h
```diff
@@ -30,6 +30,7 @@ typedef NS_OPTIONS(NSUInteger, ASExperimentalFeatures) {
   ASExperimentalOptimizeDataControllerPipeline = 1 << 9,                    // exp_optimize_data_controller_pipeline
   ASExperimentalDisableGlobalTextkitLock = 1 << 10,                         // exp_disable_global_textkit_lock
   ASExperimentalMainThreadOnlyDataController = 1 << 11,                     // exp_main_thread_only_data_controller
+  ASExperimentalRangeUpdateOnChangesetUpdate = 1 << 12,                     // exp_main_thread_only_data_controller
```
jparise, line 33:
```suggestion
  ASExperimentalRangeUpdateOnChangesetUpdate = 1 << 12,                     // exp_range_update_on_changeset_update
```

rqueue: Thanks, updated

Source/ASCollectionView.mm
```diff
@@ -2311,7 +2311,11 @@ - (void)rangeController:(ASRangeController *)rangeController updateWithChangeSet
       
       // Flush any range changes that happened as part of submitting the update.
       as_activity_scope(changeSet.rootActivity);
-      [self->_rangeController updateIfNeeded];
+      if (numberOfUpdates > 0 && ASActivateExperimentalFeature(ASExperimentalRangeUpdateOnChangesetUpdate)) {
+        [self->_rangeController setNeedsUpdate];
+      } else {
+        [self->_rangeController updateIfNeeded];
```
nguyenhuy, line 2317: Do we also want to run `updateIfNeeded` right after `setNeedsUpdate` in the experiment? I don't think it's a big deal either ways because `updateIfNeeded` will be called shortly by other call sites (like the one in the `completion` block [here](https://github.com/TextureGroup/Texture/pull/1976/files#diff-a763091124ee49f35c39703ce93f1d38fb204a058ff6483896416ffa678aa155R2258)), but wrapping it in an `else` statement seems like an unnecessary change for this experiment?

```
      if (numberOfUpdates > 0 && ASActivateExperimentalFeature(ASExperimentalRangeUpdateOnChangesetUpdate)) {
        [self->_rangeController setNeedsUpdate];
      }
      
      [self->_rangeController updateIfNeeded];
```

Source/ASCollectionView.mm
```diff
@@ -2311,7 +2311,11 @@ - (void)rangeController:(ASRangeController *)rangeController updateWithChangeSet
       
       // Flush any range changes that happened as part of submitting the update.
       as_activity_scope(changeSet.rootActivity);
-      [self->_rangeController updateIfNeeded];
+      if (numberOfUpdates > 0 && ASActivateExperimentalFeature(ASExperimentalRangeUpdateOnChangesetUpdate)) {
+        [self->_rangeController setNeedsUpdate];
```
nguyenhuy, line 2315: When we first talked about this diff, I was thinking about calling `setNeedsUpdate` in the `completion` block ([here](https://github.com/TextureGroup/Texture/pull/1976/files#diff-a763091124ee49f35c39703ce93f1d38fb204a058ff6483896416ffa678aa155R2258)). But thinking about it more, I think you're right that it's better do so here because the any range changes is the result of submitting the update (instead of the following update animations).

Source/ASCollectionView.mm
rqueue: Originally I put it in the `else` because it would get dispatched in a call within `setNeedsUpdate` [here](https://github.com/TextureGroup/Texture/blob/5a205d84c842c8709819898bd6ee4a74c1f91a3d/Source/Details/ASRangeController.mm#L139), but now I actually think it's better to leave the `updateIfNeeded` call just in case the range controller's `_needsRangeUpdate` is already set to `YES` and the call to `updateIfNeeded` is not dispatched. I'll update this

Source/ASCollectionView.mm
rqueue: Ah I was actually unaware of this subtlety. This makes sense though

nguyenhuy (approved):
LGTM!

Not a blocking issue, but would it be possible to verify if the issue also occurs on ASTableView (perhaps by porting your new unit test), and if yes, address it in a follow up PR?

rqueue:
> LGTM!
> 
> Not a blocking issue, but would it be possible to verify if the issue also occurs on ASTableView (perhaps by porting your new unit test), and if yes, address it in a follow up PR?

Good call. I'll test that out and create a follow up PR if it's an issue

PR #1977, opened 2021-03-25
[ASTextNode2] Fix centered text shorter than container size

rcancro:
ASTextNode2 renders centered text a little strangely. If we have a small bit of text in a container size larger than the text, ASTN2 will use the entire container width to center the text but the text node's frame itself will not be the entire container size. For example, here is what ASTN2 looks like when centering text across the width of the phone:

Screen Shot 2021-03-23 at 12 37 46 PM

While the text is centered across the phone's width, the text is not centered in the text node itself. 

This fix will align the text node's frame with the container size that computes the offset of the centered text, like this:
Screen Shot 2021-03-25 at 10 47 59 AM


nguyenhuy (approved):
I'm not familiar with this part of the code base but this seems like a reasonable short-term fix, considering it's for `master_p` branch.

PR #1927, opened 2020-10-06
Remove Facebook from "Showcase"

adamjernst:
The Facebook app no longer uses Texture.

Promote NYT to the first row and choose a random app from the last row to promote to the second row, so I don't have to rejigger the entire table.

PR #1978, opened 2021-03-25
Docs: Remove Facebook and shift everything around, add Remix by Buffer

ay8s:
Adds Remix by Buffer and also removes Facebook. Should resolve #1927. 👍


garrettmoon (approved):
Thank you @ay8s !

PR #1979, opened 2021-03-25
Expand ASExperimentalRangeUpdateOnChangesetUpdate to ASTableView

rqueue:
A previous commit
(https://github.com/TextureGroup/Texture/commit/8f7444e0ece61d6ab12ecceb590be26d8d7cc99d)
aimed to fix a preloading bug for ASCollectionView. This commit expands this
fix to ASTableView as the bug occurs there too.

Previous commit message for context:

This experiment makes sure a ASCollectionView's `rangeController` updates when
a changeset WITH updates is applied. Currently it is possible for nodes
inserted into the preload range to not get preloaded when performing a batch
update.

For example, suppose a collection node has:
- Tuning parameters with a preload range of 1 screenful for the given range
  mode.
- Nodes A and B where A is visible and B is off screen.
Currently if node B is deleted and a new node C is inserted in its place, node
C will not get preloaded until the collection node is scrolled. This is because
the preloading mechanism relies on a `setNeedsUpdate` call on the range
controller as part of the `-collectionView:willDisplayCell:forItemAtIndexPath:`
delegate method when the batch update is submitted. However, in the example
outlined above, this sometimes doesn't happen automtically, causing the range
update to be delayed until the next time the view scrolls.

### DIFF
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -1679,6 +1679,9 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   LOG(@"--- UITableView endUpdates");
   ASPerformBlockWithoutAnimation(!changeSet.animated, ^{
     [super endUpdates];
+    if (numberOfUpdates > 0 && ASActivateExperimentalFeature(ASExperimentalRangeUpdateOnChangesetUpdate)) {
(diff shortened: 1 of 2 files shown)


nguyenhuy (approved):
LGTM. Thanks for verifying the issue on ASTableView with a test and fixing it!

PR #1985, opened 2021-05-03
[RTL/Batching] Make ASDisplayShouldFetchBatchForScrollView aware of flipped CV layouts

rcancro:
`UICollectionViewLayout` has a property called `flipsHorizontallyInOppositeLayoutDirection`. If this is set to `YES` then a RTL collectionView’s contentOffset behaves like it does in LTR. In other words, the first item is at contentOffset 0. In this case, the existing logic for `ASDisplayShouldFetchBatchForScrollView` works in RTL.

If you don’t override `flipsHorizontallyInOppositeLayoutDirection` to be `YES`, then it means that in RTL languages the first item in your collectionView will actually be at x offset `collectionView.contentSize.width - collectionView.frame.size.width`. As you scroll to the right, the content offset will decrease until you reach the end of the data at a content offset of 0,0. In this case, `ASDisplayShouldFetchBatchForScrollView` needs to know that you are in RTL and the layout is not flipped. It can then use the contentOffset as the `remainingDistance` to determine when to fetch.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -1857,7 +1857,7 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 
 - (void)_beginBatchFetchingIfNeededWithContentOffset:(CGPoint)contentOffset velocity:(CGPoint)velocity
 {
-  if (ASDisplayShouldFetchBatchForScrollView(self, self.scrollDirection, self.scrollableDirections, contentOffset, velocity)) {
+  if (ASDisplayShouldFetchBatchForScrollView(self, self.scrollDirection, self.scrollableDirections, contentOffset, velocity, self.collectionViewLayout.flipsHorizontallyInOppositeLayoutDirection)) {
     [self _beginBatchFetching];
   }
 }
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -1857,6 +1857,9 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 
 - (void)_beginBatchFetchingIfNeededWithContentOffset:(CGPoint)contentOffset velocity:(CGPoint)velocity
 {
+  // Since we are accessing self.collectionViewLayout, we should make sure we are on main
+  ASDisplayNodeAssertMainThread();
+  
   if (ASDisplayShouldFetchBatchForScrollView(self, self.scrollDirection, self.scrollableDirections, contentOffset, velocity, self.collectionViewLayout.flipsHorizontallyInOppositeLayoutDirection)) {
     [self _beginBatchFetching];
   }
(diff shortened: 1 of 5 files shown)


Source/Private/ASBatchFetching.mm
```diff
@@ -87,11 +89,11 @@ BOOL ASDisplayShouldFetchBatchForContext(ASBatchContext *context,
   }
 
   CGFloat triggerDistance = viewLength * leadingScreens;
-  CGFloat remainingDistance = contentLength - viewLength - offset;
-  if (shouldRenderRTLLayout && ASScrollDirectionContainsHorizontalDirection(scrollableDirections)) {
+  CGFloat remainingDistance = 0;
+  if (!flipsHorizontallyInOppositeLayoutDirection && shouldRenderRTLLayout && ASScrollDirectionContainsHorizontalDirection(scrollableDirections)) {
       remainingDistance = offset;
```
nguyenhuy, line 94:
```suggestion
    remainingDistance = offset;
```

bolsinga (approved):

Source/ASCollectionView.mm
```diff
@@ -1857,7 +1857,7 @@ - (void)_checkForBatchFetching
 
 - (void)_beginBatchFetchingIfNeededWithContentOffset:(CGPoint)contentOffset velocity:(CGPoint)velocity
 {
-  if (ASDisplayShouldFetchBatchForScrollView(self, self.scrollDirection, self.scrollableDirections, contentOffset, velocity)) {
+  if (ASDisplayShouldFetchBatchForScrollView(self, self.scrollDirection, self.scrollableDirections, contentOffset, velocity, self.collectionViewLayout.flipsHorizontallyInOppositeLayoutDirection)) {
```
nguyenhuy, line 1860: As @wsdwsd0829 pointed out to me, UICollectionView actually asks for this property in its initializer and caches it. Perhaps we should do the same for ASCollectionView? Not a blocker for this diff though.

 
image(1)
Screen Shot 2021-05-03 at 10 25 00 AM

nguyenhuy (approved):
LGTM, especially with tests!

Source/ASCollectionView.mm
nguyenhuy: Nvm, it's actually cached by UICollectionView**Layout** in its initializer and so asking for `flipsHorizontallyInOppositeLayoutDirection` on the layout object here should be fast.

rcancro: Do I need to worry about calling `self.collectionViewLayout. flipsHorizontallyInOppositeLayoutDirection` on a potentially off-main thread? Since it is cached and presumably not called again, do I not need to worry about it?

nguyenhuy: I think the risk is very small. I checked and `_beginBatchFetchingIfNeededWithContentOffset` is called by multiple methods but all of them are called on the main thread. You can add a main thread assertion there just so we can be notified in the slim chance that's not the case.

foxware00: Worth noting @nguyenhuy and @rcancro `self.collectionViewLayout.flipsHorizontallyInOppositeLayoutDirection` is not availiable on iOS 10

PR #2002, opened 2021-06-03
Increase default diskCache byte limit from 20 to 50MB(PINCache default)

MussaCharles:
As discussed on our slack [thread](https://asyncdisplaykit.slack.com/archives/C0V63R86T/p1622253713035200)  since PINCache have already been optimized we no longer need to trade default size limit, thus I have simply increased the default disk cache limit from the original 20 MB to 50MB to support larger objects in the cache. 

NB: History to why the limit was set to 20MB can be found [here](https://github.com/TextureGroup/Texture/pull/595)

Thanks :)


garrettmoon (approved):
Looks good to me!

Issue #2009, opened 2021-06-19
[bug] ASCellNode has no background color in iOS 13


CainLuo:
ASCellNode has no background color in iOS 13 version, but it is normal in iOS 12 or iOS 14, how can I fix it? @garrettmoon 

```swift
class AboutCell: ASCellNode {
    private let titleNode = ASTextNode()
    private let lineNode = ASImageNode()

    init(_ model: AboutModel) {
        super.init()
        automaticallyManagesSubnodes = true
        accessoryType = .disclosureIndicator
        lineNode.isHidden = model.type == .privacy
        lineNode.style.flexGrow = 1.0
        lineNode.backgroundColor = UIColor(0xE9E9E9)
        titleNode.attributedText = NSAttributedString.attributed(model.title ?? "")
    }
    
    override func layoutSpecThatFits(_ constrainedSize: ASSizeRange) -> ASLayoutSpec {
        LayoutSpec {
            VStackLayout {
                CenterLayout(centeringOptions: .Y) {
                    InsetLayout(insets: .zero) {
                        titleNode
                    }
                }
                .minHeight(frame.maxY - 1)
                lineNode
            }
            .padding(UIEdgeInsets(top: 0, left: 15, bottom: 0, right: 15))
        }
    }
}
```

iOS 13
<img width="485" alt="image" src="https://user-images.githubusercontent.com/8857291/122630130-4c8f7600-d0f4-11eb-980d-630438ab9a01.png">

iOS 14
<img width="427" alt="image" src="https://user-images.githubusercontent.com/8857291/122630134-52855700-d0f4-11eb-97cc-d6fc0f28c299.png">

CainLuo:
Regarding the above issue, I found the specific problem, in the `- (void)setElement:(ASCollectionElement *)element` method of the ASTableView.mm file, the backgroundColor of node is used to set the _ ASTableViewCell's backgroundColor, I think this is incorrect, because node's backgroundColor should not be used to set as _ASTableViewCell's backgroundColor, but to open a separate property value of color to set separately. The temporary solution is to comment out `self.backgroundColor = node.backgroundColor;` and hope that it can be solved in subsequent versions of Texture.

denkeni:
It is actually a new behavior introduced since iOS 14 UIKit. When `cell.backgroundColor` is not set, it puts `_UISystemBackgroundView` to make your backgroundColor white, rather than transparent as before. No really related to Texture here.

[...]

PR #2010, opened 2021-06-20
Fix issues #2009

CainLuo:
[bug] ASCellNode has no background color in iOS 13

### DIFF
--- a/Source/ASCellNode.h
+++ b/Source/ASCellNode.h
@@ -211,6 +211,11 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  */
 @property UIEdgeInsets separatorInset;
 
+/* @abstract The inset of the cell background color
+ * ASTableView uses these properties when configuring UITableViewCells that host ASCellNodes.
+ */
+@property UIColor *cellBackgroundColor;
+
 @end
 
 @interface ASCellNode (Unavailable)
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -107,24 +107,27 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   ASCellNode *node = element.node;
   
   if (node) {
-    self.backgroundColor = node.backgroundColor;
-    self.selectedBackgroundView = node.selectedBackgroundView;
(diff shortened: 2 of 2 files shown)


CainLuo:
Regarding the above issue, I found the specific problem, in the - (void)setElement:(ASCollectionElement *)element method of the ASTableView.mm file, the backgroundColor of node is used to set the _ ASTableViewCell's backgroundColor, I think this is incorrect, because node's backgroundColor should not be used to set as _ASTableViewCell's backgroundColor, but to open a separate property value of color to set separately. The temporary solution is to comment out self.backgroundColor = node.backgroundColor; and hope that it can be solved in subsequent versions of Texture. #2010 

## Before: No cellBackgroundColor property

## After

ASCellNode.h add cellBackgroundColor

```swift
/* @abstract The inset of the cell background color
 * ASTableView uses these properties when configuring UITableViewCells that host ASCellNodes.
 */
@property UIColor *cellBackgroundColor;
```

ASTableNode.mm update setElement method:
```swift
- (void)setElement:(ASCollectionElement *)element
{
  _element = element;
  ASCellNode *node = element.node;
  
  if (node) {
    if (node.cellBackgroundColor != nil) {
        self.backgroundColor = node.cellBackgroundColor;
    }
    if (node.selectedBackgroundView != nil) {
        self.selectedBackgroundView = node.selectedBackgroundView;
    }
#if TARGET_OS_IOS
    self.separatorInset = node.separatorInset;
#endif
    self.selectionStyle = node.selectionStyle;
    self.focusStyle = node.focusStyle;
    self.accessoryType = node.accessoryType;

    if (node.tintColor != nil) {
        self.tintColor = node.tintColor;
    }

    // the following ensures that we clip the entire cell to it's bounds if node.clipsToBounds is set (the default)
    // This is actually a workaround for a bug we are seeing in some rare cases (selected background view
    // overlaps other cells if size of ASCellNode has changed.)
    self.clipsToBounds = node.clipsToBounds;
  }
  
  [node __setSelectedFromUIKit:self.selected];
  [node __setHighlightedFromUIKit:self.highlighted];
}
```

CainLuo:
@Adlai-Holler Hello, I fixed a problem, but I can't seem to merge it, how can I fix it?

PR #2011, opened 2021-06-24
Decode alternative image types in ASImageNode at render time

brandonli-google:
Adds support at the Texture level for decoding additional image formats at render time once the final image size is known

The Cartfile changes are because ASPINRemoteImageDownloader currently references API's that don't exist in the older version of PINRemoteImage

See (TODO, copy link to PR) for corresponding PINRemoteImage change

Upstreamed contribution from Google's fork of Texture

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
-github "pinterest/PINRemoteImage" "3.0.0-beta.14"
+github "pinterest/PINRemoteImage" "3.0.3"
 github "pinterest/PINCache" "3.0.1-beta.7"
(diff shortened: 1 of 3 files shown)


garrettmoon:
Just to confirm, this change requires the PINRemoteImage change be merged and released?

brandonli-google:
> Just to confirm, this change requires the PINRemoteImage change be merged and released?

Yes, that's right. Could add an additional build directive into PINRemoteImage (like #PINCustomDecoders or something) so that they can go in independently. WDYT?

garrettmoon:
I don't think that's necessary, PINRemoteImage is due a release soon anyway :)

PR #2014, opened 2021-07-06
Renamed ASNavigationController

zboralski:
To ASDKNavigationController.
This is to avoid conflicting with AuthenticationServices in
Xcode 13

### DIFF
--- a/Tests/ASNavigationControllerTests.mm
+++ b/Tests/ASNavigationControllerTests.mm
@@ -1,5 +1,5 @@
 //
-//  ASNavigationControllerTests.mm
+//  ASDKNavigationControllerTests.mm
 //  Texture
 //
 //  Copyright (c) Pinterest, Inc.  All rights reserved.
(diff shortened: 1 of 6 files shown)


benniebotha:
@nguyenhuy +1

Tests/ASNavigationControllerTests.mm
```diff
@@ -1,5 +1,5 @@
 //
-//  ASNavigationControllerTests.mm
```
benniebotha, line 2: Should probably rename the file too.

rcancro: Agreed! @zboralski could you add this rename and we can get this merged?

rcancro:
Thanks for this work! I apologize for not responding when you put up the original PR. We wanted to get this landed ASAP since iOS15 is imminent so I copied it and added the test rename [here](https://github.com/TextureGroup/Texture/pull/2020). Again, thank you for the contribution and sorry that we were not on the ball approving it!

PR #2017, opened 2021-08-04
[ASTextNode2] Second try to merge Google's ASTextNode2 changes

rcancro:
This time I included the accessibility changes, which required adding some of the new accessibility features in other parts of the code. Note, there are cases where not all of the accessibility was brought over because that was supposed to be covered in a separate PR made by someone from Google.


rcancro:
@wiseoldduck Here is a second attempt that includes the accessibility changes. Maybe the first time around i left those out because I thought someone from Google was looking in making an accessibility PR?

PR #2024, opened 2021-09-14
avoid padding a struct to be hashed

wsdwsd0829:
alternatively

- (NSUInteger)hash
{
#pragma pack(push,1)
  struct {
    CGSize viewportSize;
    ASScrollDirection scrollableDirections;
    NSUInteger elementsHash;
    NSUInteger layoutDelegateClassHash;
    NSUInteger additionalInfoHash;
  } data = {
    _viewportSize,
    _scrollableDirections,
    _elements.hash,
    _layoutDelegateClass.hash,
    [_additionalInfo hash]
  };
#pragma pack(pop)
  return ASHashBytes(&data, sizeof(data));
}

I am less confident about what #pragma pack(push,1) is doing

### DIFF
--- a/Source/Details/ASCollectionLayoutContext.mm
+++ b/Source/Details/ASCollectionLayoutContext.mm
@@ -85,7 +85,9 @@
 {
   struct {
     CGSize viewportSize;
-    ASScrollDirection scrollableDirections;
+    // Fields must be aligned without padding, otherwise hash result will be different for the same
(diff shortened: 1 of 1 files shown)


rcancro:
Let's try these things and see if they work:
* Try setting the values of the struct explicitly like the comment in `ASHashBytes` suggests
* Try changing the type of `ASScrollDirection` from `unsigned char` to `NSUInteger`

If neither of those work, let's land this! :)

Issue #1828, opened 2020-05-20
AssetsLibrary deprecated as of iOS9


bolsinga:
Texture supports both t the moment. When iOS9 is deprecated, AssetsLibrary can be removed.

mycroftcanner:
https://github.com/TextureGroup/Texture/issues/1992

mycroftcanner:
Any chance this could be conditionally compiled for iOS 9 and not for platforms where it is deprecated? @Adlai-Holler

Issue #1549, opened 2019-06-12
macOS Catalyst Support


b3ll:
Any chance of supporting Catalyst anytime soon?

There's a few issues with text rendering and some thread affinity stuff that seems to break when being run as UIKit on the mac.

maicki:
Hi Adam! Would you be so kind and post in some crash logs and more info in here. I don't have a Catalyst machine on hand yet, but would happily take a first look so we can tackle it as soon as we get a machine. Thanks and appreciated!

ptmt:
After explicitly building AsyncDisplayKit.framework with UIKit macOS support (meaning to enable that checkbox in Xcode 11) I was able to run my Texture app on macOS! 😃My only problem was ASCollectionView and IGListKIt, some assertions for setDatasource/setDelegate produced crashes. But after commenting them out + re-building IGLIstKit everything seems working at first glance, but need a more careful investigation. Can't express enough how awesome it is, async layout on 60 fps, the same UIView animations + other libraries. Magic!

So now I need to figure out what needs to be contributed back. Have someone else managed to built their app with Texture?

ptmt:
Although, window resizing s something: 
![Screen-Recording-2019-06-26-at-10 03 21-PM](https://user-images.githubusercontent.com/1004115/60207443-970bee00-985e-11e9-94a5-08a4b12ed624.gif)

strangeliu:
@ptmt it seems you enabled the layerBacked to the text nodes, disable it for a better resize animation. :)

ay8s:
@ptmt Got our iOS Composer at Buffer running on Mac last week. Think I hit the same issues as you, mainly around getting Texture to include the correct frameworks when building on Mac along with IGListKit tweaks. 

Things that spring to mind...
- Unsupported AssetsLibrary was being included, need to add check for UIKitForMac.
- ASDisplayNode CAEAGLLayer check

b3ll:
echoing what was already mentioned, thanks for those notes @ay8s 

The other main issue I saw was font rendering, it appears as though the CF attribute stripping of `NSAttributedString` is breaking system fonts (i.e. SF Pro, `+systemFontOfSize:`, etc.).

I have a pretty hacky fix here: https://github.com/b3ll/Texture/commit/140af43cfa9fd4de6737014c8a75269ff7973725

Didn't really test it that much though

ay8s:
[...]

PR #2034, opened 2021-09-29
Remove AssetsLibrary dependency for tvOS

### LINKED ISSUE #1992: AssetsLibrary is not available when building for tvOS Simulator.
Even setting the tvOS target to 9.0 I can't figure out a way to build Texture for tvOS it used to work fine though.

nguyenhuy:
- The framework isn't available on tvOS. This causes CocoaPods linting to fail which prevented me from pushing the new release out.
- One way to fix this is to have a different `default_subspecs` for tvOS that doesn't have AssetsLibrary subspec, but per-platform `default_subspecs` doesn't seem to be supported by CocoaPods. So I updated the subspec itself to only depend on the framework for iOS. This means the subspec is empty/useless for tvOS (and other platforms FWIW).
- Tested with `pod spec lint Texture.podspec`.
- Fixes #1992 and part of #1549. Also unblocks 3.1.0 release.
- For the long term, we can remove the subspec entirely when iOS 9 is deprecated (#1828).

### DIFF
--- a/Texture.podspec
+++ b/Texture.podspec
@@ -83,8 +83,8 @@ Pod::Spec.new do |spec|
   end
 
   spec.subspec 'AssetsLibrary' do |assetslib|
-    assetslib.frameworks = 'AssetsLibrary'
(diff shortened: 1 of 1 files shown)


bolsinga (approved):
Great find! Thanks.

PR #2075, opened 2023-01-24
[ASTextKitRenderer] Adding locking when accessing the text renderer cache

rcancro:
While NSCache is thread safe, that alone does not make `rendererForAttributes` thread safe. This experiment will add a lock to `rendererForAttributes` to see how that affects performance/stability. As second experiment will forego the cache altogether to validate that we are getting some value out of having a renderer cache.

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -115,7 +115,7 @@ static NSCache *sharedRendererCache()
  we maintain a LRU renderer cache that is queried via a unique key based on text kit attributes and constrained size. 
  */
 
-static ASTextKitRenderer *rendererForAttributes(ASTextKitAttributes attributes, CGSize constrainedSize)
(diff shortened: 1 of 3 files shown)


Source/ASTextNode.mm
```diff
@@ -130,6 +130,23 @@ - (BOOL)isEqual:(ASTextNodeRendererKey *)object
   return renderer;
 }
 
+static AS::RecursiveMutex __sharedRendererCacheInstanceLock__;
```
garrettmoon, line 133: nit: should this go at the top with the other locks? Or is the thinking to clean this up if it works?

garrettmoon (approved):

Source/ASTextNode.mm
rcancro: Which other locks? Since these are static C functions I put the lock right next to them. Is there a better place to put them?

garrettmoon: Ignore me, I was looking at a different file 😅

PR #2082, opened 2023-06-05
Fix xcode 14.3 deprecation warning

jpr98:
This PR fixes the following deprecation warning: "Function declaration without a prototype"


rcancro (approved):
Thanks for the contribution!

raycsh017 (approved):

PR #2084, opened 2023-07-25
[ASCellNodeVisibilityEvent] Add a new event when scrolling stops

rcancro:
We have `ASCellNodeVisibilityEvent` events that roughly correlate to the scrollViewDid… delegate methods in UIScrollView. With the current events we get a callback when a user stops dragging a cell, but if the cell decelerates we do not get an event when it comes to a rest. I’ve added  `ASCellNodeVisibilityEventDidStopScrolling` to have both `ASTableView` and `ASCollectionView` send this event to the cells in `_cellsForVisibilityUpdates` in `- (void)scrollViewDidEndDecelerating:(UIScrollView *)scrollView`.

I created unit tests to ensure that the proper events are being called for the proper scroll delegate methods.


raycsh017 (approved):
Seems like a pretty straightforward change adding a visibility event

PR #2089, opened 2023-08-25
ASDisplayNode documentation.

VAndrJ:
Swift does not support @abstract and other directives, leading to quick reference documentation problems in Swift code.

Example:
Screenshot 2023-08-25 at 15 20 58

This PR updates the ASDisplayNode.h documentation comments and adds information for some properties.

Example:
Screenshot 2023-08-25 at 15 20 46


rcancro (approved):
Thanks for the PR! This looks like a useful update to the comments!

PR #2093, opened 2023-09-08
Fix build errors and a crash in xcode 15

cacmartinez:
Fixes build errors when trying to compile with xcode15. 

It also fixes a crash happening on XCode15 when attempting to create a new image with either a height or width of 0 or less (UIGraphicsBeginImageContextWithOptions). This can happen when isCancelled returns true.


Source/Details/ASGraphicsContext.mm
```diff
@@ -121,7 +121,9 @@ NS_INLINE void ASConfigureExtendedRange(UIGraphicsImageRendererFormat *format)
   }
 
   // Bad OS or experiment flag. Use UIGraphics* API.
-  UIGraphicsBeginImageContextWithOptions(size, opaque, scale);
+  if (size.width > 0 && size.height > 0) {
+    UIGraphicsBeginImageContextWithOptions(size, opaque, scale);
```
rcancro, line 125: If we have an invalid size, I think we should exit early and return nil. Shortly after starting the image context we call `UIGraphicsGetImageFromCurrentImageContext` and `UIGraphicsEndImageContext`. I'm not sure what the behavior is if we end a graphics context without starting one.

cacmartinez: I've switched to an early return

rcancro (approved):

rahulVermaSimpplr:
Is this code is also available in the released version that is 3.1.0

keerthika-priya:
Is this issue happening only if the app is built using XCode 15?

PR #2107, opened 2024-05-17
Update to Xcode 15

andyfinnell:
## Summary

We want to make Texture build with recent tools like Xcode 15. Part of this is bumping the minimum supported OS to iOS 14.

I only removed the dead code (i.e. code gated on versions older than iOS 14). There are still some warnings about outdated APIs in use, but they are non-trivial changes to fix. Also, several of the examples rely on remote media (images or videos) that no longer exist. If they were trivially fixable, I fixed them. If not, I left them.

## Test plan

Run all the examples and verify it works as I expect. Run all the unit tests and make them pass.


rgnns (approved):

rgnns (approved):

mattxw:
Out of curiosity, why do Texture need to bump the minimum iOS version to 14?

brettwillis:
With this change from iOS 9 -> iOS 14, about 4% of active iOS users were cut off in one fell swoop. Is there a reason that it needed to bump it all the way to iOS 14 in terms of API usage? 

Bumping to only iOS 11 or 12, for example, would be significantly better than iOS 14 (eliminates ~2% of active users). But even then if there was no API usage that required this bump then why eliminate those users?

Source: https://iosref.com/ios.

Xcode can still compile for targets lower than iOS 14 with the `-Wl,-ld_classic` compiler flag (see https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Known-Issues). Can we have some consideration on reverting this change?

PR #2130, opened 2025-11-18
Fix: Transfer leadingScreensForBatching from pending state on node load

GianmarcoFolchi:
## Problem

When `leadingScreensForBatching` is set on `ASCollectionNode` or `ASTableNode` before the node loads, the value is not properly transferred to the underlying view. This causes the property to retain the default value of `2.0` instead of the configured value.

### Root Cause
It looks like `leadingScreensForBatching` was accidentally omitted when transferring the pendingState to the view in the `didLoad` function. 

#### Examples 
Broken Example
```
- (instancetype)init {
    self = [super init];
    if (self) {
        _collectionNode = [[ASCollectionNode alloc] init...];
        _collectionNode.leadingScreensForBatching = 3.6;  // Stored in pending state but never transferred to view
        [self addSubnode:_collectionNode];
    }
    return self;
}
```
Result: leadingScreensForBatching stays at 2.0

Working Example
```
override func didLoadInternal() {
    let collectionView = collectionNode.view  // This triggers node load
    collectionNode.leadingScreensForBatching = 2.4  // Set after load, goes directly to view
}
```
Result: leadingScreensForBatching is correctly set to 2.4 since the view had already been loaded. 

`leadingScreensForBatching` setter: leadingScreensForBatching goes to pendingState if it exists else it goes to the underlying view. The pendingState value wasn't being transferred to the underlying view on view creation. 
```
- (void)setLeadingScreensForBatching:(CGFloat)leadingScreensForBatching
{
  if ([self pendingState]) {
    _pendingState.leadingScreensForBatching = leadingScreensForBatching;
  } else {
    ASDisplayNodeAssert([self isNodeLoaded], @"ASCollectionNode should be loaded if pendingState doesn't exist");
    self.view.leadingScreensForBatching = leadingScreensForBatching;
  }
}
```

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -345,6 +345,10 @@
       [view setContentOffset:contentOffset animated:pendingState.animatesContentOffset];
     }
     
+    if (pendingState.leadingScreensForBatching != 0) {
+      view.leadingScreensForBatching = pendingState.leadingScreensForBatching;
+    }
+    
     const auto tuningParametersVector = pendingState->_tuningParameters;
     const auto tuningParametersVectorSize = tuningParametersVector.size();
     for (NSInteger rangeMode = 0; rangeMode < tuningParametersVectorSize; rangeMode++) {
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -82,6 +82,7 @@
     _flags.inverted = NO;
     _contentInset = UIEdgeInsetsZero;
     _contentOffset = CGPointZero;
+    _leadingScreensForBatching = 2.0;
     _flags.animatesContentOffset = NO;
     _flags.showsVerticalScrollIndicator = YES;
     _flags.showsHorizontalScrollIndicator = YES;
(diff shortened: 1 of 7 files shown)


Source/ASCollectionNode.mm
```diff
@@ -345,6 +345,10 @@ - (void)didLoad
       [view setContentOffset:contentOffset animated:pendingState.animatesContentOffset];
     }
     
+    if (pendingState.leadingScreensForBatching != 0) {
+      view.leadingScreensForBatching = pendingState.leadingScreensForBatching;
```
GianmarcoFolchi: With this approach, the only drawback that I can think of is that if someone sets `leadingScreensForBatching` to 0 intentionally it will be overridden with the default value of 2.0, but I don't know why anyone would set it to 0 intentionally.

rcancro: Could we set `leadingScreensForBatching` to `2.0` in the init method of `_ASCollectionPendingState`?

Source/ASTableNode.mm
```diff
@@ -176,6 +176,10 @@ - (void)didLoad
     if (!CGPointEqualToPoint(contentOffset, CGPointZero)) {
       [view setContentOffset:contentOffset animated:pendingState.animatesContentOffset];
     }
+
+     if (pendingState.leadingScreensForBatching != 0) {
```
rcancro: same as above, except set the default in `_ASTablePendingState`

rcancro (approved):

PR #2137, opened 2026-03-06
Fix compilation for Xcode 26.4: illegal operator syntax

### LINKED ISSUE #2136: Build broken with Xcode 26.4
The following code does not compile anymore with Xcode 26.4 

```
  [self _insideComposedCharacterSequences:line position:position block: ^(CGFloat left, CGFloat right, NSUInteger prev, NSUInteger next) {
    if (isVertical) {
      position = fabs(left - point.y) < fabs(right - point.y) < (right ? prev : next);
    } else {
      position = fabs(left - point.x) < fabs(right - point.x) < (right ? prev : next);
    }
  }];
  
  [self _insideEmoji:line position:position block: ^(CGFloat left, CGFloat right, NSUInteger prev, NSUInteger next) {
    if (isVertical) {
      position = fabs(left - point.y) < fabs(right - point.y) < (right ? prev : next);
    } else {
      position = fabs(left - point.x) < fabs(right - point.x) < (right ? prev : next);
    }
  }];

```
It's a bit hard to assume intent from what is basically illegal behavior, I assumed it was meant to be this:

https://github.com/FluidGroup/Texture/pull/11/changes/32c0679a4f19ce849e8d3efad77afcd43fc29e34

ntnmrndn:
Context: Original code is non defined behavior, went from warning to error with Xcode 26.4

Resolve https://github.com/TextureGroup/Texture/issues/2136
- Fix Xcode 26.4 compilation 
- Attempt to follow original intent (CC @Adlai-Holler)

### DIFF
--- a/Source/TextExperiment/Component/ASTextLayout.mm
+++ b/Source/TextExperiment/Component/ASTextLayout.mm
@@ -1597,17 +1597,17 @@ dispatch_semaphore_signal(_lock);
   
   [self _insideComposedCharacterSequences:line position:position block: ^(CGFloat left, CGFloat right, NSUInteger prev, NSUInteger next) {
     if (isVertical) {
-      position = fabs(left - point.y) < fabs(right - point.y) < (right ? prev : next);
+      position = fabs(left - point.y) < fabs(right - point.y) ? prev : next;
     } else {
-      position = fabs(left - point.x) < fabs(right - point.x) < (right ? prev : next);
+      position = fabs(left - point.x) < fabs(right - point.x) ? prev : next;
     }
   }];
-  
+
   [self _insideEmoji:line position:position block: ^(CGFloat left, CGFloat right, NSUInteger prev, NSUInteger next) {
     if (isVertical) {
-      position = fabs(left - point.y) < fabs(right - point.y) < (right ? prev : next);
+      position = fabs(left - point.y) < fabs(right - point.y) ? prev : next;
     } else {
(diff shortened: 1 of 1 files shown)


cjuneja:
hi there! Any ETA on when this fix will be merged. 

Xcode 26.4 and Xcode 26.4.1 both have the same issue.

PhoenixDeveloper:
Any movement?

Adlai-Holler:
FWIW code was adapted from here so that might provide good reference. I would be generally very careful about changing the logic — if we just need to fix the build I’d recommend changing the ternary condition to, for example, `right == YES` or whatever it is the compiler is complaining about surgically. Sorry I havent looked at the specific build issue but I’m guessing it’s a type coercion/syntax issue and we should leave the logic alone. https://github.com/ibireme/YYText/blob/7bd2aa41414736f6451241725778509fe75860b5/YYText/Component/YYTextLayout.m#L1508

cjuneja:
@Adlai-Holler 
Thanks for the update. Given that the new Xcode release is right around the corner and this issue persists, merging a fix on the library side would save thousands of developers countless hours of setup time.

While AI tools make local workarounds easy, main dependencies like Texture are critical to the ecosystem. Having an open PR sitting for months without resolution creates unnecessary friction for the community. I strongly urge the maintainers to reconsider prioritizing this fix—it would mean a lot to the developers who still rely heavily on this framework.

